PR #38, opened 2021-03-22
Restart cdc-sink delevopment

bobvawter:


go.mod
```diff
@@ -0,0 +1,8 @@
+module github.com/cockroachdb/cdc-sink
+
+go 1.16
+
+require (
+	github.com/cockroachdb/cockroach-go v2.0.1+incompatible
+	github.com/lib/pq v1.10.0
```
BramGruneir, line 7: I wanted to use pgx, but it didn't support json (or some other type) at the time.

BramGruneir (approved):
LGTM

Issue #45, opened 2021-08-06
ERROR: read binary tuple: partial copy data row (SQLSTATE XXUUU)


fabiog1901:
```bash
# Open terminal for the source
cockroach start-single-node --listen-addr :30000 --http-addr :30001 --store cockroach-data/30000 --insecure --background

cockroach sql --insecure --port 30000 <<EOF
CREATE SEQUENCE customer_seq;
create table t (id uuid, clob string, ts timestamp, primary key (id));
-- add enterprise license
SET CLUSTER SETTING cluster.organization = 'Workshop';
SET CLUSTER SETTING enterprise.license = 'crl-0-ELfszokGGAIiCFdvcmtzaG9w';
SET CLUSTER SETTING kv.rangefeed.enabled = true;
CREATE CHANGEFEED FOR TABLE defaultdb.t INTO 'experimental-http://127.0.0.1:30004/crdbusertable' WITH updated,resolved='10s';
EOF



# Open terminal for target
cockroach start-single-node --listen-addr :30002 --http-addr :30003 --store cockroach-data/30002 --insecure --background

cockroach sql --insecure --port 30002 <<EOF
create table t (id uuid, clob string, ts timestamp, primary key (id));
EOF


# Open terminal for cdc-sink
cdc-sink --port 30004 --conn postgresql://root@localhost:30002/defaultdb?sslmode=disable --config '[{"endpoint":"crdbusertable", "source_table":"t", "destination_database":"defaultdb", "destination_table":"t"}]'



# On source
cockroach sql --insecure --port 30000

insert into t values (
gen_random_uuid(),
[...]

PR #46, opened 2021-08-13
Patch pgx to work around COPY protocol bug

bobvawter:
This change uses a patched version of pgx that avoids splitting row data across
a CopyData message boundary. A specific test is added which increases the total
amount of changedata that might need to be sent to the database.

This is a partial resolution to #45, however this cannot fix cases where a
single row exceeds 64K-4 bytes in size. Futhermore, the maximum default size of
16MiB for a single wire message would also limit pathologically large rows.

X-Ref: https://github.com/cockroachdb/cockroach/issues/68805
X-Ref: https://github.com/bobvawter/pgx/commit/86f241d643fac3ddb450bfb0d52142b1b0442f19

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/46)


BramGruneir (approved):
LGTM

PR #51, opened 2021-09-08
cdc-sink: Add test for duplicate message delivery

bobvawter:
There was not a test for the case in which a CDC update is delivered twice.  As
it turns out, the COPY FROM acts as like INSERT, but with no way to ignore
conflicts. This change reverts the insert-message code to the previous
UPSERT-based approach.

An alternative approach, to use a per-cdc-sink, temporary, intermediate table
was considered, but this was less performant that just upserting into a common
table.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/51)

### DIFF
--- a/sink_table.go
+++ b/sink_table.go
@@ -19,9 +19,7 @@ import (
 	"strings"
 
 	"github.com/jackc/pgtype/pgxtype"
-	"github.com/jackc/pgx/v4"
 	"github.com/jackc/pgx/v4/pgxpool"
-	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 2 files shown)


BramGruneir (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/51#-:-Mj67N4H3RamYDZXK_kh:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @bobvawter)

Issue #64, opened 2021-11-29
Support pglogical replication


bobvawter:
This is a tracking issue to be able to consume a [logical replication feed](https://www.postgresql.org/docs/current/logical-replication.html) from PostgreSQL to act as a source of row data.

Much of the necessary wire-protocol is already present in [jackc/pglogrepl](https://github.com/jackc/pglogrepl).

In the desired end state, cdc-sink would support multiple frontends for row data, with a common backend that supports transactionally-consistent staging and reification of source data.

It's likely that a mapping/transform layer will exist in cdc-sink at some point in the future to perform on-the-fly schema or data-type adjustments. This should be accommodated as a desired end-state, but will not be added in the initial implementation.

Plan needs (revise as necessary):
* [x] Add automated integration testing with CRDB
* [x] Source-code reorganization to clearly identify front- and back-half concerns
* ~Clean up existing configuration ergonomics~ `pglogical` is now a separate subcommand.
* [x] Extend test-rig to also provide a matrix of postgres versions
* [x] Register and de-register a replication slot in a source database
* [x] Backfill and ingest
* [x] Checkpointing and error-recovery
* ~Lease the feed across multiple cdc-sink instances~ `pglogical` is now a separate subcommand.
* [x] Fail usefully in the presence of incompatible upstream schema changes

PR #67, opened 2021-12-01
cdc-sink: Persist raw json

bobvawter:
This change defers reifying the contents of a Line until the actual upsert into
a table. The goal is to improve efficiency and to make it simpler to generate
batches of lines from sources other than a CockroachDB CDC feed.

Related: #64

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/67)

### DIFF
--- a/sink.go
+++ b/sink.go
@@ -12,6 +12,7 @@ package main
 
 import (
 	"bufio"
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -147,11 +148,19 @@ func (s *Sink) deleteRows(ctx context.Context, tx pgxtype.Querier, lines []Line)
 		fmt.Fprintf(&statement, ") IN (")
 		var keys []interface{}
 		for i, line := range chunk {
+			// Parse out the primary key values.
+			key := make([]interface{}, 0, len(s.primaryKeyColumns))
+			dec := json.NewDecoder(bytes.NewReader(line.key))
+			dec.UseNumber()
+			if err := dec.Decode(&key); err != nil {
+				return err
+			}
+
 			if i > 0 {
 				fmt.Fprintf(&statement, ",")
 			}
 			fmt.Fprintf(&statement, "(")
-			for i, key := range line.Key {
+			for i, key := range key {
 				if i > 0 {
 					fmt.Fprintf(&statement, ",")
 				}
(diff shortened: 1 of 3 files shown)


BramGruneir (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/67)* status: all files reviewed, 1 unresolved discussion (waiting on @bobvawter)

___
*[sink.go, line 195 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/67#-Mpw-RiN67lbEz0ZRUo9:-Mpw-RiO7tHvu8uJBqcg:b-83sh11) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/9b235612ae884941f2a799c6b6f9c44836969621/sink.go#L195)):*
> ```Go
> 	}
> 
> 	// https://github.com/golang/go/wiki/SliceTricks#filtering-without-allocating
> ```

Nifty
___
*[sink_table.go, line 73 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/67#-Mpwd3VU6L1va-jr6oFG:-Mpwd3VU6L1va-jr6oFH:b-3t057v) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/9b235612ae884941f2a799c6b6f9c44836969621/sink_table.go#L73)):*
> ```Go
> 		into = append(into, k)
> 	}
> 	return into, nil
> ```

Why are you returning into here?  Is it not pass by reference on a string slice?

Whereas in the next parse, it's a map and you don't need to return it.

bobvawter (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/67)* status: all files reviewed, 1 unresolved discussion (waiting on @BramGruneir)

___
*[sink_table.go, line 73 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/67#-Mpwd3VU6L1va-jr6oFG:-MpxSUTp3NKKA5uSIAjO:b79ww6q) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/9b235612ae884941f2a799c6b6f9c44836969621/sink_table.go#L73)):*

Why are you returning into here?  Is it not pass by reference on a string slice?

Whereas in the next parse, it's a map and you don't need to return it.
</blockquote></details>

I'm returning `into` because I'm calling `append()`.   The type `[]T` is a by-value tuple of `(length, capacity, pointer to array)` and `append()` is guaranteed to modify at least one of them.  Passing the destination slice into the function allows the array to be allocated in the stack of a calling function, when calling `make()` with a known length.

Maps are always by-reference, opaque types.

BramGruneir (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/67#-:-MqGU3UN39T73wa6YqCw:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @bobvawter)

Issue #70, opened 2021-12-01
Single-statement promotion from staging to target tables


bobvawter:
It should be possible to construct SQL statements that promote from the staging table to the target tables with sufficiently complicated query generation. This would eliminate another round-trip of the synchronized data through the cdc-sink binary.

Steps:
* [ ] Ensure that mutation-ordering and deduplication within a range of resolved timestamps can be performed server-side.
* [ ] Change the storage of staged mutations to JSONB or to a table schema which allows querying of individual row/col/value tuples.
* [ ] Generate UPSERT statements which use the structured representation to mutate the target rows directly.

A possible complication is that it could be necessary for cdc-sink to perform some metadata queries to be able to generate expressions to perform the necessary type coercions.

bobvawter:
Closing due to effort/reward misalignment in current CRDB versions.  See commit message in 267db3a for a discussion.

PR #71, opened 2021-12-01
cdc-sink: Move promotion deduping to CRDB

bobvawter:
If there are multiple changes to the same row that have been staged, we only
want to apply the latest one, to avoid multiple updates to the same target row
when flushing a resolved timestamp. This is currently accomplished by having
cdc-sink read all staged changes in reverse-MVCC order and tracking which keys
have been touched.

This change moves the deduplication into CRDB to reduce the number of rows
returned to cdc-sink.  It is also a precursor to single-statement promotion in

The Line type is broken out with a Mutation base type that only holds the PK
and data to be upserted.  The timestamp information is unused by the UPSERT, so
we don't want to retrieve it.

X-Ref: #64
X-Ref: #70

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/71)

### DIFF
--- a/sink.go
+++ b/sink.go
@@ -120,12 +120,12 @@ func (s *Sink) HandleRequest(db *pgxpool.Pool, w http.ResponseWriter, r *http.Re
 }
 
 // deleteRows preforms all the deletes specified in lines.
-func (s *Sink) deleteRows(ctx context.Context, tx pgxtype.Querier, lines []Line) error {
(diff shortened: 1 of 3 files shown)


bobvawter:
Note that this includes PR #67

BramGruneir:
Can we find some time tomorrow to discuss this change?  The more I think about it, the more worried I am that it's a step backwards in terms of performance.

bobvawter:
Closing without merge based on effort/reward per #70 not being tractable in current versions of CRDB.

PR #73, opened 2021-12-13
cdc-sink: Overhaul

bobvawter:
This change overhauls the cdc-sink code to split it into well-defined packages
and APIs.

Notable functional changes:

A cdc-sink endpoint now operates on an entire database at once, since this is
the use-case that has been most prominent in discussions.

An "immediate" mode is supported, which applies incoming data without waiting
for resolved timestamps. This is intended for use when backfilling large
datasets or if a high-volume changefeed must catch up after an outage. It is
not expected to be the default configuration for cdc-sink.

The cdc-sink code can now detect and recover from a limited amount of
structural schema drift between the source and target databases. The schema for
each target database is held in memory and refreshed from time to time. Drift
is checked during resolved-timestamp flushes, which will effectively pause a
changefeed until the stored payloads are at least structurally-compatible with
the target tables.

Notes to reviewers:

The sinktypes package defines interfaces for the major moving parts of the
revised cdc-sink code base. At present, each interface is implemented by a
single type, but having small APIs has been useful in identifying the
independent parts of cdc-sink.

Similarly, the package structure may be overly fine-grained in favor of
identifying specific, small portions of code that compose easily.

A comment at the top of most files indicates where code was repackaged from.

Recommended package review order:
  sinktypes, cdc/handler, backend/mutation, backend/apply, backend/timestamp

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/73)


bobvawter:
I have pushed some revisions based on the desk review yesterday.

* cdc-sink is now based around a single changefeed per `SCHEMA` (i.e. the second-level namespace in each `DATABASE`). The url format for sending changes to cdc-sink now leads with `/targetDB/targetSchema/` to specify the destination table(s).
* The README file has been overhauled to reflect the new state of the code.  I have revalidated the quickstart flow.
* Added support for the new `webhook-https://` scheme in CockroachDB v21.2. cdc-sink can optionally speak TLS.

sravotto (commented):
> Reviewed 18 of 41 files at r1, 18 of 22 files at r3, 1 of 1 files at r5, 5 of 5 files at r6, 2 of 3 files at r7, 1 of 1 files at r8, 24 of 24 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/73)* status: all files reviewed, 2 unresolved discussions (waiting on @bobvawter and @BramGruneir)

___
*[internal/sinktypes/sinktypes.go, line 15 at r8](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-Mr2yG-T6XLgLsipJOke:-Mr2yG-T6XLgLsipJOkf:bl7ovmy) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/6c0245cfc1baa734f0219252707b4a92475b0a20/internal/sinktypes/sinktypes.go#L15)):*
> ```Go
> // the types into this package is to make it easy to compose
> // functionality as the cdc-sink project evolves.
> package sinktypes
> ```

Should we call this package simply "types" (or model/interfaces)? 
___
*[internal/sinktypes/sinktypes.go, line 79 at r8](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-Mr33Tp24V4D0ELRrZsl:-Mr33Tp24V4D0ELRrZsm:b-n13jza) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/6c0245cfc1baa734f0219252707b4a92475b0a20/internal/sinktypes/sinktypes.go#L79)):*
> ```Go
> 
> // A TimeSwapper maintains a durable map of string keys to timestamps.
> type TimeSwapper interface {
> ```

Perhaps we can call it TimeKeeper? And the method simply Update?
___
*[internal/util/ident/ident.go, line 12 at r8](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-Mr4C_JN5Krfk6rp6umK:-Mr4C_JN5Krfk6rp6umL:bu285gi) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/6c0245cfc1baa734f0219252707b4a92475b0a20/internal/util/ident/ident.go#L12)):*
> ```Go
> 
> // Package ident contains types for safely representing SQL identifiers.
> package ident
> ```

Like this approach. Makes the (rest of the) code much more robust and easier to read.

bobvawter (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/73)* status: 35 of 52 files reviewed, 2 unresolved discussions (waiting on @BramGruneir and @sravotto)

___
*[internal/sinktypes/sinktypes.go, line 15 at r8](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-Mr2yG-T6XLgLsipJOke:-Mr4R3H63bXM1s4nalxD:b6xm00i) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/6c0245cfc1baa734f0219252707b4a92475b0a20/internal/sinktypes/sinktypes.go#L15)):*

Should we call this package simply "types" (or model/interfaces)? 
</blockquote></details>

Renamed it to "types".
___
*[internal/sinktypes/sinktypes.go, line 79 at r8](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-Mr33Tp24V4D0ELRrZsl:-Mr4R7iZB9IgKxOEy_jh:b-yo1gbt) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/6c0245cfc1baa734f0219252707b4a92475b0a20/internal/sinktypes/sinktypes.go#L79)):*

Perhaps we can call it TimeKeeper? And the method simply Update?
</blockquote></details>

Switch it to `TimeKeeper.Put()` since it's map-like and one puts into a map.

BramGruneir:
___
*[internal/frontend/cdc/resolved.go, line 128 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNIHee8UOto0iCkPAd:-MrNIHee8UOto0iCkPAe:b-uyo6wg) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/frontend/cdc/resolved.go#L128)):*
> ```Go
> 		for table := range targetSchema {
> 			// Only perform work on target tables which match the
> 			// database and schema name specified in the resolved URL.
> ```

Do we throw an error if we get one that doesn't match?  A table we were not expecting?

BramGruneir (changes requested):
> Reviewed 15 of 41 files at r1, 5 of 22 files at r3, 1 of 5 files at r6, 1 of 3 files at r7, 1 of 1 files at r8, 12 of 24 files at r9, 17 of 17 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/73)* status: all files reviewed, 36 unresolved discussions (waiting on @bobvawter and @sravotto)

___
*[internal/backend/apply/apply.go, line 162 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-Mr8ESRXDXJrJU121TFV:-Mr8ESRXDXJrJU121TFW:bxj74s0) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/apply/apply.go#L162)):*
> ```Go
> 	defer res.Close()
> 
> 	for i, j := 0, batch.Len(); i < j; i++ {
> ```

Is the assignment of j needed here? 
___
*[internal/backend/apply/apply.go, line 194 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-Mr8GNft3o0XK_pOo2oy:-Mr8GNft3o0XK_pOo2oz:bfgnj53) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/apply/apply.go#L194)):*
> ```Go
> 			rawColName := col.Name.Raw()
> 			decoded, ok := temp[rawColName]
> 			delete(temp, rawColName)
> ```

These seems slightly out of order.  I think they'll work, (delete something that doesn't exist,) but the logic doesn't flow cleanly.
___
*[internal/backend/apply/apply.go, line 252 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-Mr8I2mA6A4ZXwqLshYu:-Mr8I2mB2fbwZYlT4iL1:b-piytqi) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/apply/apply.go#L252)):*
> ```Go
> 	_, _ = fmt.Fprintf(&delete, "DELETE FROM %s WHERE (", a.target)
> 	_, _ = fmt.Fprintf(&upsert, "UPSERT INTO %s (", a.target)
> 	for i := range colData {
> ```

why not 
for i, col := 

cleans up a lot of the colData[i] you're using all over all the place
___
*[internal/backend/apply/apply.go, line 270 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-Mr8IF1x-dXm6KNEHPFR:-Mr8IF1x-dXm6KNEHPFS:buo9nmt) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/apply/apply.go#L270)):*
> ```Go
> 	delete.WriteString(") IN (SELECT ")
> 	upsert.WriteString(") SELECT ")
> 	for i := range colData {
> ```

Same here
___
*[internal/backend/apply/factory.go, line 39 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-Mr8jNJcDgETyGrElGab:-Mr8jNJcDgETyGrElGac:bw5pm34) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/apply/factory.go#L39)):*
> ```Go
> 
> // Get returns a memoized instance of the Applier for the table.
> func (f *factory) Get(
> ```

I'm a little confused by the api here.

Get and New are the only exposed functions?

Get --> getUnlocked
New --> returns a factory

But no Create?
? --> createUnlocked
___
*[internal/backend/mutation/factory.go, line 43 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-Mr8kGWLE1T5Gx9ykqEl:-Mr8kGWLE1T5Gx9ykqEm:b-txkgjt) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/mutation/factory.go#L43)):*
> ```Go
> }
> 
> func (f *factory) createUnlocked(ctx context.Context, table ident.Table) (*store, error) {
> ```

Are createUnlocked and getUnlocked called from anywhere else?  Do you need them to be their own functions?  I'm guessing for unit tests?
___
*[internal/backend/mutation/store.go, line 50 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-Mr92HqyBi5UpmRzcdmj:-Mr92HqyBi5UpmRzcdmk:b10pf61) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/mutation/store.go#L50)):*
> ```Go
> // newStore constructs a new mutation store that will track pending
> // mutations to be applied to the given target table.
> func newStore(
> ```

Shouldn't this be New? 
___
*[internal/backend/mutation/store.go, line 133 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-Mr9GMX74HY6on6FZAwT:-Mr9GMX74HY6on6FZAwU:b-tzoljo) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/mutation/store.go#L133)):*
> ```Go
> 	batch := &pgx.Batch{}
> 
> 	for i := range mutations {
> ```

 Why not use for i, mut := range mutations instead of indexing in every time?
___
*[internal/backend/mutation/store_test.go, line 79 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-Mr9H6sJ5w2kSgnl3aID:-Mr9H6sJ5w2kSgnl3aIE:b-gdnvt3) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/mutation/store_test.go#L79)):*
> ```Go
> 
> 	// Dequeue.
> 	ret, err := s.Drain(ctx, dbInfo.Pool(),
> ```

style nit,  I thought our usual style was to put only the ctx on the first line then the rest on the 2nd.  Or to put everything on the 2nd line with the close bracket on the 3rd

Hmm... can we get the style guide from cockroach?
___
*[internal/backend/schemawatch/coldata.go, line 37 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-Mr9HhCO2dxkzZeR4HZv:-Mr9HhCO2dxkzZeR4HZw:b-yuatzs) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/schemawatch/coldata.go#L37)):*
> ```Go
> // Retrieve the primary key columns in their index-order, then append
> // any remaining non-generated columns.
> const sqlColumnsQuery = `
> ```

Wow... is the most efficient way?  It makes sense but I'm just surprised you need this complex a query.
___
*[internal/backend/schemawatch/coldata_test.go, line 90 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-Mr9I76n8JsSN3ULqglt:-Mr9I76o2qOAruNnZxS0:bw2z8qv) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/schemawatch/coldata_test.go#L90)):*
> ```Go
> 
> 	// Virtual columns not supported before v21.1
> 	if !strings.Contains(dbInfo.Version(), "v20.2.") {
> ```

If you have a few more tests that need version numbers, it makes sense to create a utility to determine if the version is newer, older or equal.  Can you just add a comment about how the older version supported is v20.2?  Perhaps that info should go in the readme as well.
___
*[internal/backend/schemawatch/watcher.go, line 48 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-Mr9J1s58pDmbR2fTjRw:-Mr9J1s58pDmbR2fTjRx:b-dxkqnt) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/schemawatch/watcher.go#L48)):*
> ```Go
> 	mu struct {
> 		sync.RWMutex
> 		cond sync.Cond // Conditional on the RLocker
> ```

Should cond be under mu if a lock isn't required for its use?
I see it's a weird mixed thing... Just needs a read lock.

So I find this a little confusing.  Simplifying the locking here might be worthwhile.  
___
*[internal/backend/schemawatch/watcher.go, line 112 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-Mr9Jp1X88OgOFQPowRt:-Mr9Jp1X88OgOFQPowRu:b8eb92s) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/schemawatch/watcher.go#L112)):*
> ```Go
> 	w.mu.data = data
> 	w.mu.Unlock()
> 	w.mu.cond.Broadcast()
> ```

Why take this out of the lock?
___
*[internal/backend/schemawatch/watcher.go, line 153 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-Mr9K-wl42XAhWTMOk1Q:-Mr9K-wl42XAhWTMOk1R:b897wzu) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/schemawatch/watcher.go#L153)):*
> ```Go
> 			// Respond to context cancellation or dropping the table.
> 			if !ok || ctx.Err() != nil {
> 				return
> ```

Shouldn't this do something with the error?
___
*[internal/backend/schemawatch/watcher_test.go, line 41 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNAela2T7ADGTCfc5r:-MrNAelb-MHEMF3ADEEK:btvj53s) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/schemawatch/watcher_test.go#L41)):*
> ```Go
> 
> 	// Bootstrap column.
> 	tblInfo, err := sinktest.CreateTable(ctx, dbName, "CREATE TABLE %s (pk INT PRIMARY KEY)")
> ```

Do you delete this table on test cleanup?
___
*[internal/backend/schemawatch/watcher_test.go, line 59 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNAQiuEmDL-L0wgtd1:-MrNAQiuEmDL-L0wgtd2:b2h3kwi) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/schemawatch/watcher_test.go#L59)):*
> ```Go
> 
> 	select {
> 	case <-time.After(10 * time.Second):
> ```

Consider setting these 10s much smaller, as this is unit test.
___
*[internal/backend/schemawatch/watchers.go, line 35 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNEMeQ7OxNr6gdi95E:-MrNEMeQ7OxNr6gdi95F:bs61fi9) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/schemawatch/watchers.go#L35)):*
> ```Go
> 
> // NewWatchers creates a Watchers factory.
> func NewWatchers(pool *pgxpool.Pool) (_ *Watchers, cancel func()) {
> ```

So in some factories you call this New, in other the name of the struct you're creating and in this case, NewStruct.  Can you do a quick audit and make them all match in style?
___
*[internal/backend/sinktest/info.go, line 23 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNFIhm3iijcZpyQfGN:-MrNFIhm3iijcZpyQfGO:b-sm1vsa) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/sinktest/info.go#L23)):*
> ```Go
> 
> // DBInfo encapsulates metadata and a connection to a database.
> type DBInfo struct {
> ```

perhaps name this file dbinfo to match?
___
*[internal/backend/sinktest/info.go, line 45 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNF8S98Nrc9uPsUXq1:-MrNF8S98Nrc9uPsUXq2:bywcj6x) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/sinktest/info.go#L45)):*
> ```Go
> }
> 
> // DeleteAll deletes (not TRUNCATEs) all rows in the table.
> ```

Why no Create here?  Just Drop?
___
*[internal/backend/sinktest/sinktest.go, line 102 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNFVNa34kSePQY5AqU:-MrNFVNa34kSePQY5AqV:bnmrnsh) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/sinktest/sinktest.go#L102)):*
> ```Go
> 
> 	cancel = func() {
> 		err := retry.Execute(ctx, db, fmt.Sprintf("DROP DATABASE IF EXISTS %s CASCADE", name))
> ```

Ah, dropping happens here.  Good, but perhaps this whole function should be part of dbinfo?

I'm having a bit an of an issue understanding the line between this file and info.go 
___
*[internal/backend/timestamp/store.go, line 30 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNGOVD-OoOTvorC933:-MrNGOVD-OoOTvorC934:b-fzh1nz) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/timestamp/store.go#L30)):*
> ```Go
> 
> // store implements a simple key/value store for HLC timestamps.
> type store struct {
> ```

Store is very generic.  Should we call it timekeeper instead?  The file and the type? 
___
*[internal/backend/timestamp/store_test.go, line 43 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNGeRXCHrTdLxxhVZz:-MrNGeRXCHrTdLxxhV_-:b-ftqj5c) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/timestamp/store_test.go#L43)):*
> ```Go
> 	prev := hlc.Zero()
> 	for i := 0; i <= count; i++ {
> 		next := hlc.New(int64(1000*i), i)
> ```

Please add comment on exactly what this test is testing.
___
*[internal/frontend/cdc/handler_test.go, line 34 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNH0ISE38U5cLpxMJD:-MrNH0ISE38U5cLpxMJE:bqj0htt) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/frontend/cdc/handler_test.go#L34)):*
> ```Go
> }
> 
> func testHandler(t *testing.T, immediate bool) {
> ```

Please add some comments within the tests.
___
*[internal/frontend/cdc/webhook.go, line 16 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNK_Vb2VK5tRDFSEu_:-MrNK_Vb2VK5tRDFSEua:b6oaet4) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/frontend/cdc/webhook.go#L16)):*
> ```Go
> )
> 
> var (
> ```

So should we be using webhook or https?  It it worth it to support both?  
___
*[internal/frontend/cdc/webhook.go, line 70 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNJ9ueDjknlVrz030Y:-MrNJ9ueDjknlVrz030Z:b8dn4wm) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/frontend/cdc/webhook.go#L70)):*
> ```Go
> 	}
> 
> 	// First, we'll aggregate the mutations by target table. We know
> ```

You have a first, but not a second or third.
___
*[internal/frontend/server/integration_test.go, line 49 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNU17y-BC9poF25qwe:-MrNU17y-BC9poF25qwf:bhmzeek) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/frontend/server/integration_test.go#L49)):*
> ```Go
> 	// Prefer the webhook format for current versions of CRDB.
> 	useWebhook := true
> 	if strings.Contains(dbInfo.Version(), "v20.2.") || strings.Contains(dbInfo.Version(), "v21.1.") {
> ```

Again, a nice version function for comparing would be better.
___
*[internal/frontend/server/integration_test.go, line 56 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNUEfD0ZlBvCHXcFqL:-MrNUEfD0ZlBvCHXcFqM:b-gupr23) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/frontend/server/integration_test.go#L56)):*
> ```Go
> 	defer func() { *GenerateSelfSigned = false }()
> 
> 	srv, err := newServer(ctx, "127.0.0.1:0", dbInfo.Pool().Config().ConnString(), false)
> ```

Do we need to worry if that's already in use?
___
*[internal/frontend/server/integration_test.go, line 84 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNUODV589Zk94bLP6h:-MrNUODV589Zk94bLP6i:bes7kjy) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/frontend/server/integration_test.go#L84)):*
> ```Go
> 	if useWebhook {
> 		feedURL = url.URL{
> 			Scheme:   "webhook-https",
> ```

So webhook is beta, http is experimental.  What is the difference?  (This is not a question for you.)
___
*[internal/frontend/server/server.go, line 45 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNVIVlE_o3QSPYQBvB:-MrNVIVlE_o3QSPYQBvC:b-r484i9) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/frontend/server/server.go#L45)):*
> ```Go
> 	ConnectionString = flag.String(
> 		"conn",
> 		"postgresql://root@localhost:26257/?sslmode=disable",
> ```

We're going to need to get this working with certs and a non-root user soon.
___
*[internal/frontend/server/server.go, line 49 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNVTxyE_0GXyNOOfU8:-MrNVTxyE_0GXyNOOfU9:b-kjok3w) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/frontend/server/server.go#L49)):*
> ```Go
> 	)
> 
> 	IgnoreResolved = flag.Bool("ignoreResolved", false,
> ```

This shouldn't be a flag on all of cdc-sink, but instead on a per feed basis.
___
*[internal/frontend/server/server.go, line 136 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNVpfr4YwomB0pkkNd:-MrNVpfr4YwomB0pkkNe:bb55or) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/frontend/server/server.go#L136)):*
> ```Go
> 	}
> 
> 	go func() {
> ```

Go really needs a better closer.  Or at least some syntactic sugar over this.
___
*[internal/frontend/server/tls_config.go, line 42 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNW6fO5n3qtkGe3SZ2:-MrNW6fO5n3qtkGe3SZ3:bseu9ol) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/frontend/server/tls_config.go#L42)):*
> ```Go
> 	// Loosely based on https://golang.org/src/crypto/tls/generate_cert.go
> 	if *GenerateSelfSigned {
> 		priv, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
> ```

I'm just going to trust this works.
___
*[internal/util/batches/batches.go, line 29 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNXCIY11P_TA-nzhy2:-MrNXCIZ15EFEvNLcIqZ:b-86d23v) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/util/batches/batches.go#L29)):*
> ```Go
> // of values in a batch-oriented fashion. The indexes provided to
> // the callback function are a half-open range [begin , end).
> func Batch(count int, fn func(begin, end int) error) error {
> ```

This is really nice.
___
*[internal/util/batches/batches.go, line 29 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNXSqQDWapkeIl1u4o:-MrNXSqQDWapkeIl1u4p:b-uogvvq) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/util/batches/batches.go#L29)):*
> ```Go
> // of values in a batch-oriented fashion. The indexes provided to
> // the callback function are a half-open range [begin , end).
> func Batch(count int, fn func(begin, end int) error) error {
> ```

Please add tests for batch.
___
*[internal/util/ident/ident.go, line 79 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNYZJM555H5zdRbY1s:-MrNYZJM555H5zdRbY1t:b6n0juc) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/util/ident/ident.go#L79)):*
> ```Go
> 
> // Empty returns true if the identifier is empty.
> func (n Ident) Empty() bool {
> ```

Shouldn't this be IsEmpty() 
___
*[internal/util/ident/ident.go, line 85 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNYNA1ARmGdeeB9TJL:-MrNYNA1ARmGdeeB9TJM:b-iiam7d) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/util/ident/ident.go#L85)):*
> ```Go
> // Raw returns the original, raw value.
> func (n Ident) Raw() string {
> 	return strings.ReplaceAll(n.q[1:len(n.q)-1], `""`, `"`)
> ```

Why not just store the raw?  Because this is clearly a lie :) 
___
*[internal/util/ident/ident_test.go, line 49 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNZ61T6o9pFWF-q7Nd:-MrNZ61T6o9pFWF-q7Ne:b-uxxx1b) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/util/ident/ident_test.go#L49)):*
> ```Go
> 		expectError bool
> 		qual        Qualification
> 	}{
> ```

Please add some more test cases that would produce errors.  Empty schema, empty db, empty table, etc.
___
*[internal/util/retry/retry.go, line 11 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNZVG3E32ZCmsXedHA:-MrNZVG3E32ZCmsXedHB:b6vi9sk) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/util/retry/retry.go#L11)):*
> ```Go
> // licenses/APL.txt.
> 
> // Package retry contains utility code for retrying database transactions.
> ```

Is there any way to not re-implement this ourselves?

BramGruneir (commented):
That's a lot of code, I went through it all.  
Not many changes needed I think.

I still don't like the frontend/backend names.  

How about sources/target? 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-:-MrNa0ZwADTqoppCZLCg:biv6qr7)* status: all files reviewed, 36 unresolved discussions (waiting on @bobvawter and @sravotto)

bobvawter (commented):
TFTR. PTAL. TANSTAAFL.

Renamed the mid-level packages to to `sources` and `target`.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-:-MrNhfq92VLfGZ0CBPMy:bptmlr7)* status: 20 of 55 files reviewed, 36 unresolved discussions (waiting on @BramGruneir and @sravotto)

___
*[internal/util/batches/batches.go, line 29 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNXCIY11P_TA-nzhy2:-MrOP-M42PUpw4H5guIP:b-ucfkn4) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/util/batches/batches.go#L29)):*

This is really nice.
</blockquote></details>

Thanks.
___
*[internal/util/batches/batches.go, line 29 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNXSqQDWapkeIl1u4o:-MrORR3uEVHRcpspPlQL:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/util/batches/batches.go#L29)):*

Please add tests for batch.
</blockquote></details>

Done.
___
*[internal/util/ident/ident.go, line 79 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNYZJM555H5zdRbY1s:-MrOReLI6iTrvCDvGEzK:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/util/ident/ident.go#L79)):*

Shouldn't this be IsEmpty() 
</blockquote></details>

Done.
___
*[internal/util/ident/ident.go, line 85 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNYNA1ARmGdeeB9TJL:-MrORxm62h9mIoA5lv0p:b-gjua7q) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/util/ident/ident.go#L85)):*

Why not just store the raw?  Because this is clearly a lie :) 
</blockquote></details>

Changed to store both the quoted and the raw, since they're used in pretty much equal proportions throughout the code.
___
*[internal/util/ident/ident_test.go, line 49 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNZ61T6o9pFWF-q7Nd:-MrOTAVREwG0c0LkUPSa:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/util/ident/ident_test.go#L49)):*

Please add some more test cases that would produce errors.  Empty schema, empty db, empty table, etc.
</blockquote></details>

Done.
___
*[internal/backend/sinktest/info.go, line 23 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNFIhm3iijcZpyQfGN:-MrO0oUl6-_KE-nFPkHk:bulkmt2) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/sinktest/info.go#L23)):*

perhaps name this file dbinfo to match?
</blockquote></details>

I split out `db_info.go` and `table_info.go`.  Moved the relevant creation functions into those two files, as well.
___
*[internal/backend/sinktest/info.go, line 45 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNF8S98Nrc9uPsUXq1:-MrO2OVU4AMAlpPom8M1:bsjte78) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/sinktest/info.go#L45)):*

Why no Create here?  Just Drop?
</blockquote></details>

See above comment.
___
*[internal/backend/apply/apply.go, line 162 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-Mr8ESRXDXJrJU121TFV:-MrNiFlA-NL_Wu-CHLsN:b-gj2er0) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/apply/apply.go#L162)):*

Is the assignment of j needed here? 
</blockquote></details>

Changed this to use a countdown loop instead.  The `pgx.Batch` type is opaque.
___
*[internal/backend/apply/apply.go, line 194 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-Mr8GNft3o0XK_pOo2oy:-MrNm0gA8EMxVISnvG1j:b-xxy25w) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/apply/apply.go#L194)):*

These seems slightly out of order.  I think they'll work, (delete something that doesn't exist,) but the logic doesn't flow cleanly.
</blockquote></details>

I reworked this to not use `delete()`, but to instead accumulate a set of known column names that are present in the incoming column data.  The variable names were expanded to make them easier to think through.
___
*[internal/backend/apply/apply.go, line 252 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-Mr8I2mA6A4ZXwqLshYu:-MrNnNi58J3padMrbgWO:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/apply/apply.go#L252)):*

why not 
for i, col := 

cleans up a lot of the colData[i] you're using all over all the place
</blockquote></details>

Done.
___
*[internal/backend/apply/apply.go, line 270 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-Mr8IF1x-dXm6KNEHPFR:-MrNnOVP5wh69TeIh9UU:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/apply/apply.go#L270)):*

Same here
</blockquote></details>

Done.
___
*[internal/backend/apply/factory.go, line 39 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-Mr8jNJcDgETyGrElGab:-MrNpR1WAAKBrYvmmZKW:b-ei687p) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/apply/factory.go#L39)):*

I'm a little confused by the api here.

Get and New are the only exposed functions?

Get --> getUnlocked
New --> returns a factory

But no Create?
? --> createUnlocked

</blockquote></details>

`Get()` calls `createUnlocked()`.  I've renamed `createUnlocked()` to `getOrCreateUnlocked()` to be more explicit.
___
*[internal/backend/mutation/factory.go, line 43 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-Mr8kGWLE1T5Gx9ykqEl:-MrNpZ2SBV0SRd-PuIf8:b-953gi0) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/mutation/factory.go#L43)):*

Are createUnlocked and getUnlocked called from anywhere else?  Do you need them to be their own functions?  I'm guessing for unit tests?
</blockquote></details>

They exist as two different functions because `getUnlocked()` is read-locked and `createUnlocked()` is write-locked.  The create method has been renamed and comments added to explain the differences in locking.
___
*[internal/backend/mutation/store.go, line 50 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-Mr92HqyBi5UpmRzcdmj:-MrNpmqnB6fxDQsJPGxl:b9ohtg) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/mutation/store.go#L50)):*

Shouldn't this be New? 
</blockquote></details>

`newStore()` isn't public because it is called only by the factory, and `new` is a keyword.
___
*[internal/backend/mutation/store.go, line 133 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-Mr9GMX74HY6on6FZAwT:-MrNqDjG86hJx_onXo0v:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/mutation/store.go#L133)):*

 Why not use for i, mut := range mutations instead of indexing in every time?
</blockquote></details>

Done.
___
*[internal/backend/mutation/store_test.go, line 79 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-Mr9H6sJ5w2kSgnl3aID:-MrNqJ2i8T_F9jpzoJG4:b-7ce2u3) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/mutation/store_test.go#L79)):*

style nit,  I thought our usual style was to put only the ctx on the first line then the rest on the 2nd.  Or to put everything on the 2nd line with the close bracket on the 3rd

Hmm... can we get the style guide from cockroach?
</blockquote></details>

Fixed the chop-down.
___
*[internal/backend/schemawatch/coldata.go, line 37 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-Mr9HhCO2dxkzZeR4HZv:-MrNqwc52cMfanvupYXj:b-injmq5) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/schemawatch/coldata.go#L37)):*

Wow... is the most efficient way?  It makes sense but I'm just surprised you need this complex a query.
</blockquote></details>

I couldn't find a simpler way to set the query up to concatenate two result sets, while preserving order in the result set.
___
*[internal/backend/schemawatch/coldata_test.go, line 90 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-Mr9I76n8JsSN3ULqglt:-MrOUNnz2005fJ51vecK:b-grrv6r) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/schemawatch/coldata_test.go#L90)):*

If you have a few more tests that need version numbers, it makes sense to create a utility to determine if the version is newer, older or equal.  Can you just add a comment about how the older version supported is v20.2?  Perhaps that info should go in the readme as well.
</blockquote></details>

Added a comment.  Will add a version enum or something if we wind up having more version checks in the future.
___
*[internal/backend/schemawatch/watcher.go, line 48 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-Mr9J1s58pDmbR2fTjRw:-MrNx6qO-i5i30B_hj7r:b-g2frvw) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/schemawatch/watcher.go#L48)):*

Should cond be under mu if a lock isn't required for its use?
I see it's a weird mixed thing... Just needs a read lock.

So I find this a little confusing.  Simplifying the locking here might be worthwhile.  
</blockquote></details>

I moved `cond` outside of the `mu` struct since it is never mutated. The condition, though, is still on `mu`'s read-lock.
___
*[internal/backend/schemawatch/watcher.go, line 112 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-Mr9Jp1X88OgOFQPowRt:-MrNxNdj1d3ClyG2v3wM:blzl1l5) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/schemawatch/watcher.go#L112)):*

Why take this out of the lock?
</blockquote></details>

Moved `cond` outside of `mu`, so the timing now makes more sense.  `sync.Cond` does not require callers to hold the lock when signaling.
___
*[internal/backend/schemawatch/watcher.go, line 153 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-Mr9K-wl42XAhWTMOk1Q:-MrNxonnAD7PXgarX7xI:bcmd6po) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/schemawatch/watcher.go#L153)):*

Shouldn't this do something with the error?
</blockquote></details>

There's nothing to do with the context's error.
___
*[internal/backend/schemawatch/watcher_test.go, line 41 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNAela2T7ADGTCfc5r:-MrNyYXG7eBtu3L2oWfw:bgf0esz) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/schemawatch/watcher_test.go#L41)):*

Do you delete this table on test cleanup?
</blockquote></details>

The database is dropped.
___
*[internal/backend/schemawatch/watcher_test.go, line 59 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNAQiuEmDL-L0wgtd1:-MrNyrcz2xJ54HZPI7b1:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/schemawatch/watcher_test.go#L59)):*

Consider setting these 10s much smaller, as this is unit test.
</blockquote></details>

Done.
___
*[internal/backend/schemawatch/watchers.go, line 35 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNEMeQ7OxNr6gdi95E:-MrNz4XpCC4a-FNLGfIW:b2q8rmb) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/schemawatch/watchers.go#L35)):*

So in some factories you call this New, in other the name of the struct you're creating and in this case, NewStruct.  Can you do a quick audit and make them all match in style?
</blockquote></details>

The methods are now all named `NewFactory()`
___
*[internal/backend/sinktest/sinktest.go, line 102 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNFVNa34kSePQY5AqU:-MrO2Sgl9YlbD1xBLN4Z:b102le) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/sinktest/sinktest.go#L102)):*

Ah, dropping happens here.  Good, but perhaps this whole function should be part of dbinfo?

I'm having a bit an of an issue understanding the line between this file and info.go 
</blockquote></details>

Moved to `db_info.go`.
___
*[internal/backend/timestamp/store.go, line 30 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNGOVD-OoOTvorC933:-MrO6weDBnNwyjIrOsde:b93nhdy) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/timestamp/store.go#L30)):*

Store is very generic.  Should we call it timekeeper instead?  The file and the type? 
</blockquote></details>

Renamed all to `timekeeper`.
___
*[internal/backend/timestamp/store_test.go, line 43 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNGeRXCHrTdLxxhVZz:-MrOCsfq7GszWF4y4iaj:btm4iml) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/backend/timestamp/store_test.go#L43)):*

Please add comment on exactly what this test is testing.
</blockquote></details>

Added comment to the test method.
___
*[internal/frontend/cdc/handler_test.go, line 34 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNH0ISE38U5cLpxMJD:-MrOFtCQ2IfveYK2lFzv:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/frontend/cdc/handler_test.go#L34)):*

Please add some comments within the tests.
</blockquote></details>

Done.
___
*[internal/frontend/cdc/resolved.go, line 128 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNIHee8UOto0iCkPAd:-MrOGJij-3GBidAeVkNX:b-81ygct) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/frontend/cdc/resolved.go#L128)):*

Do we throw an error if we get one that doesn't match?  A table we were not expecting?
</blockquote></details>

Revised the comment here that it's a filtering operation.
___
*[internal/frontend/cdc/webhook.go, line 16 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNK_Vb2VK5tRDFSEu_:-MrOMu1l48pWlA_8Jjvj:b9v858h) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/frontend/cdc/webhook.go#L16)):*

So should we be using webhook or https?  It it worth it to support both?  
</blockquote></details>

Supporting both really isn't that big of a deal, the difference is in the payload-decoder.  The bulk of the machinery is identical.  I also think we should strive to have compatibility with any CRDB release that's within the company's support window.
___
*[internal/frontend/cdc/webhook.go, line 70 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNJ9ueDjknlVrz030Y:-MrONQRbE9Y1la6Pp9qZ:b-j7aqt1) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/frontend/cdc/webhook.go#L70)):*

You have a first, but not a second or third.
</blockquote></details>

Turns out that was all that was necessary.  Fixed the comment.
___
*[internal/frontend/server/integration_test.go, line 49 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNU17y-BC9poF25qwe:-MrOUXtN1JPd5diOMxLL:b1b1zpj) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/frontend/server/integration_test.go#L49)):*

Again, a nice version function for comparing would be better.
</blockquote></details>

Agreed, but will save that for v22.1, if it becomes necessary.
___
*[internal/frontend/server/integration_test.go, line 56 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNUEfD0ZlBvCHXcFqL:-MrONjkG6CBKuVw4VLAo:blha2f3) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/frontend/server/integration_test.go#L56)):*

Do we need to worry if that's already in use?
</blockquote></details>

We don't, since we're binding to an ephemeral port number and then reading the bound address back when setting up the changefeed.
___
*[internal/frontend/server/server.go, line 45 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNVIVlE_o3QSPYQBvB:-MrONzsg6P0-2XzyGWok:b-efj7ic) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/frontend/server/server.go#L45)):*

We're going to need to get this working with certs and a non-root user soon.
</blockquote></details>

Sure, but that shouldn't affect any of the cdc-sink code.
___
*[internal/frontend/server/server.go, line 49 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNVTxyE_0GXyNOOfU8:-MrOO41j8_r18CFdTZZl:b-d985vi) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/frontend/server/server.go#L49)):*

This shouldn't be a flag on all of cdc-sink, but instead on a per feed basis.
</blockquote></details>

Per Slack discussion, immediate-mode has moved to an `?immediate=true` query parameter passed by the changefeed.  This will ensure that the configuration is globally synchronized. The integration test has been tweaked to have immediate and deferred sub-tests.
___
*[internal/frontend/server/tls_config.go, line 42 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNW6fO5n3qtkGe3SZ2:-MrOOO8A7G6hZjHi4P_K:b-ax5v8e) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/frontend/server/tls_config.go#L42)):*

I'm just going to trust this works.
</blockquote></details>

The resulting certificate is used in the `webhook-https` test.
___
*[internal/util/retry/retry.go, line 11 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNZVG3E32ZCmsXedHA:-MrOTstGC_VyMN40_eAm:b-7mbhss) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/util/retry/retry.go#L11)):*

Is there any way to not re-implement this ourselves?  
</blockquote></details>

This retry logic implements reentrancy, to delegate any retrying to the outermost call to `Loop()`.  This makes it much easier to write tests, since you can arbitrarily compose code that needs retry behavior.

bobvawter (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/73)* status: 20 of 55 files reviewed, 36 unresolved discussions (waiting on @BramGruneir and @sravotto)

___
*[internal/frontend/cdc/resolved.go, line 128 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNIHee8UOto0iCkPAd:-MrOrXcm8Kl-Q7GOB_f4:bwoc8as) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/frontend/cdc/resolved.go#L128)):*

Revised the comment here that it's a filtering operation.
</blockquote></details>

Meant to say that I moved the filtering operation into the `Snapshot()` method, which now takes the user-defined schema as an argumunt.

BramGruneir (approved):
I had to look up TANSTAAFL.

> Reviewed 35 of 35 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-:-MrSKkla6BQd5EusrKf_:b-mdafsk)* status: all files reviewed, 4 unresolved discussions (waiting on @BramGruneir and @sravotto)

___
*[internal/frontend/cdc/webhook.go, line 16 at r10](https://reviewable.io/reviews/cockroachdb/cdc-sink/73#-MrNK_Vb2VK5tRDFSEu_:-MrSIJUT7iyrdsFuOvUm:bb73djm) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f4daee50f44949354615172e7a619f7635bb5ca2/internal/frontend/cdc/webhook.go#L16)):*

Supporting both really isn't that big of a deal, the difference is in the payload-decoder.  The bulk of the machinery is identical.  I also think we should strive to have compatibility with any CRDB release that's within the company's support window.
</blockquote></details>

SGTM

BramGruneir:
:lgtm:

Issue #77, opened 2021-12-22
Authorize incoming feed connections


bobvawter:
In v21.2, the `webhook_auth_header` parameter allows an auth header to be added to the http requests made by cockroach.

The header value is arbitrary, so we can use a JWT token for stateless authorization. It would be straightforward to embed a list of allowable target db/schema pairs.  Code for hmac-signed JWT can be lifted from cacheroach. The signing keys could be loaded either from the filesystem, or stored in the target database.

An alternative would be to use a basic auth header, in combination with a bcrypted user table in the target database. This could also have a column to limit the allowable target schemas.

Auth support for older versions of cockroach could be added by allowing the auth token to be provided as a query parameter. Not ideal, though, since params are often logged by load balancers.

bobvawter:
Tracking issue for sending client certificates in `webhook-https://` scheme: https://github.com/cockroachdb/cockroach/issues/74230

PR #83, opened 2021-12-30
auth: Add JWT-based authorization

bobvawter:
This change updates cdc-sink with an authorization scheme based on signed JSON
Web Tokens. JWT's are chosen because they are well supported by many
authentication providers (e.g.: Google Cloud Identity, Auth0, AWS Cognito). We
use a custom claim in a globally-unique namespace to communicate which target
database schemas a caller may write into.

The list of public signing keys and revoked token ids are stored within the
_cdc_sink database. These are periodically refrshed by the cdc-sink process, or
in response to a HUP signal. Extending this to support the OpenID Connect
discovery protocol would be straightforward, if needed; the public keys would
need to be extrated from the discovered certificates.

A CLI tool is added to cdc-sink (and the existing behavior moved into a "start"
subcommand) to generate a signed JWT token using a user-provided private key.
The help message for the "make-jwt" subcommand provides a recipe using OpenSSL,
but any source of PEM-encoded private keys should work.

The "start" subcommand now accepts a --disableAuthentication flag to simplify
quickstarts, however all of the automated testing exercises the token flow.

X-Ref: #77

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/83)


sravotto (commented):
> Reviewed 21 of 21 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/83)* status: all files reviewed, 3 unresolved discussions (waiting on @bobvawter and @BramGruneir)

___
*[README.md, line 215 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/83#-MsjpK-h31fVgxZSM9s5:-MsjpK-h31fVgxZSM9s6:bfh5wx5) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/9b8c220ed637f1afba7345955f13f1c85b7659dd/README.md#L215)):*
> ```Markdown
> list of one or more target databases or user-defined schemas. Wildcards are supported.
> 
> ```json
> ```

I am not sure I understand the example. Wouldn't "*","*" allow all database/schemas? 
Also, could we change "sch" to "targets" (or something similar).
___
*[internal/cmd/mkjwt/mkjwt.go, line 44 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/83#-MsjrV66DQGp1UmfMRvP:-MsjrV671edOVZzN1hme:bkrhpks) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/9b8c220ed637f1afba7345955f13f1c85b7659dd/internal/cmd/mkjwt/mkjwt.go#L44)):*
> ```Go
> 		Use:   "make-jwt -key pk.pem -a database.schema -a other_database.schema ...",
> 		Example: fmt.Sprintf(`
> # Generate a EC private key using OpenSSL.
> ```

As an enhancement,  we could make it even easier by creating the PrivateKey in the make-jwt utility and also loading the public key in the database. 
___
*[internal/target/auth/jwt/jwt.go, line 193 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/83#-MsjyUzL5ctwRHzr6aHn:-MsjyUzL5ctwRHzr6aHo:b-gw7fzg) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/9b8c220ed637f1afba7345955f13f1c85b7659dd/internal/target/auth/jwt/jwt.go#L193)):*
> ```Go
> 					"id":     claims.ID,
> 					"schema": schema,
> 				}).Debug("successful authentication")
> ```

nit: I think it's more correct to talk about authorization rather than authentication (also with regards to types.Authenticator)

BramGruneir (commented):
I think we should discuss with some of the potential uses if JWT based authorization will work for them.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/83#-:-MspJlsj25t6BQfdxlBV:byhtd4r)* status: 15 of 21 files reviewed, 3 unresolved discussions (waiting on @bobvawter and @sravotto)

BramGruneir (commented):
*users

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/83#-:-MspJxzX6FjoXbSW6xUg:bg04okj)* status: 15 of 21 files reviewed, 3 unresolved discussions (waiting on @bobvawter and @sravotto)

BramGruneir (commented):
This would have been easier to review if you had split this into two.  The first adding in the cobra and the second  adding in jwt.

> Reviewed 2 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/83#-:-MspKMzl2-vs-mPBzjlG:beabp0z)* status: 17 of 21 files reviewed, 3 unresolved discussions (waiting on @bobvawter and @sravotto)

BramGruneir (changes requested):
I can see why you went with JWT, it seems straightforward to setup.

How will this work across multiple nodes?  If I've got 10 nodes all running cockroach and cdc-sink.  What's the procedure to setup the 2nd node?  Should it use the same key?  Do I need to share the keys between nodes?

Also, for updating the key, do you need to do that procedure on one node or all nodes?  And then restart the cdc job at the latest updated timestamp?

Please add some details on exactly how this works on multiple machines.

> Reviewed 15 of 21 files at r1, 4 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/83#-:-Mspn-EL5FKfKoo2z0BU:b-vlp2db)* status: all files reviewed, 15 unresolved discussions (waiting on @bobvawter)

___
*[internal/cmd/mkjwt/mkjwt.go, line 43 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/83#-MsqYJJ18v8H2U6A79hx:-MsqYJJ18v8H2U6A79hy:b-djgc11) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/8394fe4942e428050a6b6c85840b6739a504a244/internal/cmd/mkjwt/mkjwt.go#L43)):*
> ```Go
> 		Short: "generate JWT tokens from user-provided keys",
> 		Use:   "make-jwt -key pk.pem -a database.schema -a other_database.schema ...",
> 		Example: fmt.Sprintf(`
> ```

Shouldn't the line start on this one?  Does this not add a return at the start of the line?
___
*[internal/cmd/mkjwt/mkjwt.go, line 147 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/83#-Msq_6pGBtSpFoYhGXRr:-Msq_6pGBtSpFoYhGXRs:b-e4pcjk) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/8394fe4942e428050a6b6c85840b6739a504a244/internal/cmd/mkjwt/mkjwt.go#L147)):*
> ```Go
> 			return jwt.SigningMethodES512, nil
> 		default:
> 			return nil, errors.Errorf("unexpected elliptic curve %s", t.Curve.Params().Name)
> ```

This is a great error.  :)
___
*[internal/cmd/version/version.go, line 54 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/83#-MspoL8B8dfwGf5oI0VY:-MspoL8C1W7E7nea8Zad:b-46demb) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/8394fe4942e428050a6b6c85840b6739a504a244/internal/cmd/version/version.go#L54)):*
> ```Go
> 		},
> 	}
> }
> ```

This needs a test.
___
*[internal/source/cdc/handler.go, line 64 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/83#-Mspoda_3T9D00W4HKXk:-Mspoda_3T9D00W4HKXl:bca20v2) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/8394fe4942e428050a6b6c85840b6739a504a244/internal/source/cdc/handler.go#L64)):*
> ```Go
> 			}
> 		} else if token = r.URL.Query().Get("access_token"); token != "" {
> 			// Delete the token query parameter to prevent logging later
> ```

Does any of these query tokens or parameters get logged in the source cluster?  And more than that, won't the job be visible?  Something to be cautious about it.  We should at least document that in the readme.
___
*[internal/source/server/server.go, line 119 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/83#-MspqUBS2xPnr1t4TS-b:-MspqUBS2xPnr1t4TS-c:bvyxzpn) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/8394fe4942e428050a6b6c85840b6739a504a244/internal/source/server/server.go#L119)):*
> ```Go
> 	var cancelAuth func()
> 	if *DisableAuth {
> 		log.Warn("authentication disabled, any caller may write to the target database")
> ```

Maybe not warn but info?  If this is encountered, it's deliberate. 
___
*[internal/target/auth/jwt/claims.go, line 33 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/83#-MsprQlz2iFhCpWUiS4l:-MsprQlz2iFhCpWUiS4m:b-7j2qg) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/8394fe4942e428050a6b6c85840b6739a504a244/internal/target/auth/jwt/claims.go#L33)):*
> ```Go
> 
> // Valid implements jwt.Claims.
> func (c Claims) Valid() error {
> ```

Please add explicit unit tests for Valid.
___
*[internal/target/auth/jwt/jwt.go, line 210 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/83#-Msqb5jb80q0gLk_Rr-2:-Msqb5jb80q0gLk_Rr-3:bf5cqla) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/8394fe4942e428050a6b6c85840b6739a504a244/internal/target/auth/jwt/jwt.go#L210)):*
> ```Go
> )`
> 	selectKeysTemplate = `SELECT id, public_key FROM %s WHERE active`
> ```

nit: extra line
___
*[internal/target/auth/jwt/jwt.go, line 268 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/83#-MsqbGUd5mZGLnR-w30m:-MsqbGUd5mZGLnR-w30n:b-3cbkv6) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/8394fe4942e428050a6b6c85840b6739a504a244/internal/target/auth/jwt/jwt.go#L268)):*
> ```Go
> 	a.mu.publicKeys = nextKeys
> 	a.mu.revoked = nextRevoked
> 	log.Trace("refreshed JWT data")
> ```

Does this deserve a metric?
___
*[internal/target/auth/jwt/jwt.go, line 272 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/83#-Msqb__15Sc1XQAHqN5G:-Msqb__15Sc1XQAHqN5H:b-lt282b) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/8394fe4942e428050a6b6c85840b6739a504a244/internal/target/auth/jwt/jwt.go#L272)):*
> ```Go
> }
> 
> func matches(allowed, requested ident.Schema) bool {
> ```

This needs a specific test.
___
*[internal/target/auth/jwt/jwt_test.go, line 62 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/83#-MsqbfGY26KkQJevxsmx:-MsqbfGY26KkQJevxsmy:b-9tcu7q) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/8394fe4942e428050a6b6c85840b6739a504a244/internal/target/auth/jwt/jwt_test.go#L62)):*
> ```Go
> 	a.True(ok)
> 
> 	// Revoke the token id and revalidate.
> ```

Please add a test where a new token is added as well.
___
*[internal/target/auth/jwt/testing.go, line 41 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/83#-MsqcFFA2nr-LENrjbHR:-MsqcFFA2nr-LENrjbHS:bca0xho) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/8394fe4942e428050a6b6c85840b6739a504a244/internal/target/auth/jwt/testing.go#L41)):*
> ```Go
> 	}
> 
> 	method, signer, bytes, err := makeKeyBytes()
> ```

shouldn't this be = and not := ?
Doesn't this create new variables?

(might not, I'm a bit rusty)
___
*[internal/target/auth/reject/reject.go, line 32 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/83#-Msq_Xwo3bH3vERlZwc2:-Msq_Xwo3bH3vERlZwc3:bwdmt5c) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/8394fe4942e428050a6b6c85840b6739a504a244/internal/target/auth/reject/reject.go#L32)):*
> ```Go
> // Check always returns false.
> func (a *authenticator) Check(context.Context, ident.Schema, string) (bool, error) {
> 	return false, nil
> ```

Does this file need a test?  Probably not.
___
*[internal/target/auth/trust/trust.go, line 34 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/83#-Msq_hPs9SDusFJA-beT:-Msq_hPs9SDusFJA-beU:b-vzlplo) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/8394fe4942e428050a6b6c85840b6739a504a244/internal/target/auth/trust/trust.go#L34)):*
> ```Go
> // Check always returns true.
> func (a *authenticator) Check(context.Context, ident.Schema, string) (bool, error) {
> 	return true, nil
> ```

Does this one need a test?  Again, it's super basic... not sure.

BramGruneir (approved):
:lgtm:

> Reviewed 10 of 10 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/83#-:-Mt4DQ29CoNAYxfxWnLv:bnfp4nl)* status: all files reviewed, 3 unresolved discussions (waiting on @bobvawter and @sravotto)

___
*[internal/target/auth/jwt/jwt_test.go, line 62 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/83#-MsqbfGY26KkQJevxsmx:-Mt4Bf7S1R4jeiA5nNPV:bg81xho) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/8394fe4942e428050a6b6c85840b6739a504a244/internal/target/auth/jwt/jwt_test.go#L62)):*

Not sure I'm following. Token validation is stateless.
</blockquote></details>

yep, my bad here
___
*[internal/target/auth/jwt/testing.go, line 41 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/83#-MsqcFFA2nr-LENrjbHR:-Mt4Bzq651nfHojozMdw:b-5h9g49) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/8394fe4942e428050a6b6c85840b6739a504a244/internal/target/auth/jwt/testing.go#L41)):*

`bytes` is a new variable, the others are just updated.
 


</blockquote></details>

yeah, I couldn't remember if := created new every time, been a while, just reviewed the rules.

I'd be careful here, since scope matters.  If you happen to put this statement in a block, those updated variables will become new ones within that block's scope. 

I'd be inclined to declare bytes a line before instead and use =.

I would even be up for adding a linter to avoid this mixed use.

But I'll leave it up to you.

bobvawter (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/83)* status: 23 of 24 files reviewed, 3 unresolved discussions (waiting on @BramGruneir and @sravotto)

___
*[internal/target/auth/jwt/testing.go, line 41 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/83#-MsqcFFA2nr-LENrjbHR:-Mt4DvC6EPSVzyG092VP:byr3ca5) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/8394fe4942e428050a6b6c85840b6739a504a244/internal/target/auth/jwt/testing.go#L41)):*

yeah, I couldn't remember if := created new every time, been a while, just reviewed the rules.

I'd be careful here, since scope matters.  If you happen to put this statement in a block, those updated variables will become new ones within that block's scope. 

I'd be inclined to declare bytes a line before instead and use =.

I would even be up for adding a linter to avoid this mixed use.

But I'll leave it up to you.
</blockquote></details>

I see your point about the mixed case of `:=` being an easy oops to create during code movement.  Added a separate var declaration for bytes.

PR #85, opened 2022-01-26
cdc: Introduce a request type

bobvawter:
This change simplifies parameter passing from the Handler.ServeHTTP method into
the leaf handler methods. This is in preparation for adding additional, common
URL paramaters in the future.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/85)

PR #86, opened 2022-01-27
apply: Implement compare-and-set, deadlines

bobvawter:
Note to reviewers: this PR is based on PR #85 (adding the request type).

This change adds the ability to use certain columns in the target table for
compare-and-set operations. That is, staged mutations will only be applied if
the proposed data has a CAS tuple that is strictly greater than the CAS tuple
of the data currently in the target table.

Additionally, certain columns can be used to impose deadlines on the
application of certain mutations.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/86)

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,6 +216,77 @@ scrape_configs:
       - targets: [ '127.0.0.1:30004' ]
 ```
 
+## Modes
+
+`cdc-sink` supports a number of optional modes for each changefeed. All of these modes can be
+combined in a single changefeed.
+
+### Immediate
+
+Immediate mode writes incoming mutations to the target schema as soon as they are received, instead
+of waiting for a resolved timestamp. Transaction boundaries from the source database will not be
+preserved, but overall load on the destination database will be reduced.
+
+Immediate mode is enabled by adding the `?immediate=true` query parameter to the changefeed URL.
+
+### Compare-and-set
+
+A compare-and-set (CAS) mode allows `cdc-sink` to discard some updates, based on version- or
+time-like fields.
+
+**The use of CAS mode intentionally discards mutations and may cause data inconsistencies.**
+
+To opt into CAS mode for a feed, add a `?cas=<column_name>` query parameter to the changefeed. The
+named field should be of a type for which SQL defines a comparison operation. Common uses would be a
+`version INT` or an `updated_at TIMESTAMP` column. A mutation from the source database will be
+applied only if the CAS column is strictly greater than the value in the destination database.
+
+Consider a table with a `version INT` column that is used by adding a
+`?cas=version` query parameter to the changefeed URL. If `cdc-sink` receives a message to update
+some row in this table, the update will only be applied if there is no pre-existing row in the
+destination table, or if the update's `version` value is strictly greater than the existing row in
+the destination table. That is, updates are applied only if they would increase the `version` column
+or if they would insert a new row.
+
+Multiple CAS columns can be specified by repeating the query
+parameter: `?cas=version_major&cas=version_minor&cas=version_patch`. When multiple CAS columns are
+present, they are compared as a tuple. That is, the second column is only compared if the first
+column value is equal between the source and destination clusters. In this multi-column case, the
+following psuedo-sql clause is applied to each incoming update:
+
+```sql
+WHERE existing IS NULL OR (
+  (incoming.version_major, incoming.version_minor, incoming.version_patch) >
+  (existing.version_major, existing.version_minor, existing.version_patch)
+)
+```
+
+All tables in the feed must have all named `cas` columns or the changefeed will fail.
+
+Deletes from the source cluster are always applied.
+
+### Deadlines
+
+A deadline mode allows `cdc-sink` to discard some updates, by comparing a time-like field to the
+destination cluster's time. This is useful when a feed's data is advisory only or changefeed
+downtime is expected.
+
+**The use of deadlines intentionally discards mutations and may cause data inconsistencies.**
+
+To opt into deadline mode for a feed, two query parameters must be
+specified: `?dln=<column_name>&dli=<interval>`. The `dln` parameter names a column whose values can
+be coerced to a `TIMESTAMP`. The `dli` parameter specifies some number of seconds, or an interval
+value, such as `1m`.
+
+Given the configuration `?dln=updated_at&dli=1m`, each incoming update would behave as though it
+were filtered with a clause such as `WHERE incoming.updated_at > now() - '1m'::INTERVAL`.
+
+Multiple deadline columns may be specified by repeating pairs of `dln` and `dli` parameters.
+
+All tables in the feed must have all named `dln` columns.
+
+Deletes from the source cluster are always applied.
+
 ## Security Considerations
 
 At a high level, `cdc-sink` accepts network connections to apply arbitrary mutations to the target
--- a/internal/source/cdc/handler.go
+++ b/internal/source/cdc/handler.go
@@ -18,6 +18,7 @@ import (
 	"net/http"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/cockroachdb/cdc-sink/internal/types"
 	"github.com/cockroachdb/cdc-sink/internal/util/hlc"
@@ -29,10 +30,27 @@ import (
 
 // This file contains code repackaged from main.go
 
-// ImmediateParam is the name of a query parameter that will place a request
-// into "immediate" mode.  In this mode, mutations are written directly
-// to the target table and resolved timestamps are ignored.
-const ImmediateParam = "immediate"
+const (
+	// CASParam is the name of a query parameter that will apply a
+	// compare-and-set behavior to processed records. The value of the
+	// parameter should be a comma-separated list of column names.
+	CASParam = "cas"
+
+	// DeadlineColumnParam is the name of a query parameter that names
+	// a column in the incoming data to use in a deadline calculation.
+	DeadlineColumnParam = "dln"
+
+	// DeadlineIntervalParam is the name of a query parameter that
+	// specifies the deadline time, relative to now() on the target
+	// cluster.
+	DeadlineIntervalParam = "dli"
+
+	// ImmediateParam is the name of a query parameter that will place a
+	// request into "immediate" mode.  In this mode, mutations are
+	// written directly to the target table and resolved timestamps are
+	// ignored.
+	ImmediateParam = "immediate"
+)
 
 // Handler is an http.Handler for processing webhook requests
 // from a CockroachDB changefeed.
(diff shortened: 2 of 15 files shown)


bobvawter:
I'm open to changing up how the feed configuration is performed.  It feels a little bit forced to drive all of this through query parameters.  Maybe in a follow-on, all of this configuration lives in the destination database, and is stored in the `_cdc_sink` db?  We could have a utility method in the binary to set the configuration on a per-destination-schema basis.

bobvawter:
Added a benchmark to the PR to compare the different strategies. Numbers below are from my Mac running 21.2.4, two parallel workers, and 10s benchtime.


Using an in-memory store, the CAS operation is slower, as we'd expect for having to do a read-modify-write cycle.  In absolute terms, the difference isn't so bad.

```
BenchmarkApply/base-2         	   92144	    128949 ns/op
BenchmarkApply/cas-2          	   33788	    372027 ns/op
BenchmarkApply/deadline-2     	   77379	    155611 ns/op
BenchmarkApply/cas+deadline-2 	   31888	    388598 ns/op
```

With an on-disk store, we see that the disk latency dominates and the relative speed is equivalent.  The fsync on my laptop's storage is pretty awful; the average time lines up with the log-commit latency timeseries.

```
BenchmarkApply/base-2         	     387	  26827872 ns/op
BenchmarkApply/cas-2          	     409	  26391692 ns/op
BenchmarkApply/deadline-2     	     406	  25822864 ns/op
BenchmarkApply/cas+deadline-2 	     432	  26482002 ns/op
```

BramGruneir:
I really like the idea of moving the configuration into the destination database... I didn't think about that before.  We could validate the config upon startup.  It would allow for much more readable configurations and we could use the current syntax to write/compare with what is written there.  

Need to think about this more, but I think pulling on that thread would be interesting.

BramGruneir (approved):
:lgtm:

The generated sql seems correct.  But it's not that easy to check, so it has to rely on the tests.

> Reviewed 15 of 15 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/86#-:-MurVDYf9kWDnbVeyjXU:b-rfi062)* status: all files reviewed, 7 unresolved discussions (waiting on @bobvawter and @sravotto)

___
*[README.md, line 250 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/86#-MurCYWh2AOou66UNi8P:-MurCYWh2AOou66UNi8Q:b-w6sa55) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e490667f459ca887e16ad5b097ae59729351ff06/README.md#L250)):*
> ```Markdown
> All tables in the feed must have all named `cas` columns or the changefeed will fail.
> 
> Deletes from the source cluster are always applied.
> ```

Please add examples here for all of these different operation methods.  When it would discard and etc.

Also, you should describe the default operation as well.

And what, you can mix and match.
___
*[internal/target/apply/apply.go, line 292 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/86#-MurE0e59x0RQuIYwAoN:-MurE0e59x0RQuIYwAoO:b-s14gdo) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e490667f459ca887e16ad5b097ae59729351ff06/internal/target/apply/apply.go#L292)):*
> ```Go
> 	}
> 
> 	a.upserts.Add(float64(rowsUpserted))
> ```

Why not put this line directly in the loop instead of keeping here?  Is this operation slower?
___
*[internal/target/apply/apply.go, line 302 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/86#-MurES4p69vEbnR05lvN:-MurES4p69vEbnR05lvO:bu2zs1a) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e490667f459ca887e16ad5b097ae59729351ff06/internal/target/apply/apply.go#L302)):*
> ```Go
> // refreshUnlocked updates the apply with new column information.
> func (a *apply) refreshUnlocked(colData []types.ColData) error {
> 	// We want to verify that the cas columns actually exist in the
> ```

cas and deadline columns
___
*[internal/target/apply/apply.go, line 308 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/86#-MurEc6dECF9XsAU6jRO:-MurEc6dECF9XsAU6jRP:b-ccj5pp) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e490667f459ca887e16ad5b097ae59729351ff06/internal/target/apply/apply.go#L308)):*
> ```Go
> 		allColNames[col.Name] = struct{}{}
> 	}
> 	for _, col := range a.casColumns {
> ```

Perhaps we should store the column names in a map instead of an array?

Then you would only need to loop once right?  No need to loop the columns into a map, then again go through each array of column names.
___
*[internal/target/apply/apply_test.go, line 567 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/86#-MurSRqT8RJg5_AfFHQB:-MurSRqU1y_DFMwLyYjv:bjhup6) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e490667f459ca887e16ad5b097ae59729351ff06/internal/target/apply/apply_test.go#L567)):*
> ```Go
> }
> 
> func BenchmarkApply(b *testing.B) {
> ```

I love the benchmarks!
___
*[internal/target/apply/apply_test.go, line 568 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/86#-MurUUuHD7b2mugKn8st:-MurUUuHD7b2mugKn8su:b44pxs5) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e490667f459ca887e16ad5b097ae59729351ff06/internal/target/apply/apply_test.go#L568)):*
> ```Go
> 
> func BenchmarkApply(b *testing.B) {
> 	b.Run("base", func(b *testing.B) { benchConditions(b, false, false) })
> ```

There will be other options, perhaps make this a struct instead of just two bools.
___
*[internal/target/apply/templates.go, line 37 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/86#-MurHagg05x33LGyYXMG:-MurHagg05x33LGyYXMH:blawtck) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e490667f459ca887e16ad5b097ae59729351ff06/internal/target/apply/templates.go#L37)):*
> ```Go
> 
> 		// qualify is used with the "join" template to emit a list
> 		// of qualified database identifiers.
> ```

Please add an example here of what the format is.

bobvawter (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/86)* status: all files reviewed, 7 unresolved discussions (waiting on @BramGruneir and @sravotto)

___
*[README.md, line 250 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/86#-MurCYWh2AOou66UNi8P:-MuunlIe0GlHt8oU4SRA:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e490667f459ca887e16ad5b097ae59729351ff06/README.md#L250)):*

Please add examples here for all of these different operation methods.  When it would discard and etc.

Also, you should describe the default operation as well.

And what, you can mix and match.
</blockquote></details>

Done.
___
*[internal/target/apply/apply.go, line 292 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/86#-MurE0e59x0RQuIYwAoN:-Muuk203ElBI-7N-FFsX:b-p5rphl) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e490667f459ca887e16ad5b097ae59729351ff06/internal/target/apply/apply.go#L292)):*

Why not put this line directly in the loop instead of keeping here?  Is this operation slower?
</blockquote></details>

I want to defer the metric update until we know that the batch of operations has succeeded in its entirety.  Added a comment.
___
*[internal/target/apply/apply.go, line 302 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/86#-MurES4p69vEbnR05lvN:-MuugDCECY5r3TqNH65x:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e490667f459ca887e16ad5b097ae59729351ff06/internal/target/apply/apply.go#L302)):*

cas and deadline columns
</blockquote></details>

Done.
___
*[internal/target/apply/apply.go, line 308 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/86#-MurEc6dECF9XsAU6jRO:-MuugisA8UBiKjsGdig6:b5eonqa) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e490667f459ca887e16ad5b097ae59729351ff06/internal/target/apply/apply.go#L308)):*

Perhaps we should store the column names in a map instead of an array?

Then you would only need to loop once right?  No need to loop the columns into a map, then again go through each array of column names.
</blockquote></details>

I'm validating that the new columns in `colData` contain all ident referenced by the user-configured `casColumns` and `deadlines` fields in the `apply` struct.  I'd still need to loop over both `casColumns` and `deadlines`, regardless of how they're stored.  I suppose that the schema-watcher API could be changed to return a `map[ident.Ident]types.ColData` instead of `[]types.ColData`, but that seems invasive for code that is executed once and startup and then only again if there's a schema change.
___
*[internal/target/apply/apply_test.go, line 567 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/86#-MurSRqT8RJg5_AfFHQB:-MuuhXKIC777cy4S_6B0:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e490667f459ca887e16ad5b097ae59729351ff06/internal/target/apply/apply_test.go#L567)):*

I love the benchmarks!
</blockquote></details>

Done.
___
*[internal/target/apply/apply_test.go, line 568 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/86#-MurUUuHD7b2mugKn8st:-MuuiFi6-rfk5e0wz3Y5:b-fzjsaa) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e490667f459ca887e16ad5b097ae59729351ff06/internal/target/apply/apply_test.go#L568)):*

There will be other options, perhaps make this a struct instead of just two bools.
</blockquote></details>

Done.  Made `BenchmarkApply()` look more like a data-driven test with a new `benchConfig` type.
___
*[internal/target/apply/templates.go, line 37 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/86#-MurHagg05x33LGyYXMG:-Muuj9jGAD_HA38llwQh:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e490667f459ca887e16ad5b097ae59729351ff06/internal/target/apply/templates.go#L37)):*

Please add an example here of what the format is.
</blockquote></details>

Done.

PR #94, opened 2022-03-03
pglogical: Add delay between retries

bobvawter:
This changes adds a configurable, ten-second delay between restarting the
replication loop. This prevents excessive logging when, for example, there is a
schema mismatch between the source and target tables.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/94)

### DIFF
--- a/README.md
+++ b/README.md
@@ -302,6 +302,7 @@ Flags:
       --immediate                apply data without waiting for transaction boundaries
       --metricsAddr string       a host:port to serve metrics from at /_/varz
       --publicationName string   the publication within the source database to replicate
+      --retryDelay duration      the amount of time to sleep between replication retries (default 10s)
       --slotName string          the replication slot in the source database (default "cdc_sink")
       --sourceConn string        the source database's connection string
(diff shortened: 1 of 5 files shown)


bobvawter (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/94)* status: all files reviewed, 1 unresolved discussion (waiting on @BramGruneir and @sravotto)

___
*[internal/source/pglogical/integration_test.go, line 96 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/94#-MxopeVBD_YQxFY6oPXf:-MxqME412oZeNLE--ZM7:bk45f6x) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/30d6eb1e731361949286187fe4b0d7df32776f2c/internal/source/pglogical/integration_test.go#L96)):*

Do you have a test that actually performs the retry?
</blockquote></details>

Yes, the `TestChoas()` method below will trigger all manner of retries.

PR #96, opened 2022-03-14
workflows: Add go mod download to builds

bobvawter:
This build step will ensure that all referenced go modules will be downloaded
as part of the build. This ensures that when we send the go caches to the
GitHub workflows cache, they're complete, at least with respect to modules
used.  The GitHub cache key is based on the contents of the go.sum file and the
cache contents are only ever sent to the cache server once; you can't update an
existing cache key. This can create an opportunity for unnecessary re-downloads
if the build rules change.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/96)

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -34,14 +34,10 @@ jobs:
       - name: Go Build Cache
         uses: actions/cache@v2
         with:
-          path: ${{ steps.cache.outputs.go-build }}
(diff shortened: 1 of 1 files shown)


BramGruneir (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/96#-:-MyNMJ1vCOLVHiQ29A0o:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sravotto)

PR #106, opened 2022-03-17
timekeeper: Improve throughput.

bobvawter:
This change updates the PK structure of the resolved table to include the
timing data. This allows us to always append new values to distinct PK rows, so
we don't need to read though older MVCC data from previous resolved timestamps.

An out-of-band cleanup process is added to retire older resolved timestamps.

The benchmark shows a 10x improvement on my laptop, using an in-memory CRDB
store.

### DIFF
--- a/internal/source/cdc/handler_test.go
+++ b/internal/source/cdc/handler_test.go
@@ -53,10 +53,11 @@ func testHandler(t *testing.T, immediate bool) {
 		return
 	}
 
-	swapper, err := timekeeper.NewTimeKeeper(ctx, dbInfo.Pool(), Resolved)
+	swapper, cancel, err := timekeeper.NewTimeKeeper(ctx, dbInfo.Pool(), Resolved)
 	if !a.NoError(err) {
 		return
 	}
+	defer cancel()
 
 	watchers, cancel := schemawatch.NewWatchers(dbInfo.Pool())
 	defer cancel()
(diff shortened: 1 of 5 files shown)


BramGruneir (approved):
:lgtm:

But please add a note to the commit that this version is incompatible with pervious versions.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/106#-:-MyNuoaLA3-j3qqpFILk:bkj0de1)* status: all files reviewed, 2 unresolved discussions (waiting on @bobvawter and @sravotto)

___
*[internal/target/timekeeper/timekeeper.go, line 63 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/106#-MyNt6F_5WKS0oR13daJ:-MyNt6F_5WKS0oR13daK:berf2jl) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/839f4dfb45e8ed867db6d332da7e44d6e0f926ff/internal/target/timekeeper/timekeeper.go#L63)):*
> ```Go
> logical INT8 NOT NULL,
> PRIMARY KEY (key, nanos, logical)
> )
> ```

If using a pervious version cdc-sink, the timekeeper table will have to be deleted.

This should probably be mentioned in the readme for the commit.  Since we don't have an officially published version yet.
___
*[internal/target/timekeeper/timekeeper.go, line 139 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/106#-MyNtefc67a7fHwgi2pn:-MyNtefc67a7fHwgi2po:b-ifc6h7) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/839f4dfb45e8ed867db6d332da7e44d6e0f926ff/internal/target/timekeeper/timekeeper.go#L139)):*
> ```Go
> 		case <-ctx.Done():
> 			return
> 		case <-time.After(time.Second):
> ```

I think this should this be a configurable value.

PR #115, opened 2022-04-08
logical: Extract base behaviors for logical replication.

bobvawter:
This change separates the read-and-apply loop from the pglogical package. It
retains the two-goroutine approach of having a generator which connects to a
source database and an applier which drives an event-style API provided by the
logical replication loop.

When the producer routine encounters an error (e.g. a network glitch), it can
resynchronize with the upstream source and the applier via a rollback message.
This allows the least amount of disruption to any state which may be maintained
by the applier goroutine, as opposed to having to tear everything down and
restart.

The logical loop is tested by having a fake generator data source. A chaotic
wrapper around the Dialect APIs makes it easier to introduce errors throughout
the various code-paths.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/115)

### DIFF
--- a/internal/source/logical/chaos.go
+++ b/internal/source/logical/chaos.go
@@ -1,0 +1,101 @@
+// Copyright 2022 The Cockroach Authors.
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package logical
+
+import (
+	"context"
+	"math/rand"
+
+	"github.com/cockroachdb/cdc-sink/internal/types"
+	"github.com/cockroachdb/cdc-sink/internal/util/ident"
+	"github.com/cockroachdb/cdc-sink/internal/util/stamp"
+	"github.com/pkg/errors"
+)
+
+// ErrChaos is the error that will be injected by the WithChaos wrappers
+// in this package.
+var ErrChaos = errors.New("chaos")
+
+// WithChaos returns a wrapper around a Dialect that will inject errors
+// at various points throughout the execution.
+func WithChaos(delegate Dialect, prob float32) Dialect {
+	return &chaosDialect{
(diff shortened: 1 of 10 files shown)


sravotto (approved):
:lgtm:  

> Reviewed 8 of 10 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/115#-:-N0g0nad56v2-TZQshXC:b593w5t)* status: all files reviewed, 1 unresolved discussion (waiting on @bobvawter and @BramGruneir)

___
*[`internal/source/logical/logical.go` line 286 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/115#-N0bXioH7gNu08aar0dp:-N0bXioH7gNu08aar0dq:bq0x5ys) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/47cc8f13ab657956c7f9aa85d7188ad275edb16d/internal/source/logical/logical.go#L286)):*
> ```Go
> 
> 		if err := group.Wait(); err != nil {
> 			log.WithError(err).Errorf("error in replication loop; retrying in %s", l.retryDelay)
> ```

I have seen that the context get cancelled when there is an error, so we never really wait.

bobvawter (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/115)* status: 9 of 10 files reviewed, 1 unresolved discussion (waiting on @BramGruneir and @sravotto)

___
*[`internal/source/logical/logical.go` line 286 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/115#-N0bXioH7gNu08aar0dp:-N0g2BGO4cgZrsfV9eab:bynwjvm) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/47cc8f13ab657956c7f9aa85d7188ad275edb16d/internal/source/logical/logical.go#L286)):*

I have seen that the context get cancelled when there is an error, so we never really wait. 
</blockquote></details>

I was selecting on the wrong `context.Done`.  Fixed the variable name of the error-group context to not alias the context that's passed into the function.

PR #114, opened 2022-04-07
fan: Advance consistent point in high-rate scenario

bobvawter:
The existing code would never execute the onConsistent callback if the bucket
never becomes idle.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/114)

PR #116, opened 2022-04-25
logical replication from MySQL

sravotto:
This change adds initial support for logical replication from a MySQL server.
 It implements a transaction-based replication feed using global transaction identifiers (GTIDs).
  
Summary of changes:

- mylogical:  row level replication feed using global transaction identifiers (GTIDs).   See https://dev.mysql.com/doc/refman/8.0/en/replication-gtids.html
- logical: adding support for persisting replication state in the target database.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/116)


bobvawter (changes requested):
I know this is still a work in progress; my main feedback here is about the API that `logical` should expose to the `Dialect` implementations.  The need to save and restore the consistent point seems like it should just be handled by the logical loop (and maybe the pglogical code just ignores it, since it's tracked server-side).

> Reviewed 4 of 16 files at r1, 21 of 21 files at r2.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-:-N0XEPN82eZmz1LDv61m:b-wvh8u5)* status: 24 of 25 files reviewed, 14 unresolved discussions (waiting on @bobvawter, @BramGruneir, and @sravotto)

___
*[`-- commits` line 3 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0X7f1C31ljjXvzOk4Y):*
This base commit is a bit stale.  The current one is rebased onto master after merging PR #114 .  I had found and fixed the same race condition that you addressed, so you can unwind at least some of your test changes in the logical package.
___
*[`.github/docker-compose.yml` line 24 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0X89FoC2QSfhD7Vjvu:-N0X89FoC2QSfhD7Vjvv:b-qckk34) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/62299ff7c1c7dae774202ee1d9738e76a51789bd/.github/docker-compose.yml#L24)):*
> ```YAML
>     command: start-single-node --insecure
>   mysql-v8:
>     image: mysql:8-debian
> ```

Please keep the Debian base image.
___
*[`internal/cmd/mylogical/mylogical.go` line 75 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0X8l1Y9ftDq2-OoDeH:-N0X8l1Y9ftDq2-OoDeI:b-wyl45b) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/42c379cdb299a417036b1c5a082db3bc86666a69/internal/cmd/mylogical/mylogical.go#L75)):*
> ```Go
> // metricsServer starts a trivial prometheus endpoint server which runs
> // until the context is canceled.
> func metricsServer(bindAddr string) error {
> ```

I'll add a todo item to extract the common parts of the logical subcommands into a support package.
___
*[`internal/source/logical/checkpoint.go` line 26 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0X95TN8-0Kh-FbyLcj:-N0X95TN8-0Kh-FbyLck:b-xq1kvh) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/42c379cdb299a417036b1c5a082db3bc86666a69/internal/source/logical/checkpoint.go#L26)):*
> ```Go
> // A checkpoint is specific for a particular source,
> // identified by a key that should be universally unique
> type Checkpoint struct {
> ```

I'm trying to keep any code which executes sql commands within the `target` or `util` packages.  This type is a pretty generic string-string mapping, maybe call it `Memo` instead?  It should also have its own test cases as well.
___
*[`internal/source/logical/dialect.go` line 81 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0XANI62FB9deqbafWE:-N0XANI62FB9deqbafWF:b-tzvnyq) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/42c379cdb299a417036b1c5a082db3bc86666a69/internal/source/logical/dialect.go#L81)):*
> ```Go
> 	GetTargetDB() ident.Ident
> 	// RestoreConsistentPoint restores the state saved into the target database
> 	RestoreConsistentPoint(ctx context.Context, t stamp.Stamp) error
> ```

Why does this API exist?  If the consistent point needs to be tracked by the logical loop, that should happen without the involvement of the Dialect.
___
*[`internal/source/mylogical/config.go` line 84 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0XBJ-D5JbvS1WzMMkJ:-N0XBJ-EBh_7F3m2gHP2:b-spx7p7) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/42c379cdb299a417036b1c5a082db3bc86666a69/internal/source/mylogical/config.go#L84)):*
> ```Go
> 		return errors.New("no ServerID was configured")
> 	}
> 	u, err := dburl.Parse(c.SourceConn)
> ```

What does this API do differently from the regular `url.Parse`?
___
*[`internal/source/mylogical/conn.go` line 37 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0XBUS426YXWa-IQAda:-N0XBUS426YXWa-IQAdb:bej1ph1) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/42c379cdb299a417036b1c5a082db3bc86666a69/internal/source/mylogical/conn.go#L37)):*
> ```Go
> 
> // lsnStamp adapts the LSN offset type to a comparable Stamp value.
> type lsnStamp int64
> ```

I think the base type needs to be a GTIDSet, not just the sequence number, since the target mysql instance could be aggregating changes from multiple upstream servers.  Either way, the type probably needs a name other than `lsn`, since that's a postgres-ism.
___
*[`internal/source/mylogical/conn.go` line 48 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0XBqZ9-64sCvdEJj3-:-N0XBqZ9-64sCvdEJj30:b-vnpeep) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/42c379cdb299a417036b1c5a082db3bc86666a69/internal/source/mylogical/conn.go#L48)):*
> ```Go
> func (s lsnStamp) Less(other stamp.Stamp) bool { return s < other.(lsnStamp) }
> 
> func (s lsnStamp) Marshall() string { return strconv.FormatInt(int64(s), 10) }
> ```

The canonical interface for this kind of behavior is https://pkg.go.dev/encoding#TextMarshaler

IIRC, implementing that interface makes a type directly usable by the `db/sql` and `pgx` packages as a parameter value.
___
*[`internal/source/mylogical/conn.go` line 62 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0XCE6vDSxve98HP_YI:-N0XCE6vDSxve98HP_YJ:b-uytnfk) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/42c379cdb299a417036b1c5a082db3bc86666a69/internal/source/mylogical/conn.go#L62)):*
> ```Go
> 	columns map[ident.Table][]types.ColData
> 	// Last GTIDEvent
> 	lastGTIDEvent int64
> ```

This should probably be an appropriately-typed `Stamp`, not a raw number.
___
*[`internal/source/mylogical/conn.go` line 81 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0XCNBx4JyqNM97TxkJ:-N0XCNBx4JyqNM97TxkK:bambo7g) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/42c379cdb299a417036b1c5a082db3bc86666a69/internal/source/mylogical/conn.go#L81)):*
> ```Go
> )
> 
> func (s MutationType) String() string {
> ```

Use stringer instead. See https://github.com/cockroachdb/cdc-sink/blob/master/internal/util/ident/ident.go#L50 for an example.  The nice thing about that helper is that the generate code will fail to compile if it's ever out of date.
___
*[`internal/source/mylogical/conn.go` line 87 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0XCiZd-ehI9lMoCjz8:-N0XCiZd-ehI9lMoCjz9:b-hef2s5) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/42c379cdb299a417036b1c5a082db3bc86666a69/internal/source/mylogical/conn.go#L87)):*
> ```Go
> var _ logical.Dialect = (*Conn)(nil)
> 
> // NewConn constructs a new pglogical replication feed.
> ```

s/pglogical/mysql/
___
*[`internal/source/mylogical/conn.go` line 160 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0XCmqp5N-AQQYzvq4z:-N0XCmqp5N-AQQYzvq5-:bgfhjar) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/42c379cdb299a417036b1c5a082db3bc86666a69/internal/source/mylogical/conn.go#L160)):*
> ```Go
> 				// and the smallest nonzero transaction sequence number not yet used on this server (GNO)
> 				log.Tracef("GTIDEvent %d \n", e.GNO)
> 				c.lastGTIDEvent = e.GNO
> ```

The GNO is only unique from a specific source id, so I think you need to convert this to an actual GTID.

Can't the GTIDEvent trigger the call to `OnBegin()`?  If statement replication isn't enabled, then you aren't guaranteed to see the `QueryEvent` below.
___
*[`internal/util/stamp/stamp.go` line 23 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0XDypa6eCA7pUHbx84:-N0XDypa6eCA7pUHbx85:bl9fuo4) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/42c379cdb299a417036b1c5a082db3bc86666a69/internal/util/stamp/stamp.go#L23)):*
> ```Go
> 	// Stamp.
> 	Less(other Stamp) bool
> 	Marshall() string
> ```

If this is really necessary to add to `Stamp`, embed the `encoding.TextMarshaler` interface instead.
___
*[`internal/util/stamp/stamp.go` line 24 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0XDlAy-EZr4zjYdyPR:-N0XDlAy-EZr4zjYdyPS:b2hanu2) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/42c379cdb299a417036b1c5a082db3bc86666a69/internal/util/stamp/stamp.go#L24)):*
> ```Go
> 	Less(other Stamp) bool
> 	Marshall() string
> 	Unmarshall(string) (Stamp, error)
> ```

Shouldn't `Unmarshall()` be part of `logical.Dialect` a-la `Dialect.UnmarshalStamp()`?

sravotto (commented):
Made changes to move the consistent point impl within the logical loop.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-:-N0lJPbN857wQbJigNiX:bu34hnx)* status: 0 of 29 files reviewed, 14 unresolved discussions (waiting on @bobvawter and @BramGruneir)

___
*[`-- commits` line 3 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0X7f1C31ljjXvzOk4Y):*

This base commit is a bit stale.  The current one is rebased onto master after merging PR #114 .  I had found and fixed the same race condition that you addressed, so you can unwind at least some of your test changes in the logical package.
</blockquote></details>

Done
___
*[`.github/docker-compose.yml` line 24 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0X89FoC2QSfhD7Vjvu:-N0lHy-g2577VTscJugk:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/62299ff7c1c7dae774202ee1d9738e76a51789bd/.github/docker-compose.yml#L24)):*

Please keep the Debian base image.
</blockquote></details>

Done.
___
*[`internal/cmd/mylogical/mylogical.go` line 75 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0X8l1Y9ftDq2-OoDeH:-N0lI-f-BFo5rBUWrWOO:b32w0x) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/42c379cdb299a417036b1c5a082db3bc86666a69/internal/cmd/mylogical/mylogical.go#L75)):*

I'll add a todo item to extract the common parts of the logical subcommands into a support package.
</blockquote></details>

OK
___
*[`internal/source/logical/dialect.go` line 81 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0XANI62FB9deqbafWE:-N0lI7-t-aiDnj5m72uI:b363lxw) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/42c379cdb299a417036b1c5a082db3bc86666a69/internal/source/logical/dialect.go#L81)):*

Why does this API exist?  If the consistent point needs to be tracked by the logical loop, that should happen without the involvement of the Dialect.
</blockquote></details>

Done - moved the code within the logical loop
___
*[`internal/source/mylogical/config.go` line 84 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0XBJ-D5JbvS1WzMMkJ:-N0lIKit7M-JtgQgdN-F:bzbpaj8) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/42c379cdb299a417036b1c5a082db3bc86666a69/internal/source/mylogical/config.go#L84)):*

What does this API do differently from the regular `url.Parse`?
</blockquote></details>

Done. Using url.Parse intead.
___
*[`internal/source/mylogical/conn.go` line 37 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0XBUS426YXWa-IQAda:-N0lIQ3U6JJpoIBWOrxy:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/42c379cdb299a417036b1c5a082db3bc86666a69/internal/source/mylogical/conn.go#L37)):*

I think the base type needs to be a GTIDSet, not just the sequence number, since the target mysql instance could be aggregating changes from multiple upstream servers.  Either way, the type probably needs a name other than `lsn`, since that's a postgres-ism.
</blockquote></details>

Done.
___
*[`internal/source/mylogical/conn.go` line 48 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0XBqZ9-64sCvdEJj3-:-N0lJH4H1rL9qNSfg1k7:b-8l9sc2) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/42c379cdb299a417036b1c5a082db3bc86666a69/internal/source/mylogical/conn.go#L48)):*

The canonical interface for this kind of behavior is https://pkg.go.dev/encoding#TextMarshaler

IIRC, implementing that interface makes a type directly usable by the `db/sql` and `pgx` packages as a parameter value.
</blockquote></details>

Using fmt.Stringer.
___
*[`internal/source/mylogical/conn.go` line 62 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0XCE6vDSxve98HP_YI:-N0lISM11wkPo1I0rrHr:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/42c379cdb299a417036b1c5a082db3bc86666a69/internal/source/mylogical/conn.go#L62)):*

This should probably be an appropriately-typed `Stamp`, not a raw number.
</blockquote></details>

Done.
___
*[`internal/source/mylogical/conn.go` line 81 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0XCNBx4JyqNM97TxkJ:-N0lIU5E1hks97w8e63f:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/42c379cdb299a417036b1c5a082db3bc86666a69/internal/source/mylogical/conn.go#L81)):*

Use stringer instead. See https://github.com/cockroachdb/cdc-sink/blob/master/internal/util/ident/ident.go#L50 for an example.  The nice thing about that helper is that the generate code will fail to compile if it's ever out of date.
</blockquote></details>

Done.
___
*[`internal/source/mylogical/conn.go` line 87 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0XCiZd-ehI9lMoCjz8:-N0lIVEA-KIoOxZu0eDE:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/42c379cdb299a417036b1c5a082db3bc86666a69/internal/source/mylogical/conn.go#L87)):*

s/pglogical/mysql/
</blockquote></details>

Done.
___
*[`internal/source/mylogical/conn.go` line 160 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0XCmqp5N-AQQYzvq4z:-N0lI_rK4rAwg7lc8V5-:bmw3168) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/42c379cdb299a417036b1c5a082db3bc86666a69/internal/source/mylogical/conn.go#L160)):*

The GNO is only unique from a specific source id, so I think you need to convert this to an actual GTID.

Can't the GTIDEvent trigger the call to `OnBegin()`?  If statement replication isn't enabled, then you aren't guaranteed to see the `QueryEvent` below.
</blockquote></details>

It seems that we do need to check for BEGIN - See https://dev.mysql.com/blog-archive/taking-advantage-of-new-transaction-length-metadata/

I did convert the GNO to a GTID
___
*[`internal/util/stamp/stamp.go` line 23 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0XDypa6eCA7pUHbx84:-N0lJ00aDXvsuCiEWsNx:bnsr1pw) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/42c379cdb299a417036b1c5a082db3bc86666a69/internal/util/stamp/stamp.go#L23)):*

If this is really necessary to add to `Stamp`, embed the `encoding.TextMarshaler` interface instead.
</blockquote></details>

Decided to use fmt.Stringer
___
*[`internal/util/stamp/stamp.go` line 24 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0XDlAy-EZr4zjYdyPR:-N0lJ8xY04cTA5oVe5SV:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/42c379cdb299a417036b1c5a082db3bc86666a69/internal/util/stamp/stamp.go#L24)):*

Shouldn't `Unmarshall()` be part of `logical.Dialect` a-la `Dialect.UnmarshalStamp()`? 
</blockquote></details>

Done.
___
*[`internal/source/logical/checkpoint.go` line 26 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0X95TN8-0Kh-FbyLcj:-N0lJDIIBJW2bpPb2ixc:bb6un1j) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/42c379cdb299a417036b1c5a082db3bc86666a69/internal/source/logical/checkpoint.go#L26)):*

I'm trying to keep any code which executes sql commands within the `target` or `util` packages.  This type is a pretty generic string-string mapping, maybe call it `Memo` instead?  It should also have its own test cases as well.
</blockquote></details>

Done

bobvawter (changes requested):
> Reviewed 2 of 25 files at r4, 27 of 27 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/116)* status: all files reviewed, 41 unresolved discussions (waiting on @BramGruneir and @sravotto)

___
*[`internal/source/logical/dialect.go` line 43 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pZUh06GX_EJVXaxGs:-N0pZUh06GX_EJVXaxGt:b4b0xfz) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/logical/dialect.go#L43)):*
> ```Go
> 
> 	// UnmarshalStamp decodes a stamp in string format to a stamp.Stamp.
> 	UnmarshalStamp(stamp string) (stamp.Stamp, error)
> ```

The canonical datatype for serialized data is `[]byte`.
___
*[`internal/source/logical/logical.go` line 44 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pZkcQCDAuMBDMUrHS:-N0pZkcQCDAuMBDMUrHT:b-1yaz4i) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/logical/logical.go#L44)):*
> ```Go
> type impl struct {
> 	// Optional checkpoint saved into the target database
> 	memo *memo.Memo
> ```

nit: Alphabetize members.
___
*[`internal/source/logical/logical.go` line 80 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pZrG87K7ejZMwF2Kt:-N0pZrG87K7ejZMwF2Ku:b-yc0q6w) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/logical/logical.go#L80)):*
> ```Go
> 
> var (
> 	standbyTimeout time.Duration = time.Second * 5
> ```

nits: this is usually written `x * duration`. The type declarations are also redundant, since you're not asserting a variable type that's different from the expression type.  Alphabetize as well.
___
*[`internal/source/logical/logical.go` line 137 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0p_moR7AfNHDrjOXuE:-N0p_moR7AfNHDrjOXuF:b-i9gbkl) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/logical/logical.go#L137)):*
> ```Go
> 		targetDB:   config.TargetDB,
> 		memo:       m,
> 		sourceID:   config.ConsistentPointKey,
> ```

`sourceID` vs `ConsistentPointKey` seems mis-named here.
___
*[`internal/source/logical/logical.go` line 141 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0p_gHkAdsjjHHaEbJF:-N0p_gHkAdsjjHHaEbJG:bddmytq) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/logical/logical.go#L141)):*
> ```Go
> 			sync.Mutex
> 			consistentPoint stamp.Stamp
> 		}{},
> ```

This `mu` assignment is a no-op.
___
*[`internal/source/logical/logical.go` line 145 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pa2l3-tEb-4Dx1OPB:-N0pa2l3-tEb-4Dx1OPC:b-srgw4z) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/logical/logical.go#L145)):*
> ```Go
> 	}
> 
> 	if cp != "" {
> ```

Move the `if config.ConsistentPointKey != ""` block down here, to reduce the total number of conditional blocks to have to think about.
___
*[`internal/source/logical/logical.go` line 196 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0paFWG68vdl5Duyzqk:-N0paFWG68vdl5Duyzql:bf70gdj) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/logical/logical.go#L196)):*
> ```Go
> }
> 
> func retrieveConsistentPoint(
> ```

This seems like it should also be a member method.
___
*[`internal/source/logical/logical.go` line 272 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0paYd340gd7cpVwg8J:-N0paYd340gd7cpVwg8K:b-dhfot4) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/logical/logical.go#L272)):*
> ```Go
> 		err = l.saveConsistentPoint(ctx)
> 		if err != nil {
> 			return errors.WithStack(err)
> ```

I've been using `errors.WithStack()` only on calls to external modules and assuming that any error coming from code within cdc-sink is already usefully annotated.
___
*[`internal/source/mylogical/config.go` line 43 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pbCQH30JTwtQDHHDU:-N0pbCQI-Kyw0CAAOvsh:b-tcvwpj) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/config.go#L43)):*
> ```Go
> 	caPem, certPem, keyPem []byte, insecureSkipVerify bool, serverName string,
> ) (*tls.Config, error) {
> 	pool := x509.NewCertPool()
> ```

Consider initializing the pool from `x509.SystemCertPool()`?
___
*[`internal/source/mylogical/config.go` line 93 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pbR5E37yqbyxKctIu:-N0pbR5E37yqbyxKctIv:bfbg7vz) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/config.go#L93)):*
> ```Go
> 	params := u.Query()
> 	sslmode := params.Get("sslmode")
> 	// TODO add Tlssupport
> ```

Isn't this done?
___
*[`internal/source/mylogical/config.go` line 98 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pbTly3c4B_yf_cPux:-N0pbTly3c4B_yf_cPuy:b4ofpys) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/config.go#L98)):*
> ```Go
> 	switch sslmode {
> 	case "disable":
> 		tls = nil
> ```

This is a no-op, just add a comment here.
___
*[`internal/source/mylogical/config.go` line 100 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pbqtJ4ptyrukTe08V:-N0pbqtJ4ptyrukTe08W:b-seoqx5) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/config.go#L100)):*
> ```Go
> 		tls = nil
> 	case "require", "verify-ca", "verify-full":
> 		caCert, err := ioutil.ReadFile(params.Get("sslrootcert"))
> ```

This would read better the guts of this case statement were extracted into a separate function or pushed down into `newClientTLSConfig`.
___
*[`internal/source/mylogical/config.go` line 120 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pbXOhEhT7JnKfLeju:-N0pbXOhEhT7JnKfLejv:b-sc6suq) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/config.go#L120)):*
> ```Go
> 		}
> 	default:
> 		return errors.New("invalid sslmode")
> ```

`erorrs.Errorf("invalid sslmode: %q", sslmode)`
___
*[`internal/source/mylogical/config.go` line 126 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pc1s8-g3OwZcQLkZ9:-N0pc1s8-g3OwZcQLkZA:b-65ul8c) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/config.go#L126)):*
> ```Go
> 	cfg := replication.BinlogSyncerConfig{
> 		ServerID:  uint32(serverID),
> 		Flavor:    "mysql",
> ```

Just a heads-up, but we do have an upcoming customer commit to support MariaDB, but let's do that in a separate PR.
___
*[`internal/source/mylogical/config.go` line 134 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pcCDD85Iaa8rHPwCX:-N0pcCDD85Iaa8rHPwCY:bkk7pau) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/config.go#L134)):*
> ```Go
> 	}
> 
> 	c.binlogSyncerConfig = cfg
> ```

No need for the separate local variable.
___
*[`internal/source/mylogical/conn.go` line 48 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0XBqZ9-64sCvdEJj3-:-N0pcPr6BZEg_LjPvAu7:b-tw51e6) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/42c379cdb299a417036b1c5a082db3bc86666a69/internal/source/mylogical/conn.go#L48)):*

Using fmt.Stringer.
</blockquote></details>

`fmt.Stringer` is not the right interface for persistence use.  `Stringer` is intended for producing human-readable representations of the data, which likely differ from how you'd want to persist it.  For simple datatypes, they'll often implement the `encoding.TextMarshaler` API as a delegate to their `fmt.Stringer` implementation.
___
*[`internal/source/mylogical/conn.go` line 86 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pd4MD043UFCBAEie_:-N0pd4MD043UFCBAEiea:b-78vhp5) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/conn.go#L86)):*
> ```Go
> 	sourceConfig replication.BinlogSyncerConfig
> 	// Key to set/retrieve state
> 	memoKey string
> ```

nit: alphabetize
___
*[`internal/source/mylogical/conn.go` line 96 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pg-A260rs9v8MsZfh:-N0pg-A260rs9v8MsZfi:bb9y40v) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/conn.go#L96)):*
> ```Go
> 	update
> 	delete
> 	unknown
> ```

Make `unknown` the `iota` (and therefore default) value for the enum type.  That way, a declaration like `var mt MutationType` would automatically have the `unknown` value.  As it stands, the default value is `insert`.

It's also odd to have `MutationType` exported from the package, while its values are unexported.
___
*[`internal/source/mylogical/conn.go` line 151 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pdG_N17R9oG1xDYhf:-N0pdG_N17R9oG1xDYhg:b-4bl07u) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/conn.go#L151)):*
> ```Go
> 			if err := events.OnRollback(ctx, msg); err != nil {
> 				return err
> 			}
> ```

Add a `continue` here. 
___
*[`internal/source/mylogical/conn.go` line 155 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pdWtl5gnMW_j0OYId:-N0pdWtl5gnMW_j0OYIe:bchz5rw) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/conn.go#L155)):*
> ```Go
> 		var err error
> 		var ev, ok = msg.(replication.BinlogEvent)
> 		if ok {
> ```

Flip this to `if !ok { return errors... }` so as to be able to outdent that large swath of code below.
___
*[`internal/source/mylogical/conn.go` line 182 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pdws12-lOkpT0yIOI:-N0pdws12-lOkpT0yIOJ:b-e331h0) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/conn.go#L182)):*
> ```Go
> 				// and the smallest nonzero transaction sequence number not yet used on this server (GNO)
> 				u, _ := uuid.FromBytes(e.SID)
> 				s := fmt.Sprintf("%s:%d", u.String(), e.GNO)
> ```

Push this formatting into the call to `Tracef` to avoid unnecessary work when tracing isn't enabled.

The calls to `value.String()` when using an `%s` or `%v` directive also create unnecessary work.
___
*[`internal/source/mylogical/conn.go` line 186 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pfcb05yQW39R9vx0A:-N0pfcb05yQW39R9vx0B:b-pnsinf) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/conn.go#L186)):*
> ```Go
> 				a, err := mysql.ParseUUIDSet(s)
> 				if err == nil {
> 					clone := &mysql.MysqlGTIDSet{
> ```

Can you use https://pkg.go.dev/github.com/go-mysql-org/go-mysql/mysql#MysqlGTIDSet.Clone  instead?
___
*[`internal/source/mylogical/conn.go` line 203 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pfKAUB9DQWVELYlhC:-N0pfKAUB9DQWVELYlhD:bumxxex) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/conn.go#L203)):*
> ```Go
> 				}
> 
> 				//e.Dump(os.Stdout)
> ```

nit: Dangling comment.
___
*[`internal/source/mylogical/conn.go` line 216 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pft_T3o12_9T2H1Rh:-N0pft_T3o12_9T2H1Ri:blc4dw1) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/conn.go#L216)):*
> ```Go
> 					operation = insert
> 				default:
> 					operation = unknown
> ```

This could just set `err` and `break` from the `switch` statement.
___
*[`internal/source/mylogical/conn.go` line 238 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pgZwoEGgMvJFDvnj9:-N0pgZwoEGgMvJFDvnjA:b-69siol) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/conn.go#L238)):*
> ```Go
> // ReadInto implements logical.Dialect, opens a replication connection,
> // and writes parsed messages into the provided channel. This method
> // also manages the keepalive protocol.
> ```

Is the comment about keepalive relevant to mysql?
___
*[`internal/source/mylogical/conn.go` line 249 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pgfnV-9e69PTciWAD:-N0pgfnV-9e69PTciWAE:b-j3d3p8) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/conn.go#L249)):*
> ```Go
> 	gtidset, err := mysql.ParseMysqlGTIDSet(state.GetConsistentPoint().String())
> 	if err != nil {
> 		return errors.New("unable to parse gtidset")
> ```

Use `errors.Wrap()` to propagate the underlying error info.
___
*[`internal/source/mylogical/conn.go` line 276 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pgv5E3E0b_lQ0sXOs:-N0pgv5E3E0b_lQ0sXOt:bg8g3xq) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/conn.go#L276)):*
> ```Go
> 			case ch <- *ev:
> 			case <-ctx.Done():
> 				return errors.WithStack(ctx.Err())
> ```

Don't need `WithStack` here, since the error will come from code within cdc-sink.
___
*[`internal/source/mylogical/conn.go` line 292 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0ph2Vv8jOKEktxNI3u:-N0ph2Vv8jOKEktxNI3v:b-203rkj) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/conn.go#L292)):*
> ```Go
> 		default:
> 			log.Warningf("event type %T is not currently supported", ev.Event)
> 			ev.Dump(os.Stdout)
> ```

Remove call to `ev.Dump()` or guard it with a call to `log.IsLevelEnabled()`.
___
*[`internal/source/mylogical/conn.go` line 307 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0phYh68HZ39ZISHPL3:-N0phYh68HZ39ZISHPL4:bkw8bwb) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/conn.go#L307)):*
> ```Go
> 	s, err := mysql.ParseMysqlGTIDSet(stamp)
> 	if err != nil {
> 		return nil, err
> ```

I like to use `errors.Wrapf()` to also provide the user with the value that was passed in.
___
*[`internal/source/mylogical/conn.go` line 324 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0phjl43hDieUurBPqZ:-N0phjl43hDieUurBPq_:b-kx1dxn) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/conn.go#L324)):*
> ```Go
> 		return errors.Errorf("unknown relation id %d", tuple.TableID)
> 	}
> 	if string(tbl.Database().String()) != events.GetTargetDB().String() {
> ```

This string casting shouldn't be necessary, since the `ident` package types are comparable.
___
*[`internal/source/mylogical/conn.go` line 352 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pid1C0-Vv9bsstnFX:-N0pid1C0-Vv9bsstnFY:by54vn8) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/conn.go#L352)):*
> ```Go
> 			return err
> 		}
> 		if operation == update && (rowNum%2 == 0) {
> ```

Add a comment explaining this modular math.  Presumably this is before/after?
___
*[`internal/source/mylogical/conn.go` line 369 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pi2L22Melunk_zgS2:-N0pi2L22Melunk_zgS3:b-ctf4ag) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/conn.go#L369)):*
> ```Go
> }
> 
> // learn updates the source database namespace mappings.
> ```

s/learn/onRelation/
___
*[`internal/source/mylogical/conn.go` line 385 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pi8xp7_gKosbeSSmy:-N0pi8xp7_gKosbeSSmz:bgooaui) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/conn.go#L385)):*
> ```Go
> 	}
> 	if len(msg.ColumnName) != len(msg.ColumnType) {
> 		return errors.Errorf("all columns names are required 'set global binlog_row_metadata = full'")
> ```

There are no formatting directives, so `errors.New()` would be the function to use.
___
*[`internal/source/mylogical/integration_test.go` line 44 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pixpUBgH3a8qqrQEC:-N0pixpUBgH3a8qqrQED:b6e2xh0) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/integration_test.go#L44)):*
> ```Go
> 	sinktest.IntegrationMain(m, sinktest.MySQLName)
> }
> func TestMYLogical(t *testing.T) {
> ```

Add a test in here to verify of the datatypes supported by mysql.  See the code in `pglogical`.  If there are types which cdc-sink/cockroach can't support, that's fine, and it's also a place to document that.
___
*[`internal/target/sinktest/integration.go` line 31 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pjQzR8YI2mm3lprQx:-N0pjQzS6gvvz69wipdN:b-78vhp5) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/target/sinktest/integration.go#L31)):*
> ```Go
> 	PostgreSQLName = "postgresql"
> 	// MySQLName same as above.
> 	MySQLName = "mysql"
> ```

nit: alphabetize
___
*[`internal/util/memo/memo.go` line 25 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pjxJ4EP6_MIcXCsay:-N0pjxJ4EP6_MIcXCsaz:b-8e3k9x) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/util/memo/memo.go#L25)):*
> ```Go
> // Memo is a simple kv store with string keys and string values.
> type Memo struct {
> 	pool pgxtype.Querier
> ```

This should be passed into the `Get` and `Put` methods so that the caller has control over the enclosing transaction (if any).

Also, add an interface to `types.go` and use that in the callers.  This isn't strictly necessary, but I've found that defining the core set of "services" within cdc-sink in a single file makes it easier for folks to understand the big moving pieces.
___
*[`internal/util/memo/memo.go` line 34 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pjWnQ61gMIk8PukSF:-N0pjWnQ61gMIk8PukSG:b-yub0iv) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/util/memo/memo.go#L34)):*
> ```Go
> // NewMemo creates a structure to manage the key value store.
> // It will create the table passed as an argument in the database, if it does not exists.
> func NewMemo(ctx context.Context, tx pgxtype.Querier, target ident.Table) (*Memo, error) {
> ```

This can be called `New`, since the package name doesn't need to be repeated in symbols exported from the package.
___
*[`internal/util/memo/memo.go` line 73 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pkUXR7t-GeHA5csIL:-N0pkUXS6xRPhLz6AmzH:b-cvp0zy) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/util/memo/memo.go#L73)):*
> ```Go
> // If the key doesn't exists, it will return the default value supplied.
> func (m *Memo) Get(ctx context.Context, key string, def string) (string, error) {
> 	s := def
> ```

Is this assignment necessary?
___
*[`internal/util/memo/memo.go` line 80 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pkZIc4ofwHEWykMmp:-N0pkZIc4ofwHEWykMmq:bz1ianm) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/util/memo/memo.go#L80)):*
> ```Go
> 			key,
> 		)
> 		res.Scan(&s)
> ```

This ignores actual errors.  Use `errors.Is(err, pgx.ErrNoRows)` to distinguish the not-found case.
___
*[`internal/util/memo/memo_test.go` line 21 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pktMZ423fkjeRRpx3:-N0pktMZ423fkjeRRpx4:b-mxm4i9) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/util/memo/memo_test.go#L21)):*
> ```Go
> )
> 
> func TestMemo_Roundtrip(t *testing.T) {
> ```

This can be `TestRoundtrip`
___
*[`internal/util/stamp/stamp.go` line 23 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0XDypa6eCA7pUHbx84:-N0pl3rH-IdmbyQl9laC:bhip2s) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/42c379cdb299a417036b1c5a082db3bc86666a69/internal/util/stamp/stamp.go#L23)):*

Decided to use fmt.Stringer
</blockquote></details>

See other comment about `fmt.Stringer` not being appropriate for persistence.

sravotto (commented):
Thanks for the review. I made additional changes, and added a test for mysql types.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-:-N14Iyae5A_G1tin00Ra:bly1dwg)* status: 11 of 30 files reviewed, 37 unresolved discussions (waiting on @bobvawter and @BramGruneir)

___
*[`internal/source/logical/dialect.go` line 43 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pZUh06GX_EJVXaxGs:-N14GMc_2LCE0ybdCCl7:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/logical/dialect.go#L43)):*

The canonical datatype for serialized data is `[]byte`.
</blockquote></details>

Done.
___
*[`internal/source/logical/logical.go` line 80 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pZrG87K7ejZMwF2Kt:-N14GRobEWk3vNBn9UAP:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/logical/logical.go#L80)):*

nits: this is usually written `x * duration`. The type declarations are also redundant, since you're not asserting a variable type that's different from the expression type.  Alphabetize as well.
</blockquote></details>

Done.
___
*[`internal/source/logical/logical.go` line 137 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0p_moR7AfNHDrjOXuE:-N14GVnc3zVatPSAdQCx:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/logical/logical.go#L137)):*

`sourceID` vs `ConsistentPointKey` seems mis-named here.
</blockquote></details>

Done.
___
*[`internal/source/logical/logical.go` line 141 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0p_gHkAdsjjHHaEbJF:-N14GXAI-pPrzIMtX9f-:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/logical/logical.go#L141)):*

This `mu` assignment is a no-op.
</blockquote></details>

Done.
___
*[`internal/source/logical/logical.go` line 145 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pa2l3-tEb-4Dx1OPB:-N14Iq5E9-Vr1Cwo2gWK:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/logical/logical.go#L145)):*

Move the `if config.ConsistentPointKey != ""` block down here, to reduce the total number of conditional blocks to have to think about.
</blockquote></details>

Done.
___
*[`internal/source/logical/logical.go` line 196 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0paFWG68vdl5Duyzqk:-N14Gk1z09gph-3xyIhi:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/logical/logical.go#L196)):*

This seems like it should also be a member method.
</blockquote></details>

Done.
___
*[`internal/source/logical/logical.go` line 272 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0paYd340gd7cpVwg8J:-N14Gm562SE-cKZmdzUf:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/logical/logical.go#L272)):*

I've been using `errors.WithStack()` only on calls to external modules and assuming that any error coming from code within cdc-sink is already usefully annotated.
</blockquote></details>

Done.
___
*[`internal/source/mylogical/config.go` line 43 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pbCQH30JTwtQDHHDU:-N14GonY0dpzFtI7HVYk:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/config.go#L43)):*

Consider initializing the pool from `x509.SystemCertPool()`?
</blockquote></details>

Done.
___
*[`internal/source/mylogical/config.go` line 93 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pbR5E37yqbyxKctIu:-N14Gpxl1KmDpwao_vWb:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/config.go#L93)):*

Isn't this done?
</blockquote></details>

Done.
___
*[`internal/source/mylogical/config.go` line 98 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pbTly3c4B_yf_cPux:-N14Gqy5DNvcGsXmTg67:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/config.go#L98)):*

This is a no-op, just add a comment here.
</blockquote></details>

Done.
___
*[`internal/source/mylogical/config.go` line 100 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pbqtJ4ptyrukTe08V:-N14Grnj-29IvFokM7c2:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/config.go#L100)):*

This would read better the guts of this case statement were extracted into a separate function or pushed down into `newClientTLSConfig`.
</blockquote></details>

Done.
___
*[`internal/source/mylogical/config.go` line 120 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pbXOhEhT7JnKfLeju:-N14Gt631gnHM2zxfxq7:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/config.go#L120)):*

`erorrs.Errorf("invalid sslmode: %q", sslmode)`
</blockquote></details>

Done.
___
*[`internal/source/mylogical/config.go` line 126 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pc1s8-g3OwZcQLkZ9:-N14GuMu3gixF9abgyrr:b-k7huu1) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/config.go#L126)):*

Just a heads-up, but we do have an upcoming customer commit to support MariaDB, but let's do that in a separate PR.
</blockquote></details>

Ok - Sounds good.
___
*[`internal/source/mylogical/config.go` line 134 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pcCDD85Iaa8rHPwCX:-N14H2r8EmWd56uROwJp:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/config.go#L134)):*

No need for the separate local variable.
</blockquote></details>

Done.
___
*[`internal/source/mylogical/conn.go` line 48 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0XBqZ9-64sCvdEJj3-:-N14H5g7DgXt0gz18Qdl:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/42c379cdb299a417036b1c5a082db3bc86666a69/internal/source/mylogical/conn.go#L48)):*

`fmt.Stringer` is not the right interface for persistence use.  `Stringer` is intended for producing human-readable representations of the data, which likely differ from how you'd want to persist it.  For simple datatypes, they'll often implement the `encoding.TextMarshaler` API as a delegate to their `fmt.Stringer` implementation.
</blockquote></details>

Done.
___
*[`internal/source/mylogical/conn.go` line 96 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pg-A260rs9v8MsZfh:-N14H8Rx7HObBvS6daa0:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/conn.go#L96)):*

Make `unknown` the `iota` (and therefore default) value for the enum type.  That way, a declaration like `var mt MutationType` would automatically have the `unknown` value.  As it stands, the default value is `insert`.

It's also odd to have `MutationType` exported from the package, while its values are unexported.
</blockquote></details>

Done.
___
*[`internal/source/mylogical/conn.go` line 151 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pdG_N17R9oG1xDYhf:-N14H9NK72ol1rEKusKq:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/conn.go#L151)):*

Add a `continue` here. 
</blockquote></details>

Done.
___
*[`internal/source/mylogical/conn.go` line 155 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pdWtl5gnMW_j0OYId:-N14HCqTDA205lSRStvs:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/conn.go#L155)):*

Flip this to `if !ok { return errors... }` so as to be able to outdent that large swath of code below.
</blockquote></details>

Done.
___
*[`internal/source/mylogical/conn.go` line 182 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pdws12-lOkpT0yIOI:-N14HJv2B7T31smDSop-:b-g5s9su) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/conn.go#L182)):*

Push this formatting into the call to `Tracef` to avoid unnecessary work when tracing isn't enabled.

The calls to `value.String()` when using an `%s` or `%v` directive also create unnecessary work.
</blockquote></details>

I do need the string s to build a GTIDSet to add to the current set. I did remove the trace though.
___
*[`internal/source/mylogical/conn.go` line 186 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pfcb05yQW39R9vx0A:-N14HgKb-zSsv4Os2Wdo:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/conn.go#L186)):*

Can you use https://pkg.go.dev/github.com/go-mysql-org/go-mysql/mysql#MysqlGTIDSet.Clone  instead?
</blockquote></details>

Done.
___
*[`internal/source/mylogical/conn.go` line 216 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pft_T3o12_9T2H1Rh:-N14HjT7AlCwogV1-ABO:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/conn.go#L216)):*

This could just set `err` and `break` from the `switch` statement.
</blockquote></details>

Done.
___
*[`internal/source/mylogical/conn.go` line 238 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pgZwoEGgMvJFDvnj9:-N14HkqRClgs_XpBDpVK:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/conn.go#L238)):*

Is the comment about keepalive relevant to mysql?
</blockquote></details>

Done.
___
*[`internal/source/mylogical/conn.go` line 249 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pgfnV-9e69PTciWAD:-N14HlZF7oL5JFe-Er0l:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/conn.go#L249)):*

Use `errors.Wrap()` to propagate the underlying error info.
</blockquote></details>

Done.
___
*[`internal/source/mylogical/conn.go` line 276 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pgv5E3E0b_lQ0sXOs:-N14HmMY78YtVZyusmDr:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/conn.go#L276)):*

Don't need `WithStack` here, since the error will come from code within cdc-sink.
</blockquote></details>

Done.
___
*[`internal/source/mylogical/conn.go` line 292 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0ph2Vv8jOKEktxNI3u:-N14Hnwr9RUf11fHMI2V:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/conn.go#L292)):*

Remove call to `ev.Dump()` or guard it with a call to `log.IsLevelEnabled()`.
</blockquote></details>

Done.
___
*[`internal/source/mylogical/conn.go` line 307 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0phYh68HZ39ZISHPL3:-N14Hovc3SXf60Bcnfk-:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/conn.go#L307)):*

I like to use `errors.Wrapf()` to also provide the user with the value that was passed in.
</blockquote></details>

Done.
___
*[`internal/source/mylogical/conn.go` line 324 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0phjl43hDieUurBPqZ:-N14Hq4_AFOsNxAOFJPH:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/conn.go#L324)):*

This string casting shouldn't be necessary, since the `ident` package types are comparable.
</blockquote></details>

Done.
___
*[`internal/source/mylogical/conn.go` line 352 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pid1C0-Vv9bsstnFX:-N14HrX13WxiULrYNgyE:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/conn.go#L352)):*

Add a comment explaining this modular math.  Presumably this is before/after?
</blockquote></details>

Done.
___
*[`internal/source/mylogical/conn.go` line 369 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pi2L22Melunk_zgS2:-N14HsD7B4U9WGaaxRMQ:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/conn.go#L369)):*

s/learn/onRelation/
</blockquote></details>

Done.
___
*[`internal/source/mylogical/conn.go` line 385 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pi8xp7_gKosbeSSmy:-N14HtJs0DN9tfkEt01Q:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/conn.go#L385)):*

There are no formatting directives, so `errors.New()` would be the function to use.
</blockquote></details>

Done.
___
*[`internal/source/mylogical/integration_test.go` line 44 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pixpUBgH3a8qqrQEC:-N14HvAc5aNEoRkoQ4sj:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/source/mylogical/integration_test.go#L44)):*

Add a test in here to verify of the datatypes supported by mysql.  See the code in `pglogical`.  If there are types which cdc-sink/cockroach can't support, that's fine, and it's also a place to document that.
</blockquote></details>

Done.
___
*[`internal/util/memo/memo.go` line 25 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pjxJ4EP6_MIcXCsay:-N14I1Z74uSUzOdz2SN-:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/util/memo/memo.go#L25)):*

This should be passed into the `Get` and `Put` methods so that the caller has control over the enclosing transaction (if any).

Also, add an interface to `types.go` and use that in the callers.  This isn't strictly necessary, but I've found that defining the core set of "services" within cdc-sink in a single file makes it easier for folks to understand the big moving pieces.
</blockquote></details>

Done.
___
*[`internal/util/memo/memo.go` line 34 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pjWnQ61gMIk8PukSF:-N14IsyX2at77sdnh32l:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/util/memo/memo.go#L34)):*

This can be called `New`, since the package name doesn't need to be repeated in symbols exported from the package.
</blockquote></details>

Done.
___
*[`internal/util/memo/memo.go` line 73 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pkUXR7t-GeHA5csIL:-N14I4CQDbABJr4XxYFh:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/util/memo/memo.go#L73)):*

Is this assignment necessary?
</blockquote></details>

Done.
___
*[`internal/util/memo/memo.go` line 80 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pkZIc4ofwHEWykMmp:-N14I55hDS4_4P7-A3-N:b-yvsqm8) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/util/memo/memo.go#L80)):*

This ignores actual errors.  Use `errors.Is(err, pgx.ErrNoRows)` to distinguish the not-found case.
</blockquote></details>

Done. Adding error logic
___
*[`internal/util/memo/memo_test.go` line 21 at r5](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0pktMZ423fkjeRRpx3:-N14IA-m1QjhLIUYKaoL:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/56f18f1d9fe7d8468724854c72a397100da6cab4/internal/util/memo/memo_test.go#L21)):*

This can be `TestRoundtrip`
</blockquote></details>

Done.
___
*[`internal/util/stamp/stamp.go` line 23 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N0XDypa6eCA7pUHbx84:-N14ICPB0GPgKStTDky9:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/42c379cdb299a417036b1c5a082db3bc86666a69/internal/util/stamp/stamp.go#L23)):*

See other comment about `fmt.Stringer` not being appropriate for persistence.
</blockquote></details>

Done.

bobvawter (approved):
:shipit: Looks good, let's go ahead and commit this.  The next point of work will be to add MariaDB support to the code.

> Reviewed 19 of 19 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-:-N15HVSn2_Bp2P4FwUe8:b-b1fvs6)* status: all files reviewed, 8 unresolved discussions (waiting on @BramGruneir and @sravotto)

___
*[`internal/source/logical/logical.go` line 63 at r6](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N159Pkf66uexab6GTHT:-N159Pkf66uexab6GTHU:bkc4pzb) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f8a2177bb15ea3854f750b49bae89bac7431e7c8/internal/source/logical/logical.go#L63)):*
> ```Go
> 	serializer *serial.Pool
> 	// the key used to persist the consistentPoint stamp.
> 	consistentPointKey string
> ```

nit: Alphabetize
___
*[`internal/source/logical/logical.go` line 67 at r6](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N159RN_8LXMHGpp3rjC:-N159RN_8LXMHGpp3rjD:bx8osqg) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f8a2177bb15ea3854f750b49bae89bac7431e7c8/internal/source/logical/logical.go#L67)):*
> ```Go
> 	targetDB ident.Ident
> 	// targetPool
> 	targetPool pgxtype.Querier
> ```

nit: Update the comment what the pool is used for.
___
*[`internal/source/logical/logical.go` line 78 at r6](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N159a9u0cmieFNzPB4P:-N159a9u0cmieFNzPB4Q:b-q3fhjn) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f8a2177bb15ea3854f750b49bae89bac7431e7c8/internal/source/logical/logical.go#L78)):*
> ```Go
> 	}
> 
> 	standbyDeadline time.Time
> ```

nit: Add a comment and re-order the field.
___
*[`internal/source/logical/logical.go` line 184 at r6](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N15AIlQ2qqLLJ-XO344:-N15AIlRDwtSK3HFbrah:b-w77m5t) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f8a2177bb15ea3854f750b49bae89bac7431e7c8/internal/source/logical/logical.go#L184)):*
> ```Go
> }
> 
> func (l *impl) saveConsistentPoint(ctx context.Context, tx pgxtype.Querier) error {
> ```

Passing the querier into `save` and `retrieve` is redundant if they're only ever receiving `l.targetPool`.
___
*[`internal/source/mylogical/conn.go` line 102 at r6](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N15CbuIAJno7mgeRVKA:-N15CbuIAJno7mgeRVKB:b-8clm0y) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f8a2177bb15ea3854f750b49bae89bac7431e7c8/internal/source/mylogical/conn.go#L102)):*
> ```Go
> 	insert
> 	update
> 	delete
> ```

This declaration would shadow the language's built-in `delete` function, used to remove elements from a map.  It's probably worth renaming these declarations to be `xyzMutation` to avoid naming conflicts elsewhere.
___
*[`internal/source/mylogical/conn.go` line 343 at r6](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N15El-MEVcKFWQ6EGyC:-N15El-MEVcKFWQ6EGyD:behwtyc) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f8a2177bb15ea3854f750b49bae89bac7431e7c8/internal/source/mylogical/conn.go#L343)):*
> ```Go
> 		for idx, sourceCol := range row {
> 			targetCol := targetCols[idx]
> 			switch targetCol.Type {
> ```

Switching on `targetCol.Type` is ok, but would it make more sense to switch instead on sourceCol's type? This would reduce the amount of code duplication below.

```
switch v := sourceCol.type {
  case nil:
  case []byte:
  case int64:
  default:
}
```

Then, in the `[]byte` and `int64` case, add a sanity check on `sourceCol.Type` to return an error if there's an unexpected value.
___
*[`internal/source/mylogical/conn.go` line 377 at r6](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N15EOd-7maBirGzvdwp:-N15EOd02fOlX8xghd0I:b-jwl3fw) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f8a2177bb15ea3854f750b49bae89bac7431e7c8/internal/source/mylogical/conn.go#L377)):*
> ```Go
> 		// on update we only care about the new value.
> 		// even rows are skipped since they contain the value before the update
> 		if operation == update && (rowNum%2 == 0) {
> ```

Can this check move the the top loop, to avoid doing all of that decoding, only to throw away the work?
___
*[`internal/types/types.go` line 103 at r6](https://reviewable.io/reviews/cockroachdb/cdc-sink/116#-N15GJAz6H8ZIMcUGWV5:-N15GJAz6H8ZIMcUGWV6:b-i71jfp) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f8a2177bb15ea3854f750b49bae89bac7431e7c8/internal/types/types.go#L103)):*
> ```Go
> 	Name    ident.Ident
> 	Primary bool
> 	Type    interface{}
> ```

Why is this `interface{}` ?  Document the types that might be expected.

PR #143, opened 2022-05-26
apply: Fix unparseable DECIMAL representations.

bobvawter:
The pgx library doesn't like the scientific notation (e.g. "4E+2") that
Cockroach can return for DECIMAL-type columns. This patch looks for these cases
and expands them via Cockroach's own APD library.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/143)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/cockroachdb/cdc-sink
 go 1.17
 
 require (
+	github.com/cockroachdb/apd v1.1.0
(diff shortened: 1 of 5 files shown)


bobvawter:
I found this issue while trying to run the tpcc workload in a cockroach-to-cockroach setup.

BramGruneir:
Does removeexponent need an explicit test?  Looks like it couldn't hurt.

Otherwise LGTM

bobvawter:
Added a unit test for `removeExponent()`.

PR #144, opened 2022-05-31
apply: De-duplicate applied mutations

bobvawter:
Note to reviewers, this includes PR #145 

This change works around a CockroachDB implementation quirk wherein an UPSERT
is unable to read its own writes, which prevents the same row from being
upserted twice within the same statement. This also corrects a potential source
of data inconsistency in the applier code, if a single call to Apply contains
both upserts and deletes of the same row.

Both of the above issues can be addressed by sorting the mutations according to
their HLC time and then de-duplicating them by primary key.

The newly-added test in the apply package checks for the specific upsert
behavior, so that this workaround can be reconsidered if a future version of
CockroachDB allows duplicate rows within an UPSERT or CTE.

X-Ref: https://github.com/cockroachdb/cockroach/issues/44466
X-Ref: https://github.com/cockroachdb/cockroach/issues/70731
X-Ref: https://github.com/cockroachdb/cockroach/pull/45372

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/144)

### DIFF
--- a/internal/target/apply/apply.go
+++ b/internal/target/apply/apply.go
@@ -25,6 +25,7 @@ import (
 	"github.com/cockroachdb/cdc-sink/internal/util/batches"
 	"github.com/cockroachdb/cdc-sink/internal/util/ident"
 	"github.com/cockroachdb/cdc-sink/internal/util/metrics"
+	"github.com/cockroachdb/cdc-sink/internal/util/msort"
 	"github.com/jackc/pgtype/pgxtype"
 	"github.com/pkg/errors"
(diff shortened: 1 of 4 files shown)


BramGruneir (approved):
:lgtm:

> Reviewed 6 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/144#-:-N3tQz2s8P-sTjbfXTk-:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sravotto)

PR #151, opened 2022-06-09
leases: Add leases package

bobvawter:
This change adds a new API for running exactly one instance of some named,
global task.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/151)

### DIFF
--- a/internal/target/leases/leases.go
+++ b/internal/target/leases/leases.go
@@ -1,0 +1,393 @@
+// Copyright 2022 The Cockroach Authors.
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+// Package leases coordinates global, singleton activities.
+package leases
+
+import (
+	"context"
+	"fmt"
+	"math/rand"
+	"time"
+
+	"github.com/cockroachdb/cdc-sink/internal/types"
+	"github.com/cockroachdb/cdc-sink/internal/util/ident"
+	"github.com/cockroachdb/cdc-sink/internal/util/retry"
+	"github.com/google/uuid"
+	"github.com/jackc/pgtype/pgxtype"
+	"github.com/jackc/pgx/v4"
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+)
+
+// Table contains a default table name for managing leases.
+var Table = ident.NewTable(ident.StagingDB, ident.Public, ident.New("leases"))
+
+const (
+	defaultLifetime = time.Minute
+	defaultPoll     = time.Second
+	defaultRetry    = time.Second
+)
+
+// Config is passed to New.
+type Config struct {
+	Pool   pgxtype.Querier // Database access.
+	Target ident.Table     // The lease table.
+
+	// Guard provides a quiescent period between when a lease is
+	// considered to be expired (i.e. a lease callback's context is
+	// canceled) and its published expiration time. This is useful for
+	// situations where it may not be possible to immediately cancel all
+	// side effects of a lease callback (e.g. it makes an external
+	// network request).
+	Guard time.Duration
+
+	Lifetime   time.Duration // Duration of the lease.
+	Poll       time.Duration // How often to re-check for an available lease.
+	RetryDelay time.Duration // Delay between re-executing a callback.
+}
+
+// sanitize checks for mis-configuration and applies sane defaults.
+func (c *Config) sanitize() error {
+	if c.Pool == nil {
+		return errors.New("pool must not be nil")
+	}
+	if c.Target == (ident.Table{}) {
+		return errors.New("target must be set")
+	}
+	// OK for Guard to be zero.
+	if c.Lifetime == 0 {
+		c.Lifetime = defaultLifetime
+	}
+	if c.Poll == 0 {
+		c.Poll = defaultPoll
+	}
+	if c.RetryDelay == 0 {
+		c.RetryDelay = defaultRetry
+	}
+	return nil
+}
+
+type lease struct {
+	expires time.Time
+	name    string
+	nonce   uuid.UUID
+}
+
+// leases coordinates global, singleton activities.
+type leases struct {
+	cfg Config
+	sql struct {
+		acquire string
+		release string
+		renew   string
+	}
+}
+
+var _ types.Leases = (*leases)(nil)
+
+const (
+	schema = `
+CREATE TABLE IF NOT EXISTS %s (
+  name STRING PRIMARY KEY,
+  expires TIMESTAMP NOT NULL,
+  nonce UUID NOT NULL
+)`
+)
+
+// New constructs an instance of types.Leases.
+func New(ctx context.Context, cfg Config) (types.Leases, error) {
+	if err := cfg.sanitize(); err != nil {
+		return nil, err
+	}
+	_, err := cfg.Pool.Exec(ctx, fmt.Sprintf(schema, cfg.Target))
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+
+	l := &leases{cfg: cfg}
+	l.sql.acquire = fmt.Sprintf(acquireTemplate, cfg.Target)
+	l.sql.release = fmt.Sprintf(releaseTemplate, cfg.Target)
+	l.sql.renew = fmt.Sprintf(renewTemplate, cfg.Target)
+
+	return l, nil
+}
+
+// Singleton executes a callback when the named lease is acquired.
+//
+// The lease will be released in the following circumstances:
+//   * The callback returns nil.
+//   * The lease cannot be renewed before it expires.
+//   * The outer context is canceled.
+//
+// If the callback returns a non-nil error, the error will be logged. If
+// the callback returns ErrCancelSingleton, it will not be retried. In
+// all other cases, the callback function is retried once a lease is
+// re-acquired.
+func (l *leases) Singleton(ctx context.Context, name string, fn func(ctx context.Context) error) {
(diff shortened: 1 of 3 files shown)


BramGruneir (changes requested):
I think you should consider pulling this out into a separate github project.  It could stand on its own. 

Just a thought, but could you use select for updates to just hold a lease and have the others waiting?  Not sure how long crdb will hold an open transaction with a select for update.  

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/151#-:-N4DHy7X7W0jkfyBPVt4:badlm4)* status: all files reviewed, 6 unresolved discussions (waiting on @bobvawter)

___
*[`internal/target/leases/leases.go` line 160 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/151#-N48ARAuD7bcfz0-ubdO:-N48ARAuD7bcfz0-ubdP:bq8g3xn) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/dbf3d7813e517f2dc170a3d6bb8d5df662a65d2b/internal/target/leases/leases.go#L160)):*
> ```Go
> 			entry.Trace("context canceled before acquisition")
> 			return lease{}, false
> 		case <-time.After(l.cfg.Poll):
> ```

Should this wait until the known lease has expired instead of just polling?
___
*[`internal/target/leases/leases.go` line 174 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/151#-N4DIa6o-gzzh8HZuE-Z:-N4DIa6o-gzzh8HZuE-_:bogog4x) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/dbf3d7813e517f2dc170a3d6bb8d5df662a65d2b/internal/target/leases/leases.go#L174)):*
> ```Go
> 		// Wait up to half of the expiration time. Subtracting the
> 		// guard time may make the delay negative, so we expire.
> 		delay := tgt.expires.Sub(now)/2 - l.cfg.GuardTime
> ```

So you have expire time, and guard time.  Why not just expire-guard?  Or perhaps set guard as a percentage of total lease time?
___
*[`internal/target/leases/leases.go` line 176 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/151#-N4DJ6R81LrWknaN_MNf:-N4DJ6R81LrWknaN_MNg:bnh5sbc) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/dbf3d7813e517f2dc170a3d6bb8d5df662a65d2b/internal/target/leases/leases.go#L176)):*
> ```Go
> 		delay := tgt.expires.Sub(now)/2 - l.cfg.GuardTime
> 		if delay <= 0 {
> 			return tgt
> ```

Is the lease considered expired here just due to the time being after the delay?  Won't it still appear valid if you're within the guard time?  I see the conditions mentioned in the comment above, but there is ambiguity here.  Or perhaps I'm just missing it.
___
*[`internal/target/leases/leases.go` line 186 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/151#-N4DJNIqD-6Dt1sxWHrX:-N4DJNIqD-6Dt1sxWHrY:bgkoovg) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/dbf3d7813e517f2dc170a3d6bb8d5df662a65d2b/internal/target/leases/leases.go#L186)):*
> ```Go
> 		select {
> 		case <-ctx.Done():
> 			return tgt
> ```

Should the lease be explicitly set to cancelled?  Or are you just relying on the context here?
___
*[`internal/target/leases/leases.go` line 235 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/151#-N4DK01y5Wv70eP3pgSj:-N4DK01y5Wv70eP3pgSk:b-ffw0ox) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/dbf3d7813e517f2dc170a3d6bb8d5df662a65d2b/internal/target/leases/leases.go#L235)):*
> ```Go
> const acquireTemplate = `
> WITH
>   proposed (name, expires, nonce) AS (VALUES ($1::STRING, $2::TIMESTAMPTZ, gen_random_uuid())),
> ```

Why TZ?  Doesn't timezones just mess with everything?  What if you have clusters spanning multiple timezones.
___
*[`internal/target/leases/leases_test.go` line 198 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/151#-N4DMd_l4R4v4uwRAiDd:-N4DMd_l4R4v4uwRAiDe:b96hxdr) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/dbf3d7813e517f2dc170a3d6bb8d5df662a65d2b/internal/target/leases/leases_test.go#L198)):*
> ```Go
> 	})
> 
> 	t.Run("keepRenewedExitsIfHijacked", func(t *testing.T) {
> ```

Can you elaborate on exactly what the expected behaviour is here?

bobvawter (commented):
Issuing a `SELECT FOR UPDATE` would bind the lifetime of the callback to the underlying network connection that holds the `SFU` query.  Some operators also get itchy when the SQL P99 increases, and these blocking-type queries would tend to drive that metric up.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/151#-:-N4XlVPj8WvPADOn4kRM:b-u2nofe)* status: 1 of 3 files reviewed, 7 unresolved discussions (waiting on @BramGruneir)

*[a discussion](https://reviewable.io/reviews/cockroachdb/cdc-sink/151#-N4Xxtub9ho2m_wMUGNF:-N4Xxtub9ho2m_wMUGNG:b-idmig1) (no related file):*
Comments addressed, PTAL.

___
___
*[`internal/target/leases/leases.go` line 160 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/151#-N48ARAuD7bcfz0-ubdO:-N4XYdHg010hw5JRcfVB:b-v7gz49) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/dbf3d7813e517f2dc170a3d6bb8d5df662a65d2b/internal/target/leases/leases.go#L160)):*

Should this wait until the known lease has expired instead of just polling?
</blockquote></details>

If some other instance deletes the lease (e.g. is shut down), we want other agents to be able to pick up sooner than waiting out the expiration time might allow for.  I added a comment to this effect as well as an extra bit of randomness to the time-to-wait.
___
*[`internal/target/leases/leases.go` line 174 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/151#-N4DIa6o-gzzh8HZuE-Z:-N4XiiUv4zCbbd34Xfi8:b-oaxcoy) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/dbf3d7813e517f2dc170a3d6bb8d5df662a65d2b/internal/target/leases/leases.go#L174)):*

So you have expire time, and guard time.  Why not just expire-guard?  Or perhaps set guard as a percentage of total lease time?
</blockquote></details>

`GuardTime` was mis-named.  It's a duration, and renamed to `Guard`.
___
*[`internal/target/leases/leases.go` line 176 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/151#-N4DJ6R81LrWknaN_MNf:-N4XjaWh8k26CPZtHrfv:bqvtnmi) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/dbf3d7813e517f2dc170a3d6bb8d5df662a65d2b/internal/target/leases/leases.go#L176)):*

Is the lease considered expired here just due to the time being after the delay?  Won't it still appear valid if you're within the guard time?  I see the conditions mentioned in the comment above, but there is ambiguity here.  Or perhaps I'm just missing it.
</blockquote></details>

I have restructured this check to be more clear.
___
*[`internal/target/leases/leases.go` line 186 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/151#-N4DJNIqD-6Dt1sxWHrX:-N4Xjqgk8-sYbgPEgMKP:blgsh5l) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/dbf3d7813e517f2dc170a3d6bb8d5df662a65d2b/internal/target/leases/leases.go#L186)):*

Should the lease be explicitly set to cancelled?  Or are you just relying on the context here?
</blockquote></details>

No, the lease is returned just to make one of the tests easier to write.
___
*[`internal/target/leases/leases.go` line 235 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/151#-N4DK01y5Wv70eP3pgSj:-N4XhDj47nKdWSNqT4CR:bmk2jpu) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/dbf3d7813e517f2dc170a3d6bb8d5df662a65d2b/internal/target/leases/leases.go#L235)):*

Why TZ?  Doesn't timezones just mess with everything?  What if you have clusters spanning multiple timezones.
</blockquote></details>

TimestampTZ's can be compared across timezones.  In either case, I changed the schema to use a plain timestamp and call the `UTC()` method on the local `time.Time` before passing it into the database.
___
*[`internal/target/leases/leases_test.go` line 198 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/151#-N4DMd_l4R4v4uwRAiDd:-N4XkUQE5ddbDs7cQ0dU:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/dbf3d7813e517f2dc170a3d6bb8d5df662a65d2b/internal/target/leases/leases_test.go#L198)):*

Can you elaborate on exactly what the expected behaviour is here?
</blockquote></details>

Done.

BramGruneir (approved):
:lgtm:

> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/151#-:-N4bWEzi122HIE5uQfd-:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @BramGruneir)

BramGruneir (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/151)* status: all files reviewed, 1 unresolved discussion (waiting on @BramGruneir)

PR #152, opened 2022-06-09
Add resolve package

bobvawter:
Note to reviewers, this PR includes #151.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/152)

### DIFF
--- a/internal/target/resolve/factory.go
+++ b/internal/target/resolve/factory.go
@@ -1,0 +1,154 @@
+// Copyright 2022 The Cockroach Authors.
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package resolve
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/cockroachdb/cdc-sink/internal/types"
+	"github.com/cockroachdb/cdc-sink/internal/util/ident"
+	"github.com/jackc/pgx/v4/pgxpool"
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+)
+
+// Config is passed to New.
+type Config struct {
+	Appliers   types.Appliers
+	MetaTable  ident.Table
+	Pool       *pgxpool.Pool
+	Stagers    types.Stagers
+	Timekeeper types.TimeKeeper
+	Watchers   types.Watchers
+}
+
+// Table is the usual choice for the timestamp table in normal operating
+// conditions.
+var Table = ident.NewTable(ident.StagingDB, ident.Public, ident.New("pending_timestamps"))
+
+type factory struct {
+	appliers   types.Appliers
+	metaTable  ident.Table
+	pool       *pgxpool.Pool
+	stagers    types.Stagers
+	timekeeper types.TimeKeeper
+	watchers   types.Watchers
+
+	noLoop bool // Set by test to disable starting resolve loops.
+
+	mu struct {
+		sync.RWMutex
+		cleanup   []func()
+		instances map[ident.Schema]*resolve
+	}
+}
+
+var _ types.Resolvers = (*factory)(nil)
+
+// New constructs a Resolver factory.
+func New(ctx context.Context, cfg Config) (_ types.Resolvers, cancel func(), _ error) {
+	if _, err := cfg.Pool.Exec(ctx, fmt.Sprintf(schema, cfg.MetaTable)); err != nil {
+		return nil, func() {}, errors.WithStack(err)
+	}
+
+	f := &factory{
+		appliers:   cfg.Appliers,
+		metaTable:  cfg.MetaTable,
+		pool:       cfg.Pool,
+		stagers:    cfg.Stagers,
+		timekeeper: cfg.Timekeeper,
+		watchers:   cfg.Watchers,
+	}
+	f.mu.instances = make(map[ident.Schema]*resolve)
+
+	// Run the bootstrap in a background context.
+	bootstrapCtx, cancelBoot := context.WithCancel(context.Background())
+	go f.bootstrapResolvers(bootstrapCtx)
+
+	return f, func() {
+		defer cancelBoot()
+
+		f.mu.Lock()
+		defer f.mu.Unlock()
+		for _, fn := range f.mu.cleanup {
+			fn()
+		}
+		f.mu.cleanup = nil
+		f.mu.instances = make(map[ident.Schema]*resolve)
+	}, nil
+}
+
+// Get implements types.Resolvers.
+func (f *factory) Get(ctx context.Context, target ident.Schema) (types.Resolver, error) {
+	if ret, ok := f.getUnlocked(target); ok {
+		return ret, nil
+	}
+
+	f.mu.Lock()
+	defer f.mu.Unlock()
+
+	if found, ok := f.mu.instances[target]; ok {
+		return found, nil
+	}
+
+	ret, err := newResolve(ctx,
+		f.appliers, f.metaTable, f.pool, f.stagers, target,
+		f.timekeeper, f.watchers)
+	if err != nil {
+		return nil, err
+	}
+
+	f.mu.instances[target] = ret
+	if !f.noLoop {
+		// Run the flush behavior in an isolated context.
+		flushCtx, cancel := context.WithCancel(context.Background())
+		f.mu.cleanup = append(f.mu.cleanup, cancel)
+		go ret.loop(flushCtx)
+	}
+
+	return ret, nil
+}
+
+func (f *factory) getUnlocked(target ident.Schema) (types.Resolver, bool) {
+	f.mu.RLock()
+	defer f.mu.RUnlock()
+	found, ok := f.mu.instances[target]
+	return found, ok
+}
+
+// bootstrapResolvers ensures that there is an active resolve instance
+// for every target schema listed in the metaTable. This ensures that,
+// in a zero-incoming-traffic situation, previously-marked values will
+// eventually be processed.
+func (f *factory) bootstrapResolvers(ctx context.Context) {
+	for {
+		toEnsure, err := scanForTargetSchemas(ctx, f.pool, f.metaTable)
+		if err != nil {
+			log.WithError(err).Warn("could not scan for bootstrap schemas")
+		}
+		// toEnsure will be nil if there was an error.
+		for _, schema := range toEnsure {
+			if _, err := f.Get(ctx, schema); err != nil {
+				log.WithField("schema", schema).WithError(err).Warn("could not bootstrap schema")
+			}
+		}
+		select {
+		case <-ctx.Done():
+			return
+		case <-time.After(time.Minute):
+			// We can run this at a slow cycle, since other nodes will
+			// create their resolve instances based on incoming traffic.
+		}
+	}
+}
--- a/internal/target/resolve/metrics.go
+++ b/internal/target/resolve/metrics.go
@@ -1,0 +1,40 @@
+// Copyright 2022 The Cockroach Authors.
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package resolve
+
+import (
+	"github.com/cockroachdb/cdc-sink/internal/util/metrics"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+var (
+	resolveFlushErrors = promauto.NewCounterVec(prometheus.CounterOpts{
+		Name: "resolve_flush_errors_total",
+		Help: "the number of flush cycles that ended in an error",
+	}, metrics.SchemaLabels)
+	resolveFlushSuccess = promauto.NewCounterVec(prometheus.CounterOpts{
+		Name: "resolve_flush_success_total",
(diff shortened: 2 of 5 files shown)


BramGruneir:
The more I think about this, the more I think you've overcomplicated the design here.

1) Every time a cdc-sink node insert a new resolved timestamp or on a random interval have it do a select for update on the oldest non-processed resolved timestamp
2) If it exists, perform the upserted resolution
3) update the resolved timestamp to having been processed

This avoids the singleton, the leases, etc.  And leverages the DB to do most of the work and requires a lot less coordination.

BramGruneir (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/152)* status: all files reviewed, 5 unresolved discussions (waiting on @bobvawter and @sravotto)

___
*[`internal/target/resolve/factory.go` line 37 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/152#-N4DoZN25oZqPWPuD4GQ:-N4DoZN25oZqPWPuD4GR:bq1s1xz) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/28c6bb3f548994e0f1cb2b9861e4b8ab710491dc/internal/target/resolve/factory.go#L37)):*
> ```Go
> }
> 
> // Table is the usual choice for the timestamp table in normal operating
> ```

Do you have plans to store these elsewhere?  Seems like an over-configuration to bake in other places to store them for now.
___
*[`internal/target/resolve/resolve.go` line 70 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/152#-N4Tm_ZWAx73d2tSc3Rn:-N4Tm_ZX5kgwlG5PXxy_:b-g4vvt8) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/28c6bb3f548994e0f1cb2b9861e4b8ab710491dc/internal/target/resolve/resolve.go#L70)):*
> ```Go
>   target_schema     STRING  NOT NULL,
>   source_nanos      INT     NOT NULL,
>   source_logical    INT     NOT NULL,
> ```

Do you think it's worth adding something to indicate which cdc-sink node did the work?
___
*[`internal/target/resolve/resolve.go` line 71 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/152#-N4DlaQqCCXog-y0raDR:-N4DlaQqCCXog-y0raDS:bpdsg0s) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/28c6bb3f548994e0f1cb2b9861e4b8ab710491dc/internal/target/resolve/resolve.go#L71)):*
> ```Go
>   source_nanos      INT     NOT NULL,
>   source_logical    INT     NOT NULL,
>   target_applied_at TIMESTAMPTZ,
> ```

TZ... eww.  Is it common to use that and not UTC?
___
*[`internal/target/resolve/resolve.go` line 73 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/152#-N4Dl33D8oQqLU9Q0uZ2:-N4Dl33D8oQqLU9Q0uZ3:bn03m0u) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/28c6bb3f548994e0f1cb2b9861e4b8ab710491dc/internal/target/resolve/resolve.go#L73)):*
> ```Go
>   target_applied_at TIMESTAMPTZ,
>   PRIMARY KEY (target_db, target_schema, source_nanos, source_logical),
>   INDEX (target_db, target_schema, source_nanos, source_logical)
> ```

why not pk of `(target_db, target_schema, target_applied_at, source_nanos, source_logical)` and skip the index?

Oh... for putting in the mark. Hmm...  I still think this might be faster than having a secondary index.
___
*[`internal/target/resolve/resolve_test.go` line 184 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/152#-N4Dn2o60XtdtIcHvpmD:-N4Dn2o60XtdtIcHvpmE:b5uo8tt) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/28c6bb3f548994e0f1cb2b9861e4b8ab710491dc/internal/target/resolve/resolve_test.go#L184)):*
> ```Go
> 				found[0])
> 		}
> 	})
> ```

Perhaps a test queuing up a backlog of resolved timestamps?

bobvawter (commented):
Per https://github.com/cockroachdb/cockroach/issues/40476 `SELECT FOR UPDATE NOWAIT` has been supported since v20.2, the title of the issue notwithstanding.  I have changed the approach so that any given resolver loop attempts to `SFU NOWAIT` the first unresolved timestamp.  If the timestamp is already being processed by another transaction, the statement will immediately return with a `55P03` code that we treat as a no-work situation.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/152#-:-N4cHlqaDcgouVaAuysW:bgkl2nl)* status: 0 of 7 files reviewed, 5 unresolved discussions (waiting on @BramGruneir and @sravotto)

___
*[`internal/target/resolve/factory.go` line 37 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/152#-N4DoZN25oZqPWPuD4GQ:-N4bU9pB8gE0UcaVJ_SR:b1ti2jc) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/28c6bb3f548994e0f1cb2b9861e4b8ab710491dc/internal/target/resolve/factory.go#L37)):*

Do you have plans to store these elsewhere?  Seems like an over-configuration to bake in other places to store them for now.
</blockquote></details>

Being able to set the table name used by utility classes like this is useful for testing. This field is used in the PR that switches the cdc handler over to using the resolve package.
___
*[`internal/target/resolve/resolve.go` line 70 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/152#-N4Tm_ZWAx73d2tSc3Rn:-N4bUF_c8CtKsrXD5fXq:b-9vg4si) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/28c6bb3f548994e0f1cb2b9861e4b8ab710491dc/internal/target/resolve/resolve.go#L70)):*

Do you think it's worth adding something to indicate which cdc-sink node did the work?
</blockquote></details>

How would it be identified, especially when deployed in an ephemeral environment like k8s?
___
*[`internal/target/resolve/resolve.go` line 71 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/152#-N4DlaQqCCXog-y0raDR:-N4cIVJiD4HzMwlGeS2A:bqp9w7p) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/28c6bb3f548994e0f1cb2b9861e4b8ab710491dc/internal/target/resolve/resolve.go#L71)):*

TZ... eww.  Is it common to use that and not UTC?
</blockquote></details>

Changed to plain `TIMESTAMP`, since the value is generated via the cluster's `now()` function.
___
*[`internal/target/resolve/resolve.go` line 73 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/152#-N4Dl33D8oQqLU9Q0uZ2:-N4cIj5d8-xuWoXdiXto:bgiyh3t) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/28c6bb3f548994e0f1cb2b9861e4b8ab710491dc/internal/target/resolve/resolve.go#L73)):*

why not pk of `(target_db, target_schema, target_applied_at, source_nanos, source_logical)` and skip the index?

Oh... for putting in the mark. Hmm...  I still think this might be faster than having a secondary index.
</blockquote></details>

`target_applied_at` is nullable, so it can't be part of the primary key, unless we use a sentinel value.
___
*[`internal/target/resolve/resolve_test.go` line 184 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/152#-N4Dn2o60XtdtIcHvpmD:-N4cIs421CvUNrr0ODJr:b-d5w6nv) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/28c6bb3f548994e0f1cb2b9861e4b8ab710491dc/internal/target/resolve/resolve_test.go#L184)):*

Perhaps a test queuing up a backlog of resolved timestamps?
</blockquote></details>

Clearing a backlog is part of a new test for validating the singleton `SFU` behavior.

sravotto (commented):
I think now it is simpler and easier to understand. Few minor nitpicks.

> Reviewed 3 of 7 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/152#-:-N4cszML2fk_7JRqW27F:b7kzy6o)* status: 5 of 7 files reviewed, 7 unresolved discussions (waiting on @bobvawter and @BramGruneir)

___
*[`internal/target/resolve/resolve.go` line 323 at r3](https://reviewable.io/reviews/cockroachdb/cdc-sink/152#-N4cp5d57UJW2oQywBhu:-N4cp5d57UJW2oQywBhv:bftp8o1) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/bf4f639bb832799c873614c0ab6cb54372c9ea66/internal/target/resolve/resolve.go#L323)):*
> ```Go
> 		if pgErr := (*pgconn.PgError)(nil); errors.As(err, &pgErr) && pgErr.Code == "55P03" {
> 			log.WithField("schema", r.target).Trace("blocked by other")
> 			return hlc.Zero(), errBlocked
> ```

Would it beneficial to track these kind of errors with a metric? (or alternatively add a label to the resolveFlushErrors to differentiate different errors)?
___
*[`internal/types/types.go` line 79 at r3](https://reviewable.io/reviews/cockroachdb/cdc-sink/152#-N4cs5zU50mnfiZ68_Ed:-N4cs5zU50mnfiZ68_Ee:b-s9g3mn) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/bf4f639bb832799c873614c0ab6cb54372c9ea66/internal/types/types.go#L79)):*
> ```Go
> type Resolver interface {
> 	// Flush is used by tests to ensure that all marked timestamps
> 	// have been processed.
> ```

Is this comment accurate? I am not sure what it means.
Shouldn't it be (from the impl)
// Flush executes a single iteration of the asynchronous resolver logic.
// This method returns true if work was actually performed.

bobvawter (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/152)* status: 5 of 7 files reviewed, 7 unresolved discussions (waiting on @BramGruneir and @sravotto)

___
*[`internal/target/resolve/resolve.go` line 323 at r3](https://reviewable.io/reviews/cockroachdb/cdc-sink/152#-N4cp5d57UJW2oQywBhu:-N4gpmQRCkQL092WIOoJ:bqholqt) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/bf4f639bb832799c873614c0ab6cb54372c9ea66/internal/target/resolve/resolve.go#L323)):*

Would it beneficial to track these kind of errors with a metric? (or alternatively add a label to the resolveFlushErrors to differentiate different errors)?
</blockquote></details>

This is a normal state if there's more than one cdc-sink instance running.  I don't see what diagnostic value this counter would provide.
___
*[`internal/types/types.go` line 79 at r3](https://reviewable.io/reviews/cockroachdb/cdc-sink/152#-N4cs5zU50mnfiZ68_Ed:-N4grbBCAcPwqvV6O6sI:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/bf4f639bb832799c873614c0ab6cb54372c9ea66/internal/types/types.go#L79)):*

Is this comment accurate? I am not sure what it means.
Shouldn't it be (from the impl)
// Flush executes a single iteration of the asynchronous resolver logic.
// This method returns true if work was actually performed.

</blockquote></details>

Done.

BramGruneir (changes requested):
Almost there, just one outstanding issue.

> Reviewed 2 of 7 files at r2, 1 of 2 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/152#-:-N4h6JiiA0fJFD7SA_dZ:b-x07lj6)* status: 5 of 7 files reviewed, 3 unresolved discussions (waiting on @bobvawter, @BramGruneir, and @sravotto)

___
*[`internal/target/resolve/resolve.go` line 70 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/152#-N4Tm_ZWAx73d2tSc3Rn:-N4h4BV674-3bZ1ApkM4:brvhydv) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/28c6bb3f548994e0f1cb2b9861e4b8ab710491dc/internal/target/resolve/resolve.go#L70)):*

How would it be identified, especially when deployed in an ephemeral environment like k8s?
</blockquote></details>

I'm thinking perhaps unique id that the node saves somewhere... yeah.  Not a practical idea.
___
*[`internal/target/resolve/resolve.go` line 73 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/152#-N4Dl33D8oQqLU9Q0uZ2:-N4h4UQf14_bLtMcD33g:bc92xyq) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/28c6bb3f548994e0f1cb2b9861e4b8ab710491dc/internal/target/resolve/resolve.go#L73)):*

`target_applied_at` is nullable, so it can't be part of the primary key, unless we use a sentinel value.
</blockquote></details>

Maybe I missed an update, but not you have pk and index that match.

BramGruneir (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/152#-:-N4hXlRj1hyozYKrrIrQ:bnfp4nl)* status: 5 of 7 files reviewed, 2 unresolved discussions (waiting on @BramGruneir and @sravotto)

___
*[`internal/target/resolve/resolve.go` line 73 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/152#-N4Dl33D8oQqLU9Q0uZ2:-N4hXcx43uoL9t1Vh1UG:beor7w4) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/28c6bb3f548994e0f1cb2b9861e4b8ab710491dc/internal/target/resolve/resolve.go#L73)):*

It's a partial index.
</blockquote></details>

Ah, this line split it.  great.

PR #154, opened 2022-06-13
hlc: Make timestamps loggable with fluent

bobvawter:
This change adds a MarshalJSON method the hlc.Time type so that it can be
pretty-printed by the json-format log output style.

### DIFF
--- a/internal/util/hlc/hlc.go
+++ b/internal/util/hlc/hlc.go
@@ -15,6 +15,7 @@ package hlc
 // The code in this file is reworked from sink_table.go.
 
 import (
+	"encoding/json"
 	"fmt"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


BramGruneir (approved):
Do you think it's worth adding a tiny unit test to make sure the format doesn't drift?  
Up to you.

:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/154#-:-N4Tl-wp3F7o-oIMXMdd:bnumubf)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sravotto)

bobvawter:
Added a check for `MarshalJSON()` to `TestParse()`, which already checked the `String()` method.

PR #164, opened 2022-07-01
Remove the webhook default when testing CDC feeds.

BramGruneir:
After discussions with the CDC team, the preferred method of http endpoint
should not be webhook as it is optimized and will throttle the outgoing CDC
feed.  This change now tests the http instead, even when the webhook endpoint
is available.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/164)

### DIFF
--- a/internal/source/server/provider.go
+++ b/internal/source/server/provider.go
@@ -21,7 +21,6 @@ import (
 	"math/big"
 	"net"
 	"net/http"
-	"strings"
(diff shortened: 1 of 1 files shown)


BramGruneir:
This change isn't right.  I'm not tackling the problem correctly, as this is just the config.  And we do want to test the webhook.

PR #165, opened 2022-07-01
Remove the preference for webhook in the integration tests.

BramGruneir:
After discussions with the CDC team, the preferred endpoint for http based CDC
feeds should not be webhook as it is unoptimized and will throttle the outgoing
CDC feed.

This change removes the preference for testing only webhook when available and
now tests both the http and webhook endpoints.

### DIFF
--- a/internal/source/server/integration_test.go
+++ b/internal/source/server/integration_test.go
@@ -70,7 +70,6 @@ func testIntegration(t *testing.T, immediate bool) {
 	targetCtx := targetFixture.Context
 	targetDB := targetFixture.TestDB.Ident()
 	targetPool := targetFixture.Pool
-	useWebhook := targetFixture.Config.GenerateSelfSigned
 
 	// Set up source and target tables.
 	source, err := sourceFixture.CreateTable(sourceCtx, "CREATE TABLE %s (pk INT PRIMARY KEY, val STRING)")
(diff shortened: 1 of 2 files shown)


bobvawter (changes requested):
You'll either need to add a `COCKROACH_DEV_LICENSE` secret to your fork of the repo or use a branch in the main repo to create the PR from in order to get the tests to pass.

> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/165#-:-N5v6cQB3OoMBV0YLMJB:bwdi9sc)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @BramGruneir)

___
*[`internal/source/server/integration_test.go` line 115 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/165#-N5v71ds0ZS347Ohzhxv:-N5v71ds0ZS347Ohzhxw:b-3jrja3) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/1a683f8148204d89b9d916987b5a610eca82296d/internal/source/server/integration_test.go#L115)):*
> ```Go
> 	var feedURL url.URL
> 	var createStmt string
> 	if useWebhook {
> ```

The current test and the new `testIntegrationWebhook` repeat most code, don't they?  How about extracting the current test method into a helper method `func testIntegration(t *testing.T, useWebhook bool)`, then call it as two separate tests?

BramGruneir:
I just realized the tests were not execution due to the wrapper test.  Update in a bit.

BramGruneir:
*[`internal/source/server/integration_test.go` line 115 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/165#-N5v71ds0ZS347Ohzhxv:-N5v94cYCAbQn2fbJjyQ:bx7bkcx) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/1a683f8148204d89b9d916987b5a610eca82296d/internal/source/server/integration_test.go#L115)):*

The current test and the new `testIntegrationWebhook` repeat most code, don't they?  How about extracting the current test method into a helper method `func testIntegration(t *testing.T, useWebhook bool)`, then call it as two separate tests?
</blockquote></details>

I walked away and was thinking the same thing. 

Done

BramGruneir (commented):
This is ready for another reivew.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/165#-:-N6xsuPu8IYMoDXyQnzA:bzb73si)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @bobvawter)

BramGruneir:
*[`internal/source/server/integration_test.go` line 46 at r3](https://reviewable.io/reviews/cockroachdb/cdc-sink/165#-N6xxgB81b_unlY3zxrS:-N6yCaYRDGEJ-ELJj6bn:b-kxj4br) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f7385602ecb62919b42143cdd83c9568ecb2f1d3/internal/source/server/integration_test.go#L46)):*

If you change the type of the `webhook` param to `shouldUseWebhook`, it would allow the untyped `true false` constants to work and obviate the need for the cast below.
</blockquote></details>

I did make this change... but had to change a bunch of other casts... there's no winning here.

PR #170, opened 2022-07-14
Remove the preference for webhook in the integration tests.

BramGruneir:
After discussions with the CDC team, the preferred endpoint for http based CDC
feeds should not be webhook as it is unoptimized and will throttle the outgoing
CDC feed.

This change removes the preference for testing only webhook when available and
now tests both the http and webhook endpoints.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/170)


BramGruneir:
This replaces https://github.com/cockroachdb/cdc-sink/pull/165

PR #173, opened 2022-07-15
logical: Prevent missed consistent-point notification

bobvawter:
The loop.setConsistentPoint() method calls sync.Cond.Broadcast() outside of a
locked region. This can lead to a missed wakeup in loop.OnCommit() when running
in serial mode. The current code also has two ways to access the mutex that
guards the consistent point stamp: through the mu field as well as the
consistentPointUpdated sync.Cond.

This change moves the condition variable (and associated Locker) into the same
struct as the consistent-point stamp. It also ensures that the lock is held
when calling Broadcast().

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/173)


bobvawter:
Ping

PR #189, opened 2022-07-27
workflow: Matrix testing refinements

bobvawter:
This change simplifies the number of integration tests that are run by only
testing a single version of CockroachDB with the various logical-replication
sources. Older versions of CRDB are still tested for compatibility, but we
assume that the logical-replication source won't make a practical difference.

There is also a prepatory change to log into the GitHub Packages docker
repository, to facilitate an upcoming logical-replication source which requires
a non-trivial emulator environment.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/189)

### DIFF
--- a/.github/workflows/golang.yaml
+++ b/.github/workflows/golang.yaml
@@ -1,6 +1,7 @@
 name: Golang
 permissions:
   contents: read
+  packages: read
 on:
   push:
     branches: [ master ]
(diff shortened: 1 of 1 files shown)


bobvawter:
See https://github.com/cockroachdb/cdc-sink/actions/runs/2747725102 for an example of the reduced test-matrix size.

BramGruneir (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/189#-:-N8PmpPPC3e9-AbdwlI8:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sravotto)

PR #190, opened 2022-07-27
logical: Add support for backfilling

bobvawter:
This change adds an explicit API for logical dialects which can support faster
backfilling of data. The goal is to use fan mode during a backfill and then
switch to a fully-transactional mode once the data load has been completed.

The backfill mode is gated behind an --backfillWindow flag, to ensure that the
operator is aware that the database may not be transactionally-consistent with
respect to the source if replication lags by more than the specified window.

A memoizing logical.Factory type is added to support a future use case that is
best expressed as a collection of independent replication loops that share the
target stack. The logical.Backfiller interface is also added to support future
work.

The loop type no longer implements the loop.Event interface. Instead, we
introduce a fan- and a serial-mode implementation, which are chosen based on
the mode in which the loop is operating. Wrappers are added to suport chaos and
metrics collection.

The pglogical and mylogical packages are updated to support the revised API.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/190)


BramGruneir (changes requested):
> Reviewed 14 of 14 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/190)* status: all files reviewed, 7 unresolved discussions (waiting on @bobvawter and @sravotto)

___
*[`internal/source/logical/config.go` line 32 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/190#-N84L5geAZTvoePJhNfu:-N84L5geAZTvoePJhNfv:b-lult6j) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e71a3df5f3d9143ca90581e91905311c1c000923/internal/source/logical/config.go#L32)):*
> ```Go
> 	// AllowBackfill enables the use of a non-transaction,
> 	// high-throughput mode for backfilling data sources.
> 	AllowBackfill bool
> ```

Can this not just be a duration instead of a bool?

If this is only available per dialect, then it should probably be pushed down to the dialects themselves.
___
*[`internal/source/logical/config.go` line 86 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/190#-N83vxjv2FaY2XcR9BSt:-N83vxjv2FaY2XcR9BSu:bf66p4r) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e71a3df5f3d9143ca90581e91905311c1c000923/internal/source/logical/config.go#L86)):*
> ```Go
> 	// DefaultConsistentPoint used only by mylogical.
> 	f.BoolVar(&c.Immediate, "immediate", false, "apply data without waiting for transaction boundaries")
> 	f.IntVar(&c.FanShards, "fanShards", 16,
> ```

Why 16?  Maybe base this on cpus?
___
*[`internal/source/logical/factory.go` line 44 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/190#-N8PhdTS6Rl15KF4Re_G:-N8PhdTS6Rl15KF4Re_H:b-9043bv) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e71a3df5f3d9143ca90581e91905311c1c000923/internal/source/logical/factory.go#L44)):*
> ```Go
> 	defer f.mu.Unlock()
> 	for _, cancel := range f.mu.cancels {
> 		cancel()
> ```

Would this be faster if the closes were all called at once?  Using a waitgroup?
___
*[`internal/source/logical/fan_events.go` line 29 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/190#-N8PiKac2-nNwVNYARpT:-N8PiKac2-nNwVNYARpU:b24u9jk) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e71a3df5f3d9143ca90581e91905311c1c000923/internal/source/logical/fan_events.go#L29)):*
> ```Go
> 	config *Config
> 	fans   *fan.Fans
> ```

Ok, what's the difference between, fan, fans and fanevents?  Also, why the empty line here?
___
*[`internal/source/logical/fan_events.go` line 73 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/190#-N8Pj_ev9zfr6PxCz9OV:-N8Pj_ev9zfr6PxCz9OW:b-651rv4) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e71a3df5f3d9143ca90581e91905311c1c000923/internal/source/logical/fan_events.go#L73)):*
> ```Go
> }
> 
> // OnRollback implements Events and resetes the enclosed fan.
> ```

nit resets
___
*[`internal/source/logical/loop.go` line 224 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/190#-N8Pl3y086sCUPNwWwTN:-N8Pl3y086sCUPNwWwTO:b6teus2) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e71a3df5f3d9143ca90581e91905311c1c000923/internal/source/logical/loop.go#L224)):*
> ```Go
> 
> 	// This goroutine applies the incoming mutations to the target
> 	// database. It is fragile, when it errors, we need to also
> ```

What do you mean by fragile here?
___
*[`internal/source/logical/serial_events.go` line 86 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/190#-N8Plw_E873txsA9PORY:-N8Plw_E873txsA9PORZ:b4y4b3m) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e71a3df5f3d9143ca90581e91905311c1c000923/internal/source/logical/serial_events.go#L86)):*
> ```Go
> func (e *serialEvents) stop() {
> 	if e.tx != nil {
> 		_ = e.tx.Rollback(context.Background())
> ```

Ignoring the error?

bobvawter (commented):
PTAL: Revised, with time-based backfilling approach.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/190#-:-N8QExIW8WTSXZIHAWhs:b-sjw82j)* status: 3 of 25 files reviewed, 6 unresolved discussions (waiting on @BramGruneir and @sravotto)

___
*[`internal/source/logical/config.go` line 32 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/190#-N84L5geAZTvoePJhNfu:-N8PrpziCmkMcEjr3wUW:b-yacw4h) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e71a3df5f3d9143ca90581e91905311c1c000923/internal/source/logical/config.go#L32)):*

Can this not just be a duration instead of a bool?

If this is only available per dialect, then it should probably be pushed down to the dialects themselves.
</blockquote></details>

I have pushed a new take on the configuration so that the loop looks at the approximate wall timestamp of latest consistent point (if there is one) to determine if it is sufficiently stale to start up in backfill mode.
___
*[`internal/source/logical/config.go` line 86 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/190#-N83vxjv2FaY2XcR9BSt:-N8PvC7f8POQveGK6sDG:b4btivu) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e71a3df5f3d9143ca90581e91905311c1c000923/internal/source/logical/config.go#L86)):*

Why 16?  Maybe base this on cpus?
</blockquote></details>

Strictly speaking, this should be set based on the number of vCPUs in the target cluster.  I never ran out of CPU quota for `cdc-sink` before tipping over the target cluster.
___
*[`internal/source/logical/factory.go` line 44 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/190#-N8PhdTS6Rl15KF4Re_G:-N8PtF0G-2oFAt_0oafY:b6q5zz1) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e71a3df5f3d9143ca90581e91905311c1c000923/internal/source/logical/factory.go#L44)):*

Would this be faster if the closes were all called at once?  Using a waitgroup?
</blockquote></details>

I added a comment explaining that those are context-cancellation functions, so they return immediately.
___
*[`internal/source/logical/fan_events.go` line 29 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/190#-N8PiKac2-nNwVNYARpT:-N8PuMBY66A3Ty7PcdLO:b-iiwdeh) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e71a3df5f3d9143ca90581e91905311c1c000923/internal/source/logical/fan_events.go#L29)):*

Ok, what's the difference between, fan, fans and fanevents?  Also, why the empty line here?
</blockquote></details>

Added comments to clarify their use. The top block are set once by the provider code and the bottom block is mutable state.  Sometimes, I wish golang had a `final` modifier.
___
*[`internal/source/logical/fan_events.go` line 73 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/190#-N8Pj_ev9zfr6PxCz9OV:-N8PugO4B4ChO5ojCu22:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e71a3df5f3d9143ca90581e91905311c1c000923/internal/source/logical/fan_events.go#L73)):*

nit resets
</blockquote></details>

Done.
___
*[`internal/source/logical/loop.go` line 224 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/190#-N8Pl3y086sCUPNwWwTN:-N8Pui_51ijIZB4wDfRt:bvh04mw) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e71a3df5f3d9143ca90581e91905311c1c000923/internal/source/logical/loop.go#L224)):*

What do you mean by fragile here?
</blockquote></details>

I mean to say that is expected to be a regular source of errors outside of our control (e.g. network issues).
___
*[`internal/source/logical/serial_events.go` line 86 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/190#-N8Plw_E873txsA9PORY:-N8PuuXED2aFhQiW-uQW:bbs9ls7) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e71a3df5f3d9143ca90581e91905311c1c000923/internal/source/logical/serial_events.go#L86)):*

Ignoring the error?
</blockquote></details>

Yes.  I rarely find that a failure to rollback a transaction is worth doing anything about, since it's usually in cleanup code responding to some other error.

BramGruneir (changes requested):
> Reviewed 18 of 22 files at r2, 1 of 1 files at r4, 1 of 2 files at r5, 1 of 2 files at r6, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/190)* status: all files reviewed, 5 unresolved discussions (waiting on @bobvawter and @sravotto)

___
*[`internal/source/logical/factory.go` line 65 at r8](https://reviewable.io/reviews/cockroachdb/cdc-sink/190#-N8aCmZI10WyiqZMTIbk:-N8aCmZI10WyiqZMTIbl:b8pg6ex) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/8aa3a33e3ea4315938d00551e5264f07f83e63fc/internal/source/logical/factory.go#L65)):*
> ```Go
> 
> 	cfg := f.cfg.Copy()
> 	if cfg.LoopName == "" {
> ```

how useful is this loop name?
___
*[`internal/source/logical/loop.go` line 226 at r8](https://reviewable.io/reviews/cockroachdb/cdc-sink/190#-N8aGV99AWkmnWKMTbxr:-N8aGV99AWkmnWKMTbxs:bkyg9lm) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/8aa3a33e3ea4315938d00551e5264f07f83e63fc/internal/source/logical/loop.go#L226)):*
> ```Go
> 	// process once we've caught up. This routine never errors out, and
> 	// we don't need to wait for it below.
> 	if isBackfilling {
> ```

This unbounded loop concerns me.  This just seems like a lot of extra work when we can just look at timestamp.  You have the strategy below, is that not enough?  Are you worried about thrashing?  
___
*[`internal/source/logical/metrics_events.go` line 38 at r8](https://reviewable.io/reviews/cockroachdb/cdc-sink/190#-N8aG1Jr6EUmmJWsDsZm:-N8aG1Jr6EUmmJWsDsZn:b-hm3ng8) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/8aa3a33e3ea4315938d00551e5264f07f83e63fc/internal/source/logical/metrics_events.go#L38)):*
> ```Go
> 	commitTime = promauto.NewGaugeVec(prometheus.GaugeOpts{
> 		Name: "logical_last_commit_seconds",
> 		Help: "the original time of the most recently applied commit from the source database",
> ```

Probably worth adding a "immediate vs transaction" metric.
___
*[`internal/source/mylogical/config.go` line 111 at r8](https://reviewable.io/reviews/cockroachdb/cdc-sink/190#-N8aDs-W-vj3twJAYdx2:-N8aDs-W-vj3twJAYdx3:b-6bigze) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/8aa3a33e3ea4315938d00551e5264f07f83e63fc/internal/source/mylogical/config.go#L111)):*
> ```Go
> 	processID, err := strconv.ParseInt(c.LoopName, 10, 32)
> 	if err != nil {
> 		return errors.New("the LoopName must be an integer")
> ```

?  Are you sure about this?  Why are you parsing the loopname into an int?
___
*[`internal/source/mylogical/consistent_point.go` line 158 at r8](https://reviewable.io/reviews/cockroachdb/cdc-sink/190#-N8aEcju467jCXy15jxr:-N8aEcju467jCXy15jxs:b-h05wnb) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/8aa3a33e3ea4315938d00551e5264f07f83e63fc/internal/source/mylogical/consistent_point.go#L158)):*
> ```Go
> }
> 
> func (c *consistentPoint) MarshalJSON() ([]byte, error) {
> ```

Do you have tests for these?

bobvawter (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/190)* status: 16 of 26 files reviewed, 5 unresolved discussions (waiting on @BramGruneir and @sravotto)

___
*[`internal/source/logical/factory.go` line 65 at r8](https://reviewable.io/reviews/cockroachdb/cdc-sink/190#-N8aCmZI10WyiqZMTIbk:-N8d8T1Z7EXJ9GPxgCov:b-rcr0nm) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/8aa3a33e3ea4315938d00551e5264f07f83e63fc/internal/source/logical/factory.go#L65)):*

how useful is this loop name?
</blockquote></details>

It's used in the Firestore integration.  I'm running a loop per `(source collection, target table)` pair.  Each loop needs a distinct name to store its consistent-point data under.  It, arguably, makes the metrics slightly more useful.
___
*[`internal/source/logical/loop.go` line 226 at r8](https://reviewable.io/reviews/cockroachdb/cdc-sink/190#-N8aGV99AWkmnWKMTbxr:-N8d7QhDB_LENeqnq7Zt:b7cenow) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/8aa3a33e3ea4315938d00551e5264f07f83e63fc/internal/source/logical/loop.go#L226)):*

This unbounded loop concerns me.  This just seems like a lot of extra work when we can just look at timestamp.  You have the strategy below, is that not enough?  Are you worried about thrashing?  
</blockquote></details>

I'm not following the comment.
___
*[`internal/source/logical/metrics_events.go` line 38 at r8](https://reviewable.io/reviews/cockroachdb/cdc-sink/190#-N8aG1Jr6EUmmJWsDsZm:-N8dAnxD6h_2zYP076mV:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/8aa3a33e3ea4315938d00551e5264f07f83e63fc/internal/source/logical/metrics_events.go#L38)):*

Probably worth adding a "immediate vs transaction" metric.
</blockquote></details>

Done.
___
*[`internal/source/mylogical/config.go` line 111 at r8](https://reviewable.io/reviews/cockroachdb/cdc-sink/190#-N8aDs-W-vj3twJAYdx2:-N8d7qey1eBvf4wwyrlI:b-p5ll5i) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/8aa3a33e3ea4315938d00551e5264f07f83e63fc/internal/source/mylogical/config.go#L111)):*

?  Are you sure about this?  Why are you parsing the loopname into an int?
</blockquote></details>

I don't know why the original version of config was treating the replication ID as a string to begin with.  I have fixed the field type and flag.
___
*[`internal/source/mylogical/consistent_point.go` line 158 at r8](https://reviewable.io/reviews/cockroachdb/cdc-sink/190#-N8aEcju467jCXy15jxr:-N8dYs_OEeLW9UX8jd9k:birn9k4) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/8aa3a33e3ea4315938d00551e5264f07f83e63fc/internal/source/mylogical/consistent_point.go#L158)):*

Do you have tests for these?
</blockquote></details>

They were tested indirectly. There is now a specific check added to `consistent_point_test.go`.

BramGruneir (approved):
:lgtm: 

The concern I have over the potential thrashing can be addressed later.

> Reviewed 3 of 5 files at r9, 2 of 3 files at r10, 3 of 4 files at r12, 1 of 1 files at r13, 1 of 1 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/190#-:-N8iQsy_6rBHTjmx_o-H:berd80c)* status: all files reviewed, 2 unresolved discussions (waiting on @bobvawter, @BramGruneir, and @sravotto)

___
*[`internal/source/logical/loop.go` line 226 at r8](https://reviewable.io/reviews/cockroachdb/cdc-sink/190#-N8aGV99AWkmnWKMTbxr:-N8iPyOlA3VuIRK-TZrL:b-f088uf) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/8aa3a33e3ea4315938d00551e5264f07f83e63fc/internal/source/logical/loop.go#L226)):*

I'm not following the comment.
</blockquote></details>

Sorry, two separate points here.  the first, is 
"why is this loop required?"  

The second is
If you set the backfill limit to 10 mins behind, it is possible to get thrashing (backflll -> transaction -> backfill -> transactions ... etc) if the feed is always straddling that 10 min delay?
___
*[`internal/source/logical/metrics_events.go` line 38 at r8](https://reviewable.io/reviews/cockroachdb/cdc-sink/190#-N8aG1Jr6EUmmJWsDsZm:-N8iQohM3qc9s44cBwjO:b-ucfknh) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/8aa3a33e3ea4315938d00551e5264f07f83e63fc/internal/source/logical/metrics_events.go#L38)):*

Done.
</blockquote></details>

Thanks!

PR #195, opened 2022-07-28
Clarify use of pg_export_snapshot

bobvawter:
This change clarifies that the value returned by `pg_export_snapshot` is only valid for the lifetime of the session which created the snapshot.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/195)


BramGruneir:
Belated :lgtm:

PR #200, opened 2022-08-05
apply: Support user-defined types with mixed-case idents.

bobvawter:
This change adds explict support for user-defined types (e.g. enums) to the
schemawatch code. It also ensures that UDT's with mixed-case identifiers are
correctly supported.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/200)

### DIFF
--- a/internal/target/apply/apply_test.go
+++ b/internal/target/apply/apply_test.go
@@ -27,6 +27,7 @@ import (
 	"github.com/cockroachdb/cdc-sink/internal/util/ident"
 	log "github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 // This test inserts and deletes rows from a trivial table.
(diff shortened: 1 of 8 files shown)


sravotto (approved):
:lgtm: , however I found one of the comment a bit difficult to understand.

> Reviewed 8 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/200#-:-N8jsOC97lH6eigIsC37:bnsdlh0)* status: all files reviewed, 1 unresolved discussion (waiting on @bobvawter and @BramGruneir)

___
*[`internal/target/schemawatch/coldata.go` line 102 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/200#-N8jrH2kASYKzkgWD0fU:-N8jrH2kASYKzkgWD0fV:box90mn) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/24b5027681bc13889dc23c9a819665750053c8dc/internal/target/schemawatch/coldata.go#L102)):*
> ```Go
> 				foundPrimay = true
> 			}
> 			// If the column type is a user-defined type, e.g. an enum,
> ```

I find this comment a bit unclear. I think it means  that show columns from... returns a data_type that has a format "schema.enum_name", for which we assign a ident, anything that doesn't follow that format is assumed an intrinsic type. Is that right?

bobvawter (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/200)* status: all files reviewed, 1 unresolved discussion (waiting on @BramGruneir and @sravotto)

___
*[`internal/target/schemawatch/coldata.go` line 102 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/200#-N8jrH2kASYKzkgWD0fU:-N8ymld92_S7eLKnpWlV:bmy5i7c) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/24b5027681bc13889dc23c9a819665750053c8dc/internal/target/schemawatch/coldata.go#L102)):*

I find this comment a bit unclear. I think it means  that show columns from... returns a data_type that has a format "schema.enum_name", for which we assign a ident, anything that doesn't follow that format is assumed an intrinsic type. Is that right?
</blockquote></details>

Yes, that is the approach here.

PR #206, opened 2022-08-08
fan: Make use of mapHash stable.

bobvawter:
This change tweaks how we map an incoming mutation onto a specific fan bucket
id by using the same maphash.Seed for each computation. This ensures that
mutations for the same key will be applied on the same underlying database
connection. This removes a potential source of contention in high-rate
scenarios.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/206)

### DIFF
--- a/internal/target/apply/fan/key.go
+++ b/internal/target/apply/fan/key.go
@@ -22,6 +22,10 @@ type bucketKey struct {
 	shard int
 }
 
+// Use a fixed seed value when computing hashes so that a mutation for a
+// specific key winds up in the same bucket.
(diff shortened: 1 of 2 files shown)


BramGruneir (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/206#-:-N8ynmeA49nmG8CpaRiT:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sravotto)

PR #244, opened 2022-12-19
fslogical: Track received document ids

bobvawter:
This change records document ids that have been successfully processed in order
to create idempotent behavior if a document collection must be re-processed by
clearing the collection's consistent point. It may be the case that, for large
document collections and as-yet-unknown reasons, some documents are not
received from Firestore.

The server-generated update-time for the document is recorded in the memo
table, and we ensure that incoming documents will advance the update-time. The
data for a specific document can be queried by executing `SELECT
(convert_from(value, 'utf8')::JSONB->>'u')::TIMESTAMP FROM memo WHERE key LIKE
'fs-doc-%__DOCUMENT_ID__';` in the _cdc_sink database.

This behavior can be disabled by setting `--idempotent=false`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/244)

### DIFF
--- a/internal/source/fslogical/config.go
+++ b/internal/source/fslogical/config.go
@@ -30,6 +30,9 @@ type Config struct {
 	// Copies the document id from the doc metadata into the mutation
 	// using this property name.
 	DocumentIDProperty ident.Ident
+	// Enable extra tracking to allow selective re-processing of
+	// documents in a source collection.
+	Idempotent bool
 	// The Firebase project id. Usually inferred from the credentials.
 	ProjectID string
 	// The name of a collection that contains tombstones for documents
@@ -62,6 +65,8 @@ func (c *Config) Bind(f *pflag.FlagSet) {
 	// NB: Keep default value in sync with doc on tombstones.
 	f.Var(ident.NewValue("id", &c.DocumentIDProperty), "docID",
 		"the column name (likely the primary key) to populate with the document id")
+	f.BoolVar(&c.Idempotent, "idempotent", true,
+		"track received document ids and server times to prevent reprocessing")
 	f.StringVar(&c.LoopName, "loopName", "fslogical",
 		"identifies the logical replication loops in metrics")
 	f.StringVar(&c.ProjectID, "projectID", "",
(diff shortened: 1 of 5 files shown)


BramGruneir (approved):
Just some small notes
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/244#-:-NJgNlAp4ZhU_bAsfh_5:b-qnyd5h)* status: all files reviewed, 2 unresolved discussions (waiting on @bobvawter)

___
*[`internal/source/fslogical/fslogical.go` line 496 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/244#-NJgNPRn0uftf23sDyyj:-NJgNPRn0uftf23sDyyk:bxig7zo) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/dffcdb78251b7e197b7c2c76176cd9b56b86337c/internal/source/fslogical/fslogical.go#L496)):*
> ```Go
> 	ctx context.Context, doc *firestore.DocumentRef, ts time.Time,
> ) error {
> 	payload := processedPayload{UpdatedAt: ts}
> ```

Shouldn't this also have a check for idempotent?
___
*[`internal/source/fslogical/integration_test.go` line 169 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/244#-NJgN_tS7CzAD303beS6:-NJgN_tS7CzAD303beS7:b-7efxs6) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/dffcdb78251b7e197b7c2c76176cd9b56b86337c/internal/source/fslogical/integration_test.go#L169)):*
> ```Go
> 		BackfillBatchSize:           10,
> 		DocumentIDProperty:          ident.New("id"), // Map doc id metadata to target column.
> 		Idempotent:                  true,
> ```

Please add a test with this set to false and ensure it doesn't collect the memos.  Probably best to also check that with it true it does.

BramGruneir (commented):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/244)* status: all files reviewed, 2 unresolved discussions (waiting on @bobvawter)

bobvawter:
Replaced by #325

PR #245, opened 2022-12-22
workflow: Test 22.2 release, drop 20.2

bobvawter:
This change adds v22.2 as the primary version to test all integration cases against, and drops 20.2 from the test matrix.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/245)


BramGruneir:
Lgtm

PR #247, opened 2022-12-27
Cdc logical review

bobvawter:
This is a re-slice of the work in PR #219 into somewhat smaller commits.  The overhaul of the cdc package is still somewhat large, but it does eliminate quite a bit of code.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/247)


BramGruneir (approved):
:lgtm:

This was a ton of code.  But it was all very clean and well commented.  I didn't see any major issues.

> Reviewed 3 of 3 files at r1, 2 of 2 files at r2, 2 of 2 files at r3, 2 of 2 files at r4, 2 of 2 files at r5, 53 of 53 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/247#-:-NKU5BLd3rDWZumIaZT7:b-if7l8p)* status: all files reviewed, 10 unresolved discussions (waiting on @bobvawter and @sravotto)

___
*[`internal/source/cdc/resolver.go` line 63 at r6](https://reviewable.io/reviews/cockroachdb/cdc-sink/247#-NKTob03EPMt6M9CKhm6:-NKTob03EPMt6M9CKhm7:b-8wjdd2) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/5e6e84186d3e8b9f146e8e78beb6e875b8d36955/internal/source/cdc/resolver.go#L63)):*
> ```Go
> // we can roll back to the previously committed stamp through the usual
> // logical-loop error-handling code.
> type resolvedStamp struct {
> ```

Can you split the resolvedStamp into a separate file?  Just to break this up a bit.  
___
*[`internal/source/cdc/resolver.go` line 176 at r6](https://reviewable.io/reviews/cockroachdb/cdc-sink/247#-NKTmomy2VJIkiJZ4uZf:-NKTmomy2VJIkiJZ4uZg:bc1kvyd) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/5e6e84186d3e8b9f146e8e78beb6e875b8d36955/internal/source/cdc/resolver.go#L176)):*
> ```Go
> 		OffsetTime:    s.OffsetTime,
> 		ProposedTime:  s.ProposedTime,
> 	}
> ```

If you do this instantiation a 3rd time, might be worth adding a helper function.  Looks like this and the one in NewProposed only differ in the ProposedTime 
___
*[`internal/source/cdc/resolver.go` line 683 at r6](https://reviewable.io/reviews/cockroachdb/cdc-sink/247#-NKTt09w7O68a0k3FImD:-NKTt09w7O68a0k3FImE:b6705rc) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/5e6e84186d3e8b9f146e8e78beb6e875b8d36955/internal/source/cdc/resolver.go#L683)):*
> ```Go
> 			}
> 			if needsFlush {
> 				if err := flush(toApply); err != nil {
> ```

does it make sense for flush to clear toApply?  So you don't have to make a new one?
___
*[`internal/source/cdc/resolver.go` line 719 at r6](https://reviewable.io/reviews/cockroachdb/cdc-sink/247#-NKTtORL3rBI2vW3wbQ4:-NKTtORL3rBI2vW3wbQ5:b-szzp6) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/5e6e84186d3e8b9f146e8e78beb6e875b8d36955/internal/source/cdc/resolver.go#L719)):*
> ```Go
> // The FOR UPDATE NOWAIT will quickly fail the query with a 55P03 code
> // if another transaction is processing that timestamp. We use the WHERE
> // ... IN clause to ensure a lookup join.
> ```

Could the optimizer change that?
___
*[`internal/source/cdc/resolver.go` line 763 at r6](https://reviewable.io/reviews/cockroachdb/cdc-sink/247#-NKTvTIJAWkbTiBp3Rnx:-NKTuVOh2ttpr-SRQQe6:b1s93to) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/5e6e84186d3e8b9f146e8e78beb6e875b8d36955/internal/source/cdc/resolver.go#L763)):*
> ```Go
> 		after.Logical(),
> 	).Scan(&sourceNanos, &sourceLogical)
> 	if err != nil {
> ```

You could put this if in the line above, unless you find it less readable.
___
*[`internal/source/logical/loop.go` line 175 at r6](https://reviewable.io/reviews/cockroachdb/cdc-sink/247#-NKTyXtuDz_gaUkqvO-D:-NKTyXtuDz_gaUkqvO-E:blnaahr) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/5e6e84186d3e8b9f146e8e78beb6e875b8d36955/internal/source/logical/loop.go#L175)):*
> ```Go
> 
> 	// Use a separate goroutine that is blocked by the sync.Cond.
> 	result := make(chan stamp.Stamp, 1)
> ```

Why?  Just for so you can also wait for the done?
___
*[`internal/target/sinktest/fixture.go` line 68 at r6](https://reviewable.io/reviews/cockroachdb/cdc-sink/247#-NKKShP6-0KYnUfZjwTq:-NKKShP6-0KYnUfZjwTr:bbewm3) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/5e6e84186d3e8b9f146e8e78beb6e875b8d36955/internal/target/sinktest/fixture.go#L68)):*
> ```Go
> 	Watchers types.Watchers
> 
> 	Watcher types.Watcher // A watcher for TestDB.
> ```

Perhaps call it TestDBWatcher?  That way you don't need the comment and it's quite clear?

Yes, I know this is from a previous pr.  Just a passing comment.
___
*[`internal/target/stage/factory.go` line 148 at r6](https://reviewable.io/reviews/cockroachdb/cdc-sink/247#-NKU1hrI3oU-qatDGodJ:-NKU1hrI3oU-qatDGodK:b-t0714n) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/5e6e84186d3e8b9f146e8e78beb6e875b8d36955/internal/target/stage/factory.go#L148)):*
> ```Go
> 			}
> 
> 			// Mangle the real table name into its staging table name.
> ```

I'm a fan of calling this mangle.
___
*[`internal/target/stage/factory.go` line 282 at r6](https://reviewable.io/reviews/cockroachdb/cdc-sink/247#-NKU2dRBAdUO7Zv2R2yS:-NKU2dRBAdUO7Zv2R2yT:b-rhp82k) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/5e6e84186d3e8b9f146e8e78beb6e875b8d36955/internal/target/stage/factory.go#L282)):*
> ```Go
> 		}).Debug("retrieved staged mutations")
> 
> 		if count < q.Limit {
> ```

What happens if the limit is reached to the rows not processed?
___
*[`internal/target/stage/stage.go` line 336 at r6](https://reviewable.io/reviews/cockroachdb/cdc-sink/247#-NKU4Wdz3BGhBKn297q4:-NKU4We-BPYnrzdxkL2p:bo22uw5) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/5e6e84186d3e8b9f146e8e78beb6e875b8d36955/internal/target/stage/stage.go#L336)):*
> ```Go
> 	jsons := make([][]byte, len(mutations))
> 
> 	numWorkers := runtime.GOMAXPROCS(0)
> ```

Is this really the best default? Not 2x?  Is the bottleneck on this cdc-sink side or the dbs?  
___
*[`internal/types/types.go` line 99 at r6](https://reviewable.io/reviews/cockroachdb/cdc-sink/247#-NKU0Vz81F0FtFAvd6SI:-NKU0Vz81F0FtFAvd6SJ:b5b6z22) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/5e6e84186d3e8b9f146e8e78beb6e875b8d36955/internal/types/types.go#L99)):*
> ```Go
> // Stager describes a service which can durably persist some
> // number of Mutations.
> type Stager interface {
> ```

Any reason not to pull Stager into it's own file?
___
*[`internal/util/pjson/pjson.go` line 26 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/247#-NKK7mIs56fc98slivt1:-NKK7mIs56fc98slivt2:bh2cu8q) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/99f8aee9ededaf1f6c4edb8699b13c25b54d3df7/internal/util/pjson/pjson.go#L26)):*
> ```Go
> 
> // Decode decodes a source of JSON bytes into a target slice in a
> // parallel manner.
> ```

This is some of the most clean parallelism code I've seen in a while.  I wish golang would add the forparallel stuff that C# added 15 years ago, would eliminate this type of thing completely.
___
*[`internal/util/pjson/pjson.go` line 50 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/247#-NKK7WSD4aCkAhvomVD_:-NKK7WSD4aCkAhvomVDa:b-7je4b7) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/99f8aee9ededaf1f6c4edb8699b13c25b54d3df7/internal/util/pjson/pjson.go#L50)):*
> ```Go
> }
> 
> // Encode encodes objects into a target slide in a parallel manner.
> ```

slice
___
*[`internal/util/txguard/guard.go` line 119 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/247#-NKK4uc6C_61qVubU_Dq:-NKK4uc6C_61qVubU_Dr:blt460s) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/254e4e34ab1b8ca8c477508a2dbbc3d64c56febe/internal/util/txguard/guard.go#L119)):*
> ```Go
> 	}
> 	g.mu.tx = nil
> 	_ = tx.Rollback(context.Background())
> ```

It's ok to drop the error here?
___
*[`internal/util/txguard/guard.go` line 138 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/247#-NKK590OAwW8RXgZFvCw:-NKK590OAwW8RXgZFvCx:bq1bas6) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/254e4e34ab1b8ca8c477508a2dbbc3d64c56febe/internal/util/txguard/guard.go#L138)):*
> ```Go
> }
> 
> // getTX is used by test code to safely access the mu.tx field.
> ```

safely?  :P

sravotto (approved):
:lgtm: just a question on a couple of metrics.

> Reviewed 3 of 3 files at r1, 2 of 2 files at r2, 2 of 2 files at r3, 2 of 2 files at r4, 2 of 2 files at r5, 53 of 53 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/247#-:-NKUKo-82yJtntS9T1A7:b3qfhi6)* status: all files reviewed, 12 unresolved discussions (waiting on @bobvawter)

___
*[`internal/util/stdpool/metrics.go` line 27 at r3](https://reviewable.io/reviews/cockroachdb/cdc-sink/247#-NKPmNyz57TyQSi7SmuT:-NKPmNyz57TyQSi7SmuU:b-wgl5pi) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/7118de1becb0df6354d54cff560ff29218e3ad87/internal/util/stdpool/metrics.go#L27)):*
> ```Go
> 	hostLabels = []string{"host"}
> 
> 	poolAcquireCount = promauto.NewGaugeVec(prometheus.GaugeOpts{
> ```

Shouldn't this be a counter?
___
*[`internal/util/stdpool/metrics.go` line 31 at r3](https://reviewable.io/reviews/cockroachdb/cdc-sink/247#-NKPmYEm-JQYo2yjNNRy:-NKPmYEm-JQYo2yjNNRz:b-wgl5pi) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/7118de1becb0df6354d54cff560ff29218e3ad87/internal/util/stdpool/metrics.go#L31)):*
> ```Go
> 		Help: "the total number of times we waited to add a connection to the pool",
> 	}, hostLabels)
> 	poolAcquireDelay = promauto.NewGaugeVec(prometheus.GaugeOpts{
> ```

Shouldn't this be a counter?

bobvawter (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/247#-:-NKs8i7zAFIb0aV2T8Mh:b-814g8k)* status: 58 of 65 files reviewed, 10 unresolved discussions (waiting on @BramGruneir and @sravotto)

___
*[`internal/source/cdc/resolver.go` line 63 at r6](https://reviewable.io/reviews/cockroachdb/cdc-sink/247#-NKTob03EPMt6M9CKhm6:-NKs8n5B7itXJnTPM6Rz:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/5e6e84186d3e8b9f146e8e78beb6e875b8d36955/internal/source/cdc/resolver.go#L63)):*

Can you split the resolvedStamp into a separate file?  Just to break this up a bit.  
</blockquote></details>

Done.
___
*[`internal/source/cdc/resolver.go` line 683 at r6](https://reviewable.io/reviews/cockroachdb/cdc-sink/247#-NKTt09w7O68a0k3FImD:-NKs0xBU6DzSuYXZJvoj:b4rd8xn) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/5e6e84186d3e8b9f146e8e78beb6e875b8d36955/internal/source/cdc/resolver.go#L683)):*

does it make sense for flush to clear toApply?  So you don't have to make a new one?
</blockquote></details>

Done. It now resets the slices to zero length.
___
*[`internal/source/cdc/resolver.go` line 719 at r6](https://reviewable.io/reviews/cockroachdb/cdc-sink/247#-NKTtORL3rBI2vW3wbQ4:-NKs13dJ696r4U_BqPBs:btapmjy) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/5e6e84186d3e8b9f146e8e78beb6e875b8d36955/internal/source/cdc/resolver.go#L719)):*

Could the optimizer change that?
</blockquote></details>

There are tests that validate the expected locking behavior, so if the optimizer does change and break the locking, we'll see builds fail.
___
*[`internal/source/cdc/resolver.go` line 763 at r6](https://reviewable.io/reviews/cockroachdb/cdc-sink/247#-NKTvTIJAWkbTiBp3Rnx:-NKs1WmrDACW-KHiclbS:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/5e6e84186d3e8b9f146e8e78beb6e875b8d36955/internal/source/cdc/resolver.go#L763)):*

You could put this if in the line above, unless you find it less readable.
</blockquote></details>

Done.
___
*[`internal/source/logical/loop.go` line 175 at r6](https://reviewable.io/reviews/cockroachdb/cdc-sink/247#-NKTyXtuDz_gaUkqvO-D:-NKs1tG7DzSMGTr2r1A1:b-quq58s) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/5e6e84186d3e8b9f146e8e78beb6e875b8d36955/internal/source/logical/loop.go#L175)):*

Why?  Just for so you can also wait for the done?
</blockquote></details>

Clarified the comment to say that it's to allow for cancellation. The `sync.Cond.Await()` method. doesn't accept a `Context`.
___
*[`internal/target/stage/factory.go` line 148 at r6](https://reviewable.io/reviews/cockroachdb/cdc-sink/247#-NKU1hrI3oU-qatDGodJ:-NKs3jdL7c-Wa0pAIhwp:b-13dlgu) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/5e6e84186d3e8b9f146e8e78beb6e875b8d36955/internal/target/stage/factory.go#L148)):*

I'm a fan of calling this mangle.
</blockquote></details>

The call to `NewTable` was repeated in two locations. I replaced `mangleTableName()` with `stagingTable(stagingDB, targetTable) ident.Table`
___
*[`internal/target/stage/factory.go` line 282 at r6](https://reviewable.io/reviews/cockroachdb/cdc-sink/247#-NKU2dRBAdUO7Zv2R2yS:-NKs4Cy6DeZRejNxU0PC:b-it223j) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/5e6e84186d3e8b9f146e8e78beb6e875b8d36955/internal/target/stage/factory.go#L282)):*

What happens if the limit is reached to the rows not processed?
</blockquote></details>

Added a comment to clarify that this is the stopping condition.
___
*[`internal/target/stage/stage.go` line 336 at r6](https://reviewable.io/reviews/cockroachdb/cdc-sink/247#-NKU4Wdz3BGhBKn297q4:-NKs4PGJ93tBjtBYoEc3:bu1p5my) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/5e6e84186d3e8b9f146e8e78beb6e875b8d36955/internal/target/stage/stage.go#L336)):*

Is this really the best default? Not 2x?  Is the bottleneck on this cdc-sink side or the dbs?  
</blockquote></details>

In doing some performance profiling for the `F` use-case, I found that the `cdc-sink` process was only ever consuming about a 1.5 cpu cores, since we were single-threading on JSON decoding.  With just this number of workers, I can reliably max out the CPU allocation for the process and improve throughput.
___
*[`internal/types/types.go` line 99 at r6](https://reviewable.io/reviews/cockroachdb/cdc-sink/247#-NKU0Vz81F0FtFAvd6SI:-NKs5cRm5Y_zJ12Koh0f:b6dtb69) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/5e6e84186d3e8b9f146e8e78beb6e875b8d36955/internal/types/types.go#L99)):*

Any reason not to pull Stager into it's own file?
</blockquote></details>

I think the `types` package may have served its purpose and could be removed.  It was created in the initial modularization of the cdc-sink code base so I had stub types to avoid red squiggles while prototyping.  There's only ever a single implementation of any of these interfaces and it would be more idiomatic to declare these interfaces in the consuming packages.
___
*[`internal/util/pjson/pjson.go` line 26 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/247#-NKK7mIs56fc98slivt1:-NKs60k58XML43SkyuVI:b-9755fa) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/99f8aee9ededaf1f6c4edb8699b13c25b54d3df7/internal/util/pjson/pjson.go#L26)):*

This is some of the most clean parallelism code I've seen in a while.  I wish golang would add the forparallel stuff that C# added 15 years ago, would eliminate this type of thing completely.
</blockquote></details>

Thank you.
___
*[`internal/util/pjson/pjson.go` line 50 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/247#-NKK7WSD4aCkAhvomVD_:-NKs6Ccd1LeaT-esuwcX:b-2jz69o) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/99f8aee9ededaf1f6c4edb8699b13c25b54d3df7/internal/util/pjson/pjson.go#L50)):*

slice
</blockquote></details>

Fixed. `chan` should have been called `slide`
___
*[`internal/util/stdpool/metrics.go` line 31 at r3](https://reviewable.io/reviews/cockroachdb/cdc-sink/247#-NKPmYEm-JQYo2yjNNRy:-NKs86Os518-C0mYt44X:b-wlwzix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/7118de1becb0df6354d54cff560ff29218e3ad87/internal/util/stdpool/metrics.go#L31)):*

Shouldn't this be a counter?
</blockquote></details>

Computing the deltas for these.
___
*[`internal/util/txguard/guard.go` line 119 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/247#-NKK4uc6C_61qVubU_Dq:-NKs89AoAEOdN2by6_db:bx01314) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/254e4e34ab1b8ca8c477508a2dbbc3d64c56febe/internal/util/txguard/guard.go#L119)):*

It's ok to drop the error here?
</blockquote></details>

There's really nothing useful to do with an error from rolling back a transaction.  If it fails, then nothing happens, which is the intended result.  Really, I'm calling this to ensure that the db pool can clean up appropriately.
___
*[`internal/util/txguard/guard.go` line 138 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/247#-NKK590OAwW8RXgZFvCw:-NKs8YyPEN3tDQrSUi9G:b2b8bi5) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/254e4e34ab1b8ca8c477508a2dbbc3d64c56febe/internal/util/txguard/guard.go#L138)):*

safely?  :P
</blockquote></details>

The tests are run with `-race`, so I still need to make safe accesses.

PR #288, opened 2023-03-23
Add a new commit latency metric

### LINKED ISSUE #286: Provide commit latency metric
Currently, the `logical_last_commit_seconds` is a bit awkward to use for monitoring. It uses the original time (in nanoseconds) of the most recently applied commit from the source database. In services such as DataDog, it doesn't look like there is a way to use an expression to convert that into a diff w/ current time.

It would be more useful to have a commit latency metric, similar to what [CRDB provides for CDC](https://www.cockroachlabs.com/docs/stable/monitor-and-debug-changefeeds.html#metrics). This metric would automatically calculate the difference between current time and the logical last commit. This would give an overall view into how long it takes for rows from a transaction committed on the source cluster to reach the destination cluster.

BramGruneir:
Add a new metric that shows how far behind the destination cluster is from the current time.

Closes #286

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/288)

### DIFF
--- a/internal/source/logical/metrics_events.go
+++ b/internal/source/logical/metrics_events.go
@@ -12,6 +12,7 @@ package logical
 
 import (
 	"context"
+	"time"
 
 	"github.com/cockroachdb/cdc-sink/internal/util/stamp"
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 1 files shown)


bobvawter:
LGTM, but please reorder the lines to keep all of the field definitions and assignments in alphabetical order.

jonstjohn:
Do you think it would have been better to have this in nanoseconds, like we do for the CDC commit latency metric? I'm not sure that it matters but just occurred to me looking at what we do in CDC.

PR #298, opened 2023-04-14
fan: Add sanity-check for semaphore over-relase

bobvawter:
This change calculates the mutation weight of a stampedBatch once and updates it as any batch coalescing occurs. An environment variable, STAMPED_BATCH_CHECKS, is added for temporary testing purposes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/298)


BramGruneir:
Lgtm

PR #313, opened 2023-04-25
apply: Fix key-length check for computed PK columns

bobvawter:
Schemas that have generated columns in their PK structures would incorrectly trigger a sanity-check on the length of the "key" attribute in incoming changefeed messages.

This change does have some knock-on effects on the SQL used to generate delete operations.  Even though some pk columns might be ignored for upsert operations, we do want to use all known pk columns for deletes, to ensure that we don't accidentally delete whole swaths of rows.


BramGruneir (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/313#-:-NTt3YpEBHW7D2ct8Vxz:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sean-)

PR #316, opened 2023-04-28
logical: Rework fan event mode

bobvawter:
This change deletes the apply/fan package, which is now only used by the logical package and replaces it with a simpler model.  Previously, the fan code would try to accumulate batches of mutations that were interleaved with consistent-point markers. When cdc-sink was in a FK-preserving backfill mode, the orderedEvents facade would violate the API contract that fan expected of its callers, leading to restartable runtime errors.

This new implementation of fanWorkers is specific to the fanEvents implementation and lives in the logical package. The main difference in the API is that a single instance of fanWorkers is used to process a single consistent-point stamp (i.e. some subset of a backfill) and the Wait() method is used to ensure that the consistent point has been reached, rather than using an asynchronous callback.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/316)


BramGruneir (approved):
:lgtm:

This is much much cleaner and easier to understand and reason about.  

> Reviewed 27 of 27 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/316#-:-NUNALuv6xRvae4l0t3c:b-lzcvhu)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jonstjohn)

PR #325, opened 2023-05-11
logical: Improve graceful draining

bobvawter:
This change adds a stopper utility package to assist with gracefully
transitioning a Dialect from backfill to transactional modes. The API is very
much like an errgroup, in that it starts some number of goroutines, however the
associated context is canceled after the Stop() method is called and the
goroutines have exited. The existing logical-loop code uses normal context
cancellation to signal a need to change modes, but this winds up being too
coarse-grained, we want the dialect's message producer to exit and for the
message consumer to drain the channel.

The logical.Event interface gains an explicit Flush() method to allow all
in-flight mutations passed to OnData() to be written to the backing store. This
further simplifies FK-preserving backfill mode, since we no longer attempt to
synthesize fake transaction boundary markers between FK levels. Instead, the
orderedEvents facade just calls Flush() between FK levels.

The logical.State interface switches to a channel-driven api when awaiting
changes to the consistent point. This allows callers to compose behavior with
select statements, instead of relying on various forms of context cancellation.
The graceful-stopping signal is also made available through the State
interface.

The private logical.Events.stop() method was renamed to drain() to avoid
confusion.

The various logical Dialects are updated to look for the stopping signal.  This
PR additionally includes the previously-reviewed, but unmerged, changes from PR #244 

This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/325)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,8 +4,9 @@ go 1.20
 
 require (
 	cloud.google.com/go/firestore v1.9.0
+	github.com/bobvawter/latch v1.0.1
 	github.com/cockroachdb/apd v1.1.0
-	github.com/cockroachdb/crlfmt v0.0.0-20220610162206-024b567ce87b
+	github.com/cockroachdb/crlfmt v0.0.0-20221214225007-b2fc5c302548
 	github.com/dop251/goja v0.0.0-20220915101355-d79e1b125a30
 	github.com/evanw/esbuild v0.17.17
 	github.com/go-mysql-org/go-mysql v1.6.0
--- a/go.sum
+++ b/go.sum
@@ -55,6 +55,8 @@ github.com/beorn7/perks v0.0.0-20180321164747-3a771d992973/go.mod h1:Dwedo/Wpr24
 github.com/beorn7/perks v1.0.0/go.mod h1:KWe93zE9D1o94FZ5RNwFwVgaQK1VOXiVxmqh+CedLV8=
 github.com/beorn7/perks v1.0.1 h1:VlbKKnNfV8bJzeqoa4cOKqO6bYr3WgKZxO8Z16+hsOM=
 github.com/beorn7/perks v1.0.1/go.mod h1:G2ZrVWU2WbWT9wwq4/hrbKbnv/1ERSJQ0ibhJ6rlkpw=
+github.com/bobvawter/latch v1.0.1 h1:pSQUCMKlucP4V2OOzSKpMoGdTQ0CucSc+PpH04qub+I=
+github.com/bobvawter/latch v1.0.1/go.mod h1:Y2y27j+pGSfiJVvw+xF8YDFV7cjq5vUS0Vmp40eSupk=
 github.com/census-instrumentation/opencensus-proto v0.2.1/go.mod h1:f6KPmirojxKA12rnyqOA5BBL4O983OfeGPqjHWSTneU=
 github.com/cespare/xxhash/v2 v2.1.1/go.mod h1:VGX0DQ3Q6kWi7AoAeZDth3/j3BFtOZR5XLFGgcrjCOs=
 github.com/cespare/xxhash/v2 v2.1.2/go.mod h1:VGX0DQ3Q6kWi7AoAeZDth3/j3BFtOZR5XLFGgcrjCOs=
@@ -72,12 +74,10 @@ github.com/cncf/xds/go v0.0.0-20210922020428-25de7278fc84/go.mod h1:eXthEFrGJvWH
 github.com/cncf/xds/go v0.0.0-20211011173535-cb28da3451f1/go.mod h1:eXthEFrGJvWHgFFCl3hGmgk+/aYT6PnTQLykKQRLhEs=
 github.com/cockroachdb/apd v1.1.0 h1:3LFP3629v+1aKXU5Q37mxmRxX/pIu1nijXydLShEq5I=
 github.com/cockroachdb/apd v1.1.0/go.mod h1:8Sl8LxpKi29FqWXR16WEFZRNSz3SoPzUzeMeY4+DwBQ=
-github.com/cockroachdb/crlfmt v0.0.0-20220610162206-024b567ce87b h1:kGrgvdN+SBR+DsZIf51EguT29Y7Am1/v9DYoDvhzeyU=
-github.com/cockroachdb/crlfmt v0.0.0-20220610162206-024b567ce87b/go.mod h1:EOI6rrXIdP+4EXwM8837kmmb6IJesf7k7W6bUu8BDOg=
-github.com/cockroachdb/gostdlib v1.13.0/go.mod h1:eXX95p9QDrYwJfJ6AgeN9QnRa/lqqid9LAzWz/l5OgA=
+github.com/cockroachdb/crlfmt v0.0.0-20221214225007-b2fc5c302548 h1:i0bnjanlWAvM50wHMT7EFyxlt5HQusznWrkwl+HBIsU=
+github.com/cockroachdb/crlfmt v0.0.0-20221214225007-b2fc5c302548/go.mod h1:qtkxNlt5i3rrdirfJE/bQeW/IeLajKexErv7jEIV+Uc=
 github.com/cockroachdb/gostdlib v1.19.0 h1:cSISxkVnTlWhTkyple/T6NXzOi5659FkhxvUgZv+Eb0=
 github.com/cockroachdb/gostdlib v1.19.0/go.mod h1:+dqqpARXbE/gRDEhCak6dm0l14AaTymPZUKMfURjBtY=
-github.com/cockroachdb/ttycolor v0.0.0-20180709150743-a1d5aaeb377d/go.mod h1:NltwFG0VBANi1jHKpn5KL9AbsHTE+8fPaAHT0TzL20k=
 github.com/cockroachdb/ttycolor v0.0.0-20210902133924-c7d7dcdde4e8 h1:Hli+oX84dKq44sLVCcsGKqifm5Lg9J8VoJ2P3h9iPdI=
 github.com/cockroachdb/ttycolor v0.0.0-20210902133924-c7d7dcdde4e8/go.mod h1:75wnig8+TF6vst9hChkpcFO7YrRLddouJ5is8uqpfv0=
 github.com/cpuguy83/go-md2man/v2 v2.0.2/go.mod h1:tgQtvFlXSQOSOSIRvRPT7W67SCa46tRHOmNcaadrF8o=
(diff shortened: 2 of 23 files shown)


BramGruneir (approved):
:lgtm:

Just a few small questions.

> Reviewed 23 of 23 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/325#-:-NVGCmgq4-OqrmWiGX-p:bvo0z7o)* status: all files reviewed, 10 unresolved discussions (waiting on @bobvawter)

___
*[`internal/source/fslogical/fslogical.go` line 116 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/325#-NVFh4qW5-WL56IvygsO:-NVFh4qW5-WL56IvygsP:b-br12s8) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e42bee27792c385cf36a90e04eabd426bfd75ee9/internal/source/fslogical/fslogical.go#L116)):*
> ```Go
> 	// We need to make the call to snaps.Next() interruptable.
> 	ctx, cancel := context.WithCancel(ctx)
> 	defer cancel()
> ```

is cancelled called twice?  Once on the defer and once when stopping/done?  I see you've used this double cancel more than once in the code.  So I'm probably missing something.
___
*[`internal/source/fslogical/fslogical.go` line 143 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/325#-NVFhG4h8PwplTe1lwaa:-NVFhG4h8PwplTe1lwab:b-2bamob) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e42bee27792c385cf36a90e04eabd426bfd75ee9/internal/source/fslogical/fslogical.go#L143)):*
> ```Go
> 
> 	snap, err := snaps.Next()
> 	if status.Code(err) == codes.Canceled || errors.Is(err, iterator.Done) {
> ```

Perhaps this would be better to first check if nil, then check the code if not? 
___
*[`internal/source/fslogical/fslogical.go` line 218 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/325#-NVFhh3MBgOqafbQdYbd:-NVFhh3MBgOqafbQdYbe:bff2fu3) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e42bee27792c385cf36a90e04eabd426bfd75ee9/internal/source/fslogical/fslogical.go#L218)):*
> ```Go
> 	// The call to snaps.Next() below needs to be made interruptable.
> 	ctx, cancel := context.WithCancel(ctx)
> 	defer cancel()
> ```

(repeat) double cancel?
___
*[`internal/source/fslogical/fslogical.go` line 248 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/325#-NVFhc1LDgfZhvZV2NAI:-NVFhc1LDgfZhvZV2NAJ:bay2x3f) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e42bee27792c385cf36a90e04eabd426bfd75ee9/internal/source/fslogical/fslogical.go#L248)):*
> ```Go
> 		log.Tracef("getting snapshot for %s", d.sourcePath)
> 		snap, err := snaps.Next()
> 		if status.Code(err) == codes.Canceled || errors.Is(err, iterator.Done) {
> ```

(repeat) nil check first?
___
*[`internal/source/fslogical/fslogical.go` line 576 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/325#-NVFiTJz2nHREu9hoFXt:-NVFiTJz2nHREu9hoFXu:br4uke9) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e42bee27792c385cf36a90e04eabd426bfd75ee9/internal/source/fslogical/fslogical.go#L576)):*
> ```Go
> 	}
> 
> 	log.Tracef("ignoring document %s at %s <= %s", doc.ID, ts, payload.UpdatedAt)
> ```

Under what circumstances would this happen?
___
*[`internal/source/fslogical/tombstones.go` line 106 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/325#-NVFibcPEPOeho7grNwJ:-NVFibcPEPOeho7grNwK:bff2fu3) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e42bee27792c385cf36a90e04eabd426bfd75ee9/internal/source/fslogical/tombstones.go#L106)):*
> ```Go
> 	// Make call to snaps.Next() interruptable.
> 	ctx, cancel := context.WithCancel(ctx)
> 	defer cancel()
> ```

(repeat) double cancel?
___
*[`internal/source/logical/loop.go` line 253 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/325#-NVFkpEAANO_vX5mfc2c:-NVFkpEAANO_vX5mfc2d:b-z35ykt) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e42bee27792c385cf36a90e04eabd426bfd75ee9/internal/source/logical/loop.go#L253)):*
> ```Go
> 
> // runOnce is called by run. If the Dialect implements a leasing
> // behavior, a least will be obtained before any further action is
> ```

s/least/lease
___
*[`internal/util/stopper/stopper.go` line 161 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/325#-NVGBMsz6Q2_Yb1Dv5kw:-NVGBMsz6Q2_Yb1Dv5kx:b-y48pgs) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e42bee27792c385cf36a90e04eabd426bfd75ee9/internal/util/stopper/stopper.go#L161)):*
> ```Go
> // [Stopping] for a notification-based API.
> func (s *Context) IsStopping() bool {
> 	s.mu.Lock()
> ```

Would it be better to use a RWlock so isStopping can be called with just a read lock?  Just in case you end up doing that a lot.
___
*[`internal/util/stopper/stopper.go` line 196 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/325#-NVGBov76EE8ogecncSy:-NVGBov76EE8ogecncSz:b-ndmvzf) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e42bee27792c385cf36a90e04eabd426bfd75ee9/internal/util/stopper/stopper.go#L196)):*
> ```Go
> 				s.cancel(ErrGracePeriodExpired)
> 			}
> 		}()
> ```

Doesn't this mean that the stopping will be triggered after the stopper has been stopped + grace period, instead of waiting for it to stop?  (That was a lot of overuse of the word stop in one sentence.)

I just want to make sure the intent is clear.
___
*[`internal/util/stopper/stopper.go` line 232 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/325#-NVGCTD48vu4pQjP6GHx:-NVGCTD48vu4pQjP6GHy:bbo4nnt) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e42bee27792c385cf36a90e04eabd426bfd75ee9/internal/util/stopper/stopper.go#L232)):*
> ```Go
> // apply is used to maintain the count of started goroutines. It returns
> // true if the delta was applied.
> func (s *Context) apply(delta int) bool {
> ```

I see only +1s and -1s.  Do you ever see a need for >1 or <1?

bobvawter (commented):
TFTR.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/325#-:-NVHhxXL3Ac_8e4Tgqk8:b-814g87)* status: 19 of 23 files reviewed, 10 unresolved discussions (waiting on @BramGruneir)

___
*[`internal/source/fslogical/fslogical.go` line 116 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/325#-NVFh4qW5-WL56IvygsO:-NVHeIIB5bSHRnvtHNlW:bra8v3q) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e42bee27792c385cf36a90e04eabd426bfd75ee9/internal/source/fslogical/fslogical.go#L116)):*

is cancelled called twice?  Once on the defer and once when stopping/done?  I see you've used this double cancel more than once in the code.  So I'm probably missing something.
</blockquote></details>

Added some comments to clarify why there's both a `defer cancel()` and an explicit `cancel()`.
___
*[`internal/source/fslogical/fslogical.go` line 143 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/325#-NVFhG4h8PwplTe1lwaa:-NVHeI1N2rh2AKMC0jpt:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e42bee27792c385cf36a90e04eabd426bfd75ee9/internal/source/fslogical/fslogical.go#L143)):*

Perhaps this would be better to first check if nil, then check the code if not? 
</blockquote></details>

Done.
___
*[`internal/source/fslogical/fslogical.go` line 218 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/325#-NVFhh3MBgOqafbQdYbd:-NVHel1L1uTT0VwIIyXl:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e42bee27792c385cf36a90e04eabd426bfd75ee9/internal/source/fslogical/fslogical.go#L218)):*

(repeat) double cancel?
</blockquote></details>

Done.
___
*[`internal/source/fslogical/fslogical.go` line 248 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/325#-NVFhc1LDgfZhvZV2NAI:-NVHelV1-iy2VRcVWLVJ:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e42bee27792c385cf36a90e04eabd426bfd75ee9/internal/source/fslogical/fslogical.go#L248)):*

(repeat) nil check first?
</blockquote></details>

Done.
___
*[`internal/source/fslogical/fslogical.go` line 576 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/325#-NVFiTJz2nHREu9hoFXt:-NVHem_XELVwMYxTRRjF:b-9we81w) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e42bee27792c385cf36a90e04eabd426bfd75ee9/internal/source/fslogical/fslogical.go#L576)):*

Under what circumstances would this happen?
</blockquote></details>

This code was added to allow (selective) re-processing of collections as a backstop against unknown errors when querying the contents of a collection.  That is, you can just run the sync again and it'll pick up any missing documents.
___
*[`internal/source/fslogical/tombstones.go` line 106 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/325#-NVFibcPEPOeho7grNwJ:-NVHfQK77DxA3JFrnz7_:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e42bee27792c385cf36a90e04eabd426bfd75ee9/internal/source/fslogical/tombstones.go#L106)):*

(repeat) double cancel?
</blockquote></details>

Done.
___
*[`internal/source/logical/loop.go` line 253 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/325#-NVFkpEAANO_vX5mfc2c:-NVHfTcyEjsuyKP2K8wD:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e42bee27792c385cf36a90e04eabd426bfd75ee9/internal/source/logical/loop.go#L253)):*

s/least/lease
</blockquote></details>

Done.
___
*[`internal/util/stopper/stopper.go` line 161 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/325#-NVGBMsz6Q2_Yb1Dv5kw:-NVHfWbz8WNkYE4fdd6C:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e42bee27792c385cf36a90e04eabd426bfd75ee9/internal/util/stopper/stopper.go#L161)):*

Would it be better to use a RWlock so isStopping can be called with just a read lock?  Just in case you end up doing that a lot.
</blockquote></details>

Done.
___
*[`internal/util/stopper/stopper.go` line 196 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/325#-NVGBov76EE8ogecncSy:-NVHfw8lCiQRL6gLnIUN:b-2jf5a0) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e42bee27792c385cf36a90e04eabd426bfd75ee9/internal/util/stopper/stopper.go#L196)):*

Doesn't this mean that the stopping will be triggered after the stopper has been stopped + grace period, instead of waiting for it to stop?  (That was a lot of overuse of the word stop in one sentence.)

I just want to make sure the intent is clear.
</blockquote></details>

The context will be canceled after `close(stopping) + gracePeriod`.  I've added some additional comments to clarify how this will all play out.
___
*[`internal/util/stopper/stopper.go` line 232 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/325#-NVGCTD48vu4pQjP6GHx:-NVHhJBJ0FjOJrIDCxgS:brco8g4) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e42bee27792c385cf36a90e04eabd426bfd75ee9/internal/util/stopper/stopper.go#L232)):*

I see only +1s and -1s.  Do you ever see a need for >1 or <1? 
</blockquote></details>

I don't think anything other that +1 or -1 would ever be passed.

PR #329, opened 2023-05-15
Replace use of sync.Cond with channel

bobvawter:
Condition variables are hard to use well in non-trivial locking and flow control. This change replaces the uses of sync.Cond with channels that are closed to indicate that some state has been invalidated and needs to be refreshed. This pattern plays nicely with RWMutex as long as the data and its associated channel are read from the locked data within the same critical region.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/329)

### DIFF
--- a/internal/source/logical/factory.go
+++ b/internal/source/logical/factory.go
@@ -104,7 +104,7 @@ func (f *Factory) newLoop(ctx context.Context, config *BaseConfig, dialect Diale
 		running:    stopper.WithContext(ctx),
 		targetPool: f.pool,
 	}
-	loop.consistentPoint.Cond = sync.NewCond(&sync.Mutex{})
+	loop.consistentPoint.updated = make(chan struct{})
 	initialPoint, err := loop.loadConsistentPoint(ctx)
 	if err != nil {
 		return nil, err
--- a/internal/source/logical/fan_workers.go
+++ b/internal/source/logical/fan_workers.go
@@ -35,9 +35,9 @@ type fanWorkers struct {
 
 	mu struct {
 		sync.Mutex
-		data  map[ident.Table][]types.Mutation
-		drain bool       // Graceful drain condition.
-		wake  *sync.Cond // Locks on mu.
+		data    map[ident.Table][]types.Mutation
+		drain   bool          // Graceful drain condition.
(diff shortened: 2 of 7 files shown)


BramGruneir (commented):
This change worries me a bit.  If there is only one producer, not a problem, but I worry if there are multiple.  

> Reviewed 1 of 1 files at r1, 2 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/329#-:-NVVAx8AC_7Hactuf1b0:bv3r412)* status: 3 of 7 files reviewed, all discussions resolved

bobvawter:
Can you elaborate on your concern?  I'm replacing the use of `sync.Cond.Broadcast` with a channel close, per the comments in https://pkg.go.dev/sync#Cond

> For many simple use cases, users will be better off using channels than a Cond (Broadcast corresponds to closing a channel, and Signal corresponds to sending on a channel).

BramGruneir (commented):
I always just like to double check any change in locking is all.   

I was concerned if a new channel could be created in more than one place and it doesn't automatically get updated.  Just a lot of coordination is all.

For example, in logical/loop.go, there are two places that make a new channel.  I just worry that the same condition could fire in quick succession and a loop may be lost by one of the receivers.  

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/329#-:-NVVHDbd99KtbrM2rBQ2:b6glzv)* status: 3 of 7 files reviewed, all discussions resolved

BramGruneir (approved):
:lgtm:

> Reviewed 3 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/329#-:-NVVeHwz241V0vVYFkJS:bnfp4nl)* status: 6 of 7 files reviewed, all discussions resolved

BramGruneir (approved):
> Reviewed 1 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/329)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @bobvawter)

PR #332, opened 2023-05-15
workflows: Push binaries to a GCP bucket

bobvawter:
This change pushes binaries to a GCP storage bucket to provide stable URLs for downloading builds.


BramGruneir:
Lgtm

PR #336, opened 2023-05-18
target: Extract staging package

bobvawter:
This change starts the work of allowing a separate cluster to be used for staging of data, rather than assuming that the _cdc_sink database will live in the same instance as the destination database.

There are no functional changes introduced in this commit, save that the sinktest package is also split between a "base" and an "all" package, to aid in cycle-breaking.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/336)


BramGruneir (approved):
:lgtm:

> Reviewed 69 of 69 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/336#-:-NVjR_foEdhvkHyymG-_:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @bobvawter)

PR #358, opened 2023-06-07
Support for v23.1 CDC Queries for webhook and http storage based sinks.

### LINKED ISSUE #307: Support for v23.1 CDC Queries
I noticed CDC queries produce different messages in 23.1 vs ones in 22.2.x 

for example:

22.2.x
```
  {"key":"[\"user1\"]",
  "table":"usertable",
  "value":"{\"after\": 
    {
        \"field0\": \"1\", 
        \"field1\": \"2\", 
        \"field2\": \"3\", 
        \"field3\": \"4\", 
        \"field4\": \"5\", 
        \"field5\": \"6\", 
        \"field6\": \"7\", 
        \"field7\": \"8\", 
        \"field8\": \"9\",
        \"field9\": \"10\", 
        \"ycsb_key\": \"user1\"
        }
    }"
}
```

vs 23.1

```
{"key":"[\"user4\"]",
  "table":"usertable",
  "value":
    "{\"__crdb__\": 
        {\"updated\": \"1682299912954058000.0000000000\"}, 
    \"cdc_prev\": null, 
    \"field0\": \"1\", 
    \"field1\": \"2\", 
    \"field2\": \"3\", 
    \"field3\": \"4\", 
    \"field4\": \"5\", 
    \"field5\": \"6\", 
    \"field6\": \"7\", 
    \"field7\": \"8\", 
    \"field8\": \"9\", 
    \"field9\": \"10\", 
    \"updated_at\": 
    \"2023-04-24T01:31:50.478314\", 
    \"ycsb_key\": \"user4\"}"}
```

would cdc-sink get updated to support 23.1 soon?

Thanks!

sravotto:
This PR allow cdc-sink users to leverage change data capture queries (https://www.cockroachlabs.com/docs/v23.1/cdc-queries), fixes #307                

In order to provide to cdc-sink with the information required to process the data emitted by the changefeed, the CREATE CHANGEFEED definition must follow these conventions:

. The URL in the must contain the name of the table.
. The SELECT Statement must return a  "__event__" column that contains the result of the event_op() function.

Example:
```sql
CREATE CHANGEFEED  INTO             
  'webhook-https://127.0.0.1:30004/ycsb/public/usertable?insecure_tls_skip_verify=true' 
  WITH diff, updated, resolved='10s',     
       webhook_sink_config='{"Flush":{"Messages":1000,"Frequency":"1s"}}'      
      AS 
      SELECT *, event_op() as __event__
      FROM        
      YCSB.USERTABLE WHERE event_op() = 'insert' OR 
        event_op() = 'update' OR 
        event_op() = 'delete'; 
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/358)


bobvawter:
This UX is not great, because it adds another source of mis-configuration and confusing behaviors.  The necessary key structure can already be extracted from the destination table schema.  What problem are you trying to side-step by pushing the schema into the configuration?

I'm not entirely certain that the `Mutation.Key` field is truly needed at this point for anything other than propagating deletes into the target table.  There's something of an optimization when resolving timestamps where we can only upsert the latest value for a given key within the resolution window, but that's not, strictly speaking, required for correct operation.

sravotto (commented):
I reworked the  `CREATE CHANGEFEED` requirements to only have the table name into the URL and the event_op() as part of the columns emitted. This should simplify the user experience.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/358#-:-NXk56_XDxGFtIKvkfJW:b-619jly)* status: 0 of 11 files reviewed, all discussions resolved (waiting on @bobvawter)

bobvawter (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/358)* status: 0 of 11 files reviewed, 13 unresolved discussions (waiting on @sravotto)

___
*[`internal/source/cdc/handler.go` line 126 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/358#-NXp5L0_9gjAvhtwEAgD:-NXp5L0_9gjAvhtwEAgE:bbukey2) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/d7061ed468344189060ea2c53734cb3c73c90403/internal/source/cdc/handler.go#L126)):*
> ```Go
> 	body      io.Reader
> 	leaf      func(ctx context.Context, req *request) error
> 	keys      map[ident.Ident]int
> ```

Document the value, it's not obvious by inspection.
___
*[`internal/source/cdc/handler.go` line 138 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/358#-NXp5CJk8JbRhOLRNLr9:-NXp5CJk8JbRhOLRNLrA:b44nn9h) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/d7061ed468344189060ea2c53734cb3c73c90403/internal/source/cdc/handler.go#L138)):*
> ```Go
> ) (map[ident.Ident]int, error) {
> 	if r.keys == nil {
> 		table := r.target.(ident.Table)
> ```

This isn't, strictly speaking, guaranteed to work. Test the type and return an error.
___
*[`internal/source/cdc/handler.go` line 144 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/358#-NXp5Vi40ZqUKdnRCeEV:-NXp5Vi40ZqUKdnRCeEW:b-uwsvpu) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/d7061ed468344189060ea2c53734cb3c73c90403/internal/source/cdc/handler.go#L144)):*
> ```Go
> 			return nil, err
> 		}
> 		columns := watcher.Get().Columns[table]
> ```

What happens if the table is unknown?  For instance, it could be newly-created in the destination and the schema watcher hasn't refreshed yet.
___
*[`internal/source/cdc/handler_test.go` line 45 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/358#-NXp5eoU3ZPTrdgp8qnJ:-NXp5eoU3ZPTrdgp8qnK:b-q649l9) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/d7061ed468344189060ea2c53734cb3c73c90403/internal/source/cdc/handler_test.go#L45)):*
> ```Go
> 	r := require.New(t)
> 	baseFixture, cancel, err := all.NewFixture()
> 	t.Cleanup(func() {
> ```

Can't you pass `cancel` directly to `Cleanup()`?
___
*[`internal/source/cdc/ndjson.go` line 76 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/358#-NXp691l4RjgcZXc70Q1:-NXp691l4RjgcZXc70Q2:b-1n225p) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/d7061ed468344189060ea2c53734cb3c73c90403/internal/source/cdc/ndjson.go#L76)):*
> ```Go
> // information to be able to persist it to the staging table.
> func (h *Handler) parseMutation(
> 	ctx context.Context, req *request, rawBytes []byte,
> ```

Why is this a method on `Handler`?  If the extra parameters are necessary to satisfy a type constraint, but are unused, they should be named `_` and the doc comment updated.
___
*[`internal/source/cdc/ndjson.go` line 90 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/358#-NXp6Nqf2fzpsbPWxzXR:-NXp6Nqf2fzpsbPWxzXS:bze417v) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/d7061ed468344189060ea2c53734cb3c73c90403/internal/source/cdc/ndjson.go#L90)):*
> ```Go
> 		return types.Mutation{}, err
> 	}
> 	// The updated timestamp could be at the top level or inside "__crdb__" (for CDC queries).
> ```

Stale?
___
*[`internal/source/cdc/ndjson_query.go` line 64 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/358#-NXp77cL0gWVlVoYg9ed:-NXp77cL0gWVlVoYg9ee:b-97mc8t) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/d7061ed468344189060ea2c53734cb3c73c90403/internal/source/cdc/ndjson_query.go#L64)):*
> ```Go
> // parseQueryMutation takes a single line from an ndjson and extracts enough
> // information to be able to persist it to the staging table.
> // When using CDC queries the SELECT Statement must include the event (as "__event__")
> ```

Should this method check for the presence of `__event__`?  It's not really a proper `Mutation` if the value isn't there.
___
*[`internal/source/cdc/ndjson_query.go` line 71 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/358#-NXp6w39EIP9kXtreW8N:-NXp6w39EIP9kXtreW8O:b-1rzquw) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/d7061ed468344189060ea2c53734cb3c73c90403/internal/source/cdc/ndjson_query.go#L71)):*
> ```Go
> 	ctx context.Context, req *request, rawBytes []byte,
> ) (types.Mutation, error) {
> 	keys, err := req.getPrimaryKey(ctx, h.Resolvers.watchers)
> ```

`getPrimaryKey` really isn't an attribute of the `request` type.  It would make more sense to live in `Handler`, so you don't have to pass dependencies around.
___
*[`internal/source/cdc/query_payload.go` line 53 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/358#-NXp7TmoCL2_9xIr63cY:-NXp7TmoCL2_9xIr63cZ:bvyhu0p) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/d7061ed468344189060ea2c53734cb3c73c90403/internal/source/cdc/query_payload.go#L53)):*
> ```Go
> )
> 
> //go:generate go run golang.org/x/tools/cmd/stringer -type=OperationType
> ```

The type name has the wrong case here.
___
*[`internal/source/cdc/query_payload.go` line 56 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/358#-NXp7_X80_dzNhcLBXun:-NXp7_X80_dzNhcLBXuo:b38cepm) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/d7061ed468344189060ea2c53734cb3c73c90403/internal/source/cdc/query_payload.go#L56)):*
> ```Go
> 
> // decodeOp reads the operation type from a byte slice.
> func decodeOp(msg []byte) (operationType, error) {
> ```

Is msg expected to be a json-encoded string?  Should the type be `json.RawMessage`?  If so, you can skip the whole parsing step and look for the literal byte sequences that you'd expect.

Another thing to consider is that `operationType` might implement https://pkg.go.dev/encoding/json#Unmarshaler which would allow this decoding to happen as a side-effect of reifying other payloads.
___
*[`internal/source/cdc/query_payload.go` line 77 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/358#-NXp7k6_4afo-CSaa_de:-NXp7k6_4afo-CSaa_df:b-80xfat) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/d7061ed468344189060ea2c53734cb3c73c90403/internal/source/cdc/query_payload.go#L77)):*
> ```Go
> type queryPayload struct {
> 	keys      map[ident.Ident]int
> 	after     map[string]json.RawMessage
> ```

Sort.
___
*[`internal/source/cdc/query_payload.go` line 119 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/358#-NXp8Nd37H-mHvvmTqJF:-NXp8Nd37H-mHvvmTqJG:b-y80s3b) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/d7061ed468344189060ea2c53734cb3c73c90403/internal/source/cdc/query_payload.go#L119)):*
> ```Go
> 	}
> 	if _, ok := q.after[eventLabel]; !ok {
> 		return errors.New("CREATE CHANGEFEED must specify the __crdb__op_event__ colum set to op_event()")
> ```

This error message is inconsistent.  Use `errors.Errorf()` and refer to the constant.
___
*[`internal/source/cdc/query_payload.go` line 122 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/358#-NXp8iau3n_Y9KKCacsl:-NXp8iav1kRfTXPsZWn7:bftybg7) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/d7061ed468344189060ea2c53734cb3c73c90403/internal/source/cdc/query_payload.go#L122)):*
> ```Go
> 	}
> 	for k, v := range q.after {
> 		if pos, ok := q.keys[ident.New(k)]; ok {
> ```

This if block seems like it should be moved into the `default` block below.

sravotto (commented):
Thanks! . I think I addressed your suggestions. Please review.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/358#-:-NXqMFIX2lsb2jG50pmj:bv6w3r0)* status: 0 of 11 files reviewed, 13 unresolved discussions (waiting on @bobvawter)

___
*[`internal/source/cdc/handler.go` line 126 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/358#-NXp5L0_9gjAvhtwEAgD:-NXqEfFZ6OfyfrGpq7-u:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/d7061ed468344189060ea2c53734cb3c73c90403/internal/source/cdc/handler.go#L126)):*

Document the value, it's not obvious by inspection.
</blockquote></details>

Done.
___
*[`internal/source/cdc/handler.go` line 138 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/358#-NXp5CJk8JbRhOLRNLr9:-NXqEgtl4QurbnJ4XSDX:b-9erord) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/d7061ed468344189060ea2c53734cb3c73c90403/internal/source/cdc/handler.go#L138)):*

This isn't, strictly speaking, guaranteed to work. Test the type and return an error.
</blockquote></details>

Done. (moved the function to handler)
___
*[`internal/source/cdc/handler.go` line 144 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/358#-NXp5Vi40ZqUKdnRCeEV:-NXqEkZ92Oq20kOFR7hU:bs7wumx) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/d7061ed468344189060ea2c53734cb3c73c90403/internal/source/cdc/handler.go#L144)):*

What happens if the table is unknown?  For instance, it could be newly-created in the destination and the schema watcher hasn't refreshed yet.
</blockquote></details>

Done. 
___
*[`internal/source/cdc/handler_test.go` line 45 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/358#-NXp5eoU3ZPTrdgp8qnJ:-NXqEneC8hlkANsf8fsJ:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/d7061ed468344189060ea2c53734cb3c73c90403/internal/source/cdc/handler_test.go#L45)):*

Can't you pass `cancel` directly to `Cleanup()`?
</blockquote></details>

Done.
___
*[`internal/source/cdc/ndjson.go` line 76 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/358#-NXp691l4RjgcZXc70Q1:-NXqEqfV5SZYBK-c7PhO:b-llbt6g) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/d7061ed468344189060ea2c53734cb3c73c90403/internal/source/cdc/ndjson.go#L76)):*

Why is this a method on `Handler`?  If the extra parameters are necessary to satisfy a type constraint, but are unused, they should be named `_` and the doc comment updated.
</blockquote></details>

Done. Added a function type to make it a bit clearer. 
___
*[`internal/source/cdc/ndjson.go` line 90 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/358#-NXp6Nqf2fzpsbPWxzXR:-NXqEs3uANXBKsw3b-vI:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/d7061ed468344189060ea2c53734cb3c73c90403/internal/source/cdc/ndjson.go#L90)):*

Stale?
</blockquote></details>

Done.
___
*[`internal/source/cdc/ndjson_query.go` line 64 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/358#-NXp77cL0gWVlVoYg9ed:-NXqEtzc2qZbks18YAH-:bowyov9) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/d7061ed468344189060ea2c53734cb3c73c90403/internal/source/cdc/ndjson_query.go#L64)):*

Should this method check for the presence of `__event__`?  It's not really a proper `Mutation` if the value isn't there.
</blockquote></details>

it's checked in `(q *queryPayload) UnmarshalJSON`
___
*[`internal/source/cdc/ndjson_query.go` line 71 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/358#-NXp6w39EIP9kXtreW8N:-NXqF0Gz9jjlp5gF5sD8:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/d7061ed468344189060ea2c53734cb3c73c90403/internal/source/cdc/ndjson_query.go#L71)):*

`getPrimaryKey` really isn't an attribute of the `request` type.  It would make more sense to live in `Handler`, so you don't have to pass dependencies around.
</blockquote></details>

Done.
___
*[`internal/source/cdc/query_payload.go` line 53 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/358#-NXp7TmoCL2_9xIr63cY:-NXqFPf34AxDPO2RmLjg:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/d7061ed468344189060ea2c53734cb3c73c90403/internal/source/cdc/query_payload.go#L53)):*

The type name has the wrong case here.
</blockquote></details>

Done.
___
*[`internal/source/cdc/query_payload.go` line 56 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/358#-NXp7_X80_dzNhcLBXun:-NXqFXDfEJbJ7ux3XHQz:bwzp86x) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/d7061ed468344189060ea2c53734cb3c73c90403/internal/source/cdc/query_payload.go#L56)):*

Is msg expected to be a json-encoded string?  Should the type be `json.RawMessage`?  If so, you can skip the whole parsing step and look for the literal byte sequences that you'd expect.

Another thing to consider is that `operationType` might implement https://pkg.go.dev/encoding/json#Unmarshaler which would allow this decoding to happen as a side-effect of reifying other payloads.
</blockquote></details>

Done. I am going to stick with the `json.RawMessage` suggestion. 
___
*[`internal/source/cdc/query_payload.go` line 77 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/358#-NXp7k6_4afo-CSaa_de:-NXqMQPq6rdp8EvQkbxr:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/d7061ed468344189060ea2c53734cb3c73c90403/internal/source/cdc/query_payload.go#L77)):*

Sort.
</blockquote></details>

Done.
___
*[`internal/source/cdc/query_payload.go` line 119 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/358#-NXp8Nd37H-mHvvmTqJF:-NXqFo75AOjgFpp1dPAl:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/d7061ed468344189060ea2c53734cb3c73c90403/internal/source/cdc/query_payload.go#L119)):*

This error message is inconsistent.  Use `errors.Errorf()` and refer to the constant.
</blockquote></details>

Done.
___
*[`internal/source/cdc/query_payload.go` line 122 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/358#-NXp8iau3n_Y9KKCacsl:-NXqFp5JD7iaoDL1RB18:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/d7061ed468344189060ea2c53734cb3c73c90403/internal/source/cdc/query_payload.go#L122)):*

This if block seems like it should be moved into the `default` block below.
</blockquote></details>

Done.

bobvawter (approved):
:lgtm: w/ nit

> Reviewed 4 of 11 files at r1, 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/358#-:-NXuyrmW8ynM4NRfcqOZ:b-uobk8w)* status: all files reviewed, 2 unresolved discussions (waiting on @sravotto)

___
*[`internal/source/cdc/ndjson_query.go` line 87 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/358#-NXu7EwvDT6Je0KNpg6-:-NXu7EwvDT6Je0KNpg60:boaumy2) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/5e3cf8adbe5c502d2997ef9b5096da41ec3be7df/internal/source/cdc/ndjson_query.go#L87)):*
> ```Go
> // for the target table and their ordinal position within the key.
> func (h *Handler) getPrimaryKey(ctx context.Context, req *request) (map[ident.Ident]int, error) {
> 	if req.keys == nil {
> ```

Minimize indentation:

```
if req.keys != nil {
  return req.keys
}
....
```
___
*[`internal/source/cdc/query_payload.go` line 122 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/358#-NXuy2i-2hhq_ZI5w6la:-NXuy2i-2hhq_ZI5w6lb:b-juog69) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/5e3cf8adbe5c502d2997ef9b5096da41ec3be7df/internal/source/cdc/query_payload.go#L122)):*
> ```Go
> 		return errors.Errorf("CREATE CHANGEFEED must specify the %s colum set to op_event()", eventLabel)
> 	}
> 	for k, v := range q.after {
> ```

It isn't necessary to loop over `q.after`. You can do a point-lookup of `crdbLabel` and `eventLabel`, then loop over `q.keys` to extract just the necessary keys from the payload.  If any of these required keys are missing, then you know that the incoming data is incomplete.

sravotto (commented):
One more time...

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/358#-:-NXvn_P49bEpjmMC-MPi:bvx1tn9)* status: 8 of 11 files reviewed, 2 unresolved discussions (waiting on @bobvawter)

___
*[`internal/source/cdc/ndjson_query.go` line 87 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/358#-NXu7EwvDT6Je0KNpg6-:-NXvnTAz-hxEMNuVtcG0:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/5e3cf8adbe5c502d2997ef9b5096da41ec3be7df/internal/source/cdc/ndjson_query.go#L87)):*

Minimize indentation:

```
if req.keys != nil {
  return req.keys
}
....
```
</blockquote></details>

Done.
___
*[`internal/source/cdc/query_payload.go` line 122 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/358#-NXuy2i-2hhq_ZI5w6la:-NXvnXHE2MNmF3kPf5lm:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/5e3cf8adbe5c502d2997ef9b5096da41ec3be7df/internal/source/cdc/query_payload.go#L122)):*

It isn't necessary to loop over `q.after`. You can do a point-lookup of `crdbLabel` and `eventLabel`, then loop over `q.keys` to extract just the necessary keys from the payload.  If any of these required keys are missing, then you know that the incoming data is incomplete.
</blockquote></details>

Done.

PR #375, opened 2023-06-22
cdc-sink: Use a sql.DB instance for target package

bobvawter:
This commit enables a split-mode behavior, where the target database can be
separate from the staging database. The bulk of this change is plumbing, where
the target database is accessed via a sql.DB type, to facilitate future
mixed-product deployments.

Summary of changes:
* types.StagingPool and types.TargetPool are now used only as pointer
  types. They have gained some additional metadata and they'll probably gain
  additional functionality in the future. To ensure correct use going forwards,
  the noCopy idiom has been added to these types.
* types.StagingPool now contains a sql.DB. Code that uses StagingPool has been
  updated to use the "FooContext" method names.
* Database pool setup code has been centralized in the stdpool package, and
  that package's API enlarged to support different database providers
  and configuration strategies.  This will make it straightforward to add
  additional database drivers in the future without having to operate at
  the lowest-common-denominator of the sql.DB type.
* The sinktest.DBInfo type has been removed, since it is redundant with the
  additional information carried by types.StagingPool and types.TargetPool.
* The sinktest.TableInfo type is now parameterized with the database pool
  that it lives in.  This allows us to leverage the type system to ensure that
  we don't accidentally cross the streams.
* Smoke tests for the sinktest "base" and "all" packages have been added
  to facilitate future changes in database setups.
* A generic union type, types.AnyPool, has been added to allow the retry
  package to operate over multiple pool types.
* The CDC webhook handler has some code rearrangement for the immediate-mode
  versus transactional case, since the pool types are now separate.
* types.Querier was renamed to types.StagingQuerier and a new
  types.TargetQuerier interface is added to provide some generality over
  database pools versus transactions.
* A new CLI flag, --stagingConn, has been added to allow split-mode operation
  to be configured.

This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/375)


BramGruneir (approved):
:lgtm:

> Reviewed 75 of 75 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/375#-:-NYyMJ793Y7UsUnjUv8C:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @bobvawter and @sravotto)

___
*[`internal/source/logical/config.go` line 131 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/375#-NYxMu1_5hMzjJ4Q4LbE:-NYxMu1_5hMzjJ4Q4LbF:boe0vuy) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/8317e1924ec0d628c9cb4032a1cacdc88b6d9f3c/internal/source/logical/config.go#L131)):*
> ```Go
> 		"the staging cluster's connection string; inherits from targetConn if set")
> 	f.StringVar(&c.TargetConn, "targetConn", "",
> 		"the target cluster's connection string; inherits from stagingConn if set")
> ```

Perhaps we should check if staging is cockroach?  You probably do this elsewhere.  

Let's not allow one or the other.  Always require target to be set, and if it's not cockroach, require staging to be set and ensure it is cockroach.
___
*[`internal/source/logical/provider.go` line 97 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/375#-NYxOWUEDNVFtw0RM-TO:-NYxOWUEDNVFtw0RM-TP:bp5cbnv) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/8317e1924ec0d628c9cb4032a1cacdc88b6d9f3c/internal/source/logical/provider.go#L97)):*
> ```Go
> ) (*types.StagingPool, func(), error) {
> 	options := []stdpool.Option{
> 		stdpool.WithConnectionLifetime(5 * time.Minute),
> ```

Maybe worth adding the ability to override these via a config at some point in the future, not now.
___
*[`internal/source/logical/provider.go` line 124 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/375#-NYxOfqF7ZgKghQUffPV:-NYxOfqF7ZgKghQUffPW:b-6li8ek) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/8317e1924ec0d628c9cb4032a1cacdc88b6d9f3c/internal/source/logical/provider.go#L124)):*
> ```Go
> 	options := []stdpool.Option{
> 		stdpool.WithConnectionLifetime(5 * time.Minute),
> 		stdpool.WithMetrics("target"),
> ```

These too.  Maybe worth adding the ability to override these via a config at some point in the future, not now.
___
*[`internal/util/retry/retry.go` line 47 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/375#-NYyJ2u0-AtoCMZDGA2C:-NYyJ2u0-AtoCMZDGA2D:bqvrbwa) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/8317e1924ec0d628c9cb4032a1cacdc88b6d9f3c/internal/util/retry/retry.go#L47)):*
> ```Go
> 		switch t := any(db).(type) {
> 		case *types.StagingPool:
> 			_, err = t.Exec(ctx, query, args...)
> ```

Is there a reason the staging pool doesn't also have ExecContext (or vice versa)?

bobvawter (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/375)* status: all files reviewed, 2 unresolved discussions (waiting on @BramGruneir and @sravotto)

___
*[`internal/source/logical/config.go` line 131 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/375#-NYxMu1_5hMzjJ4Q4LbE:-NYylnPiBio0aAarDR5n:b-inybe5) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/8317e1924ec0d628c9cb4032a1cacdc88b6d9f3c/internal/source/logical/config.go#L131)):*

Perhaps we should check if staging is cockroach?  You probably do this elsewhere.  

Let's not allow one or the other.  Always require target to be set, and if it's not cockroach, require staging to be set and ensure it is cockroach.
</blockquote></details>

Updated the tests and flag strings to say that target is required and staging is required for non-CRDB targets.  If the staging db isn't actually CRDB, the staging pool will fail to initialize later on.
___
*[`internal/util/retry/retry.go` line 47 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/375#-NYyJ2u0-AtoCMZDGA2C:-NYym6cO8HwgX01u-aMO:bgkzti0) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/8317e1924ec0d628c9cb4032a1cacdc88b6d9f3c/internal/util/retry/retry.go#L47)):*

Is there a reason the staging pool doesn't also have ExecContext (or vice versa)?
</blockquote></details>

`StagingPool` is a `pgxpool.Pool` and `TargetPool` is a `sql.DB`.  The names differ for methods that accept a `Context`, since the `context` package was a late addition to the go stdlib.

PR #380, opened 2023-06-26
workflow: Create status objects in-line

bobvawter:
The status meta-workflow hasn't been firing reliably for PRs in the merge
queue. This change moves the status-check creation into the two workflows we
use for branch protection. It also makes use of the "gh api" command,
rather than using curl.

This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/380)


BramGruneir (approved):
:lgtm:

I'm no expert on these, but the changes look sane.
I'm going to assume that if something isn't working right we shall detect and fix it.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/380#-:-NYyMaPw-a0Ass8AxGvE:b-67ppwo)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sravotto)

PR #385, opened 2023-06-27
workflow: Don't test on push

bobvawter:
We use the merge queue to admit changes, so there's no need to re-test once the commit hits a destination branch.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/385)


BramGruneir:
Lgtm

PR #386, opened 2023-06-28
stdpool: Add support for Oracle targets

bobvawter:
This change adds initial support for connecting to an Oracle database as a target.

At present, only the sinktest/base package test will pass if the following flag is set: -testTarget oracle://system:SoupOrSecret@127.0.0.1:1521/XE

Summary of changes:
* Docker-Compose configurations added.
* The stdpool methods return Pool types.
* A types.Product enum is added to the Pool types.
* stdpool.OpenOracleAsTarget is added.
* stdpool.OpenTarget is added to parse URLs and select an implementation.
* Additional checks are added to the logical package to check for CockroachDB.

Developer note: At present, Docker Desktop running on Apple M-series processors does not present an x86_64 processor running in a 64-bit opmode. This prevents the Oracle docker containers from initializing correctly. As a workaround, the colima project is suggested as a VM host since it provides a Docker-compatible API for docker-compose to use. Additionally, the v18.4 image fails to initialize correctly, since the embedded Java runtime experiences a fatal internal fault.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/386)


BramGruneir (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/386#-:-NZ2q6hC9hC52JAAca0J:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sravotto)

___
*[`internal/util/stdpool/pgx.go` line 133 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/386#-NZ2pyEv9kEp8jHE0QWr:-NZ2pyEv9kEp8jHE0QWs:bkvd0p4) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/46b0871e0e8491208bf0a37b602c7609fb831036/internal/util/stdpool/pgx.go#L133)):*
> ```Go
> 	case strings.HasPrefix(ret.Version, "CockroachDB"):
> 		ret.Product = types.ProductCockroachDB
> 	case strings.HasPrefix(ret.Version, "PostgreSQL"):
> ```

I guess we're going to support postgres as a target?  Seems reasonable.

PR #389, opened 2023-07-06
ident: Rework ident package

bobvawter:
This change is motivated by the need to support products that do not have a
two-level schema namespace. The assumption about table naming structure is
pervasive in the cdc-sink code, so this change overhauls the ident package to
clean up long-standing deficiencies in the package design.

The public ident types (e.g. Table, Schema, Ident) remain comparable
value-types. Their internals have changed such that they merely contain a
pointer to a canonical instance of a package-internal type (qualified, array,
atom). This two-level setup allows types that are essentially the same (e.g.
Table and UDF) to share the bulk of their implementation by embedding one or
more of the building-block types.

There are some breaking changes in the _cdc_sink table schemas, to remove the
concept of a target database.  Instead, we only track a target schema, which
may be represented by a dotted string.

The HTTP handlers also receive some updates so that they are aware of whether
or not the target product has a 1- or 2-level namespace, so as to preserve the
existing URL design for CockroachDB targets. The scattered "parseSomethingURL"
methods are unified in the cdc/request.go file.

Most other changes throughout the code are trivial changes for the revised API
or to rename fields named "SomethingDB" which now contain an ident.Schema
value.

Recommended package review order:
- util/ident (New API and canonicalization)
- source/cdc (Unified URL parsing)
- target/schemawatch (Schema-only API design)
- auth/jwt (breaking schema change)
- staging/stage (breaking schema change)


bobvawter (commented):
What do you think about using a JSON-formatted table comment for cdc-sink to track schema versions in?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/389#-:-N_4VWBQ7OfmpGsnMQXx:b-8l39b0)* status: 67 of 85 files reviewed, 23 unresolved discussions (waiting on @BramGruneir and @sravotto)

___
*[`internal/source/cdc/request.go` line 172 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/389#-NZkrtYhBSmxbncUf2k8:-N_48Qif-gvUscfIX2SB:by4xu4m) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/9b513c4361b8e3679668068e5caf27491b711c86/internal/source/cdc/request.go#L172)):*

This probably needs a unit test.  Logic seems sound.
</blockquote></details>

It's tested in `url_test.go::TestParseChangefeedURL()`. `parseUrl()` is called by way of `*handler.newRequest()` to verify that timestamps are extracted from bulk-style endpoints.
___
*[`internal/source/cdc/request.go` line 193 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/389#-NZks6iu3xhPX7lpTmkl:-N_4A6Uy8KeQC5PgsgnZ:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/9b513c4361b8e3679668068e5caf27491b711c86/internal/source/cdc/request.go#L193)):*

Unfinished comment.
</blockquote></details>

Done.
___
*[`internal/source/cdc/request.go` line 194 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/389#-NZksHdI7e_bUttfQPKT:-N_49DMFDENKDWssTRXx:b-u34o1x) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/9b513c4361b8e3679668068e5caf27491b711c86/internal/source/cdc/request.go#L194)):*

Would a regex be faster?
</blockquote></details>

It's a one-pass examination with no state, so a regex seems overkill. 
___
*[`internal/source/cdc/request.go` line 212 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/389#-NZksiK21rfD9Cd-CUur:-N_4Ab8W1M8iu4QIfVrw:b-6y3t61) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/9b513c4361b8e3679668068e5caf27491b711c86/internal/source/cdc/request.go#L212)):*

Perhaps it's worth turning requestPatterns into a struct to do this logic.  It would allow for easier unit testing I think.
</blockquote></details>

`requestPatterns` contains what were helper functions inside of `request.parseUrl()`.  That method was getting way too long, so I hoisted the repetitive code flow into the `requestPattern` helper.
___
*[`internal/source/cdc/request.go` line 216 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/389#-NZkzsmQ7EO0Xo_5iJqL:-N_4BNeFEbGob1UOBVqT:bja0i46) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/9b513c4361b8e3679668068e5caf27491b711c86/internal/source/cdc/request.go#L216)):*

Perhaps store the patterns in a map of expectedPathSegments to an array of patterns that match?
</blockquote></details>

Are you talking about building like an NFA and tracking concurrent states?
___
*[`internal/staging/auth/jwt/jwt.go` line 216 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/389#-N_-0EwX42rgrb8IIs2f:-N_4BafU9jf5nNCm_2AW:bp7i713) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/9b513c4361b8e3679668068e5caf27491b711c86/internal/staging/auth/jwt/jwt.go#L216)):*

Maybe I missed it, but this could use a unit test.
</blockquote></details>

`jwt_test.go::TestMatches()` predates this PR.
___
*[`internal/staging/auth/jwt/jwt.go` line 222 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/389#-N_--dR592HLk-voyc05:-N_4BnNz9R0Pd5TL4fU9:b-g5dt1h) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/9b513c4361b8e3679668068e5caf27491b711c86/internal/staging/auth/jwt/jwt.go#L222)):*

combine these two lines?  Maybe one if statement?
</blockquote></details>

The semantic meaning of the if statement was inscrutable, so I added a local to act as a label for the intermediate value.  Also, reading `if not part matches` seems more clear than DeMorgan'ing the logic.
___
*[`internal/staging/auth/jwt/jwt.go` line 229 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/389#-N_-01NF9yfjLIzp_3n6:-N_4CYUv4BtVo7vGveeo:b-b80scj) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/9b513c4361b8e3679668068e5caf27491b711c86/internal/staging/auth/jwt/jwt.go#L229)):*

Why not move this up to be a quick exit if either is 0?
</blockquote></details>

If either is zero, the `for` statement won't trigger.  I'd still need exactly this test in the `return` statement to detect a case where the loop didn't consume every element of both the allowed and requested Schema.
___
*[`internal/target/schemawatch/dependencies.go` line 91 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/389#-N_-289Y80u5HaQwf190:-N_4Nw1XC5n9KD-tghoD:bwkzsti) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/9b513c4361b8e3679668068e5caf27491b711c86/internal/target/schemawatch/dependencies.go#L91)):*

You do this a lot.  Perhaps add a helper function?  Just so you don't have to add the comment every time.
</blockquote></details>

Added `Identifier.Split() (first Ident, remainder Identifier)` to handle cases where you'd want to incrementally decompose an identifier.
___
*[`internal/util/ident/array.go` line 28 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/389#-N_-5R5t71PfbP7hQH3-:-N_4ONSpC5prfjzh0ov_:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/9b513c4361b8e3679668068e5caf27491b711c86/internal/util/ident/array.go#L28)):*

Unfinished comment?
</blockquote></details>

Done.
___
*[`internal/util/ident/array.go` line 42 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/389#-N_-5mdABNbwKN2Etqqg:-N_4OfL41I0-y1n0Yhwy:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/9b513c4361b8e3679668068e5caf27491b711c86/internal/util/ident/array.go#L42)):*

Should this be based on max array length?
</blockquote></details>

Done.
___
*[`internal/util/ident/atom.go` line 27 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/389#-N_-6DA53HzVKKG6zOKl:-N_4Oupn3aDYERm-gy-h:b-31akss) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/9b513c4361b8e3679668068e5caf27491b711c86/internal/util/ident/atom.go#L27)):*

q is quoted and r is raw, why not just name them appropriately?
</blockquote></details>

Done. But, `go`
___
*[`internal/util/ident/atom.go` line 42 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/389#-N_-757A1y-uODF2Gezp:-N_4Oz_T0J4U4FQfpvAk:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/9b513c4361b8e3679668068e5caf27491b711c86/internal/util/ident/atom.go#L42)):*

shouldn't all of these be `(a *atom)` instead of `(i *atom)`
</blockquote></details>

Done.
___
*[`internal/util/ident/atom.go` line 84 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/389#-N_-7TkZAlFemqA0LIz3:-N_4P8eT1XrjFMU6ZVmx:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/9b513c4361b8e3679668068e5caf27491b711c86/internal/util/ident/atom.go#L84)):*

Could you drop this in a common file?  Instead of hiding it here?
</blockquote></details>

Done.
___
*[`internal/util/ident/canonical_map.go` line 56 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/389#-N_-8XirDHEEr6F9Q3FB:-N_4PiGY5c6zieGLV1p2:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/9b513c4361b8e3679668068e5caf27491b711c86/internal/util/ident/canonical_map.go#L56)):*

Can you add a comment here on how this check has to be repeated to avoid race conditions.  I was worried when I saw the two lock sections.
</blockquote></details>

Done.
___
*[`internal/util/ident/identifier.go` line 56 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/389#-N_-9w_hARLTQvEcwi-e:-N_4Qe59-HqpHciG-VbL:bymzhfr) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/9b513c4361b8e3679668068e5caf27491b711c86/internal/util/ident/identifier.go#L56)):*

Is it worth turning the raw bool into an enum?  Just for readability?  I'm neutral on if it makes sense to or not.
</blockquote></details>

Added `type Representation bool` and two constants to improve readability.
___
*[`internal/util/ident/identifier.go` line 56 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/389#-N_-AiyO93eS8_dh5Hxh:-N_4Qql27oNHALFdbXEl:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/9b513c4361b8e3679668068e5caf27491b711c86/internal/util/ident/identifier.go#L56)):*

x is not great here, maybe id?
</blockquote></details>

Done.
___
*[`internal/util/ident/identifier.go` line 75 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/389#-N_-Arfy19vBdT1qfsIF:-N_4QtNF0XCHPrEdJXgC:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/9b513c4361b8e3679668068e5caf27491b711c86/internal/util/ident/identifier.go#L75)):*

also x here, perhaps id?
</blockquote></details>

Done.
___
*[`internal/util/ident/parse.go` line 175 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/389#-N_-CG9vA1TAWRYQnDV0:-N_4Qycy9ttvevchYPbX:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/9b513c4361b8e3679668068e5caf27491b711c86/internal/util/ident/parse.go#L175)):*

You usually use `s string`, not `str string`.  Both are fine, but perhaps choose one for consistency?
</blockquote></details>

Done.
___
*[`internal/util/ident/parse.go` line 184 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/389#-N_-Bk860eQDkZjvDHs5:-N_4R3X78GbGwW77yp5k:b-666giy) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/9b513c4361b8e3679668068e5caf27491b711c86/internal/util/ident/parse.go#L184)):*

Do you mean to append here?  What's the point of pre-allocating of that's the case. 
</blockquote></details>

`parts` is stack-allocated since the capacity is a compile-time constant.  The call to `append()` here merely increments the length in the slice header.  Otherwise, the slice could be allocated at maximum length and then shortened before returning it.
___
*[`internal/util/ident/schema.go` line 33 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/389#-N_-FYbK87Zan2yvs0L3:-N_4RcYl7FkAMwse1cW7:btd1peo) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/9b513c4361b8e3679668068e5caf27491b711c86/internal/util/ident/schema.go#L33)):*

Should this be called MustNewSchema?  I don't remember the right go idiom.  
</blockquote></details>

Packages that expose a helper like this (e.g. `regexp.Must()`) usually have only a primary, exported type.  This package has several user-visible types, so I split the difference on typing.
___
*[`internal/util/ident/schema.go` line 63 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/389#-N_-GfLzAIaFCYDDZFp2:-N_4S9hfC_ABT8NCpZzI:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/9b513c4361b8e3679668068e5caf27491b711c86/internal/util/ident/schema.go#L63)):*

training should be trailing
</blockquote></details>

Done.
___
*[`internal/util/ident/table.go` line 62 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/389#-N_-HxYR019Z2iPMIp-X:-N_4Sk6K-ZjZ-2h1_jC4:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/9b513c4361b8e3679668068e5caf27491b711c86/internal/util/ident/table.go#L62)):*

Could you add a sample of what the json looks like for each case as a comment?
</blockquote></details>

Done.

BramGruneir (approved):
I was thinking a single table with a version column.  But yes, a table comment or DB comment would be fine as well.

:lgtm: 

But let's iron out the breaking change check before merging.

> Reviewed 18 of 18 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/389#-:-N_4gyHE1lG8kzoys9Wz:beiet71)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sravotto)

___
*[`internal/source/cdc/request.go` line 216 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/389#-NZkzsmQ7EO0Xo_5iJqL:-N_4fkm03oYHyNT0Usb-:b5uj95q) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/9b513c4361b8e3679668068e5caf27491b711c86/internal/source/cdc/request.go#L216)):*

Are you talking about building like an NFA and tracking concurrent states?
</blockquote></details>

Nothing that crazy.  I was just thinking of a smaller loop.

So you only need to loop over the patterns that could match, instead of all of them.  
mapSegCountsToPatterns[int]requestPattern
You would have to also store if it was the +1 or version or not.

But this would be a micro-optimizaiton, so probably not worth it.

bobvawter:
PTAL, rebased this change onto PR #400 that adds a versioning mechanism.

BramGruneir (changes requested):
> Reviewed 15 of 15 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/389)* status: all files reviewed, 1 unresolved discussion (waiting on @bobvawter and @sravotto)

___
*[`-- commits` line 39 at r3](https://reviewable.io/reviews/cockroachdb/cdc-sink/389#-N_Fo-AJEEW5E5eFGNQe):*
Can you put these instructions in a file under upgrade or some other directory name?  That way people won't need to look up the commit message (and we could even include the steps in the version output if we wanted.

BramGruneir:
Oh, we should probably test that the required schema change does indeed work if cdc-sink is stopped and then upgraded.  Not an automated test, just manually to ensure that nothing gets dropped.

bobvawter:
I added a `migrations` directory to accumulate these breaking-change instructions.  I did also test the SQL commands in the migration.

BramGruneir (approved):
:lgtm:

> Reviewed 25 of 88 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/389#-:-N_GXIVA6acQuqazIzz6:bnfp4nl)* status: 35 of 98 files reviewed, 1 unresolved discussion (waiting on @bobvawter and @sravotto)

PR #404, opened 2023-07-24
sinktest: Separate source from target test schemas

bobvawter:
This PR separates test code that depends on a "source" database from the
existing target database. In the past, the source, staging, and target database
were all the same CockroachDB instance, so there was never a need to make this
distinction in the test code. This change enables non-CockroachDB targets to be
tested.

The source, staging, and target setup in this PR would also allow us to move
cdc-sink in a direction where we test against older versions of CockroachDB as
a source, but we only verify that the staging and target behaviors work on more
recent versions.

Summary of changes:
- `sinktest.base` looks for three environment variables to determine what
  database(s) to connect to.
- Each of source, staging, and test have distinct table schemas for each test.
  These are available through new injection types, `sinktest.SourceSchema` and
  `sinktest.TargetSchema`. The `TestDB` injection point is removed.
- `types.StagingDB` is renamed to `StagingSchema` for consistency.
- `types.SourcePool` is added for tests, but it could be used if we add the
  proposed S4 (stupid, simple SELECT *) frontend.
- The Fixture.CreateTable method is bifurcated into CreateSourceTable and
  CreateTargetTable. These return types that are parameterized with their
  pool-ness, so the typesystem gives us a hand in avoiding cases where we might
  cross the streams.
- The Golang workflow adds a test-matrix entry to test completely separated
  source, staging, and target databases. The docker container logs are also
  captured, as this was necessary to troubleshoot some startup issues.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/404)

### DIFF
--- a/.github/docker-compose.yml
+++ b/.github/docker-compose.yml
@@ -24,6 +24,9 @@
 # as the values passed to sinktest.IntegrationMain.
 version: "3.9"
 services:
+  # These define the tested source databases. The network_mode is host
+  # so that outgoing changefeed network requests can connect to the test
+  # server.
   cockroachdb-v20.2:
     image: cockroachdb/cockroach:latest-v20.2
     network_mode: host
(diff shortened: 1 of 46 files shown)


BramGruneir (approved):
:lgtm:

> Reviewed 46 of 46 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/404#-:-NaHfuZc0Dm_cjhKzlQj:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sravotto)

PR #417, opened 2023-08-09
cmap: Add utility for canonicalized mappings

bobvawter:
This package contains a Map implementation where the keys aren't required to be comparable.  Instead, the Map has a mapper function which converts a key into a comparable representation that is used to locate a value.

The motivation for this change is to readily support case-preserving, but case-insentivite identifier mappings.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/417)

### DIFF
--- a/internal/util/cmap/cmap.go
+++ b/internal/util/cmap/cmap.go
@@ -1,0 +1,160 @@
+// Copyright 2023 The Cockroach Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+// Package cmap contains an implementation of a canonicalizing map.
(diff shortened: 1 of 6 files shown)


BramGruneir (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/417#-:-NbQ41Lg39_QNK5pNg9P:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @bobvawter and @sravotto)

___
*[`internal/util/cmap/cmap_test.go` line 109 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/417#-NbQ2g4E5dMXIfyILEGm:-NbQ2g4E5dMXIfyILEGn:bc4rm8n) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e8c193884765343e4a2bec33bc3efbabaa2d50f7/internal/util/cmap/cmap_test.go#L109)):*
> ```Go
> 		{"2", '2'},
> 		{"3", '3'},
> 	}, entries)
> ```

Test for delete?
GetZero?
___
*[`internal/util/cmap/identity_test.go` line 77 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/417#-NbQ3i_r5saOu4Gg9oQr:-NbQ3i_r5saOu4Gg9oQs:b-kevc0b) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e8c193884765343e4a2bec33bc3efbabaa2d50f7/internal/util/cmap/identity_test.go#L77)):*
> ```Go
> 	m.CopyInto(cpy)
> 	r.Equal(m.Len(), cpy.Len())
> }
> ```

Missing tests for delete and GetZero

PR #418, opened 2023-08-10
ident: Make identifiers case-preserving, case-insensitive

bobvawter:
This change makes cdc-sink treat identifiers in a case-insensitive manner, while preserving the case. This will help us in casese where two different databases have differing case rules. It also brings the user experience closer to what one might normally expect from a SQL database.

This change assumes that no reasonable schema has two columns or tables which are distinguishable solely by their case.

The ident package is changed in the following ways:
* The internal types (atom, array, qualified) now maintain a pointer to a lower-cased version of the identifier. We continue to intern all incoming identifiers, so the lower-casing happens exactly once for any unique identifier.
* New Map types, for Ident, Schema, and Table, are added which provide case-insensitive treatments of identifiers. It would be straightforward, should the need arise, to use build flags to replace the case-insensitive implementation with a case-sensitive one.
* The exported ident types have a new zero-length field member added called noCompare. The noCompare type is defined in a file with build rules and will prevent the identifier types from being treated as comparable by the go compiler when tests are run. This forces future developers to only use the ident.Map types, ident.Compare, and ident.Equals.
* applycfg.Config had relied upon comparable-type behavior, so it gets a proper Equals method.
* The remaining changes throughout the code base are mechanical in order to make the code compile again.
* The test code for schema-related APIs is updated to pass jumbled-case identifiers. The cdc handler code also uses jumbled-case JSON payloads and URLs to ensure good coverage.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/418)

### DIFF
--- a/internal/script/provider.go
+++ b/internal/script/provider.go
@@ -123,8 +123,8 @@ func ProvideUserScript(
 	if boot == nil {
 		// Un-configured case, return a dummy object.
 		return &UserScript{
-			Sources: map[ident.Ident]*Source{},
-			Targets: map[ident.Table]*Target{},
+			Sources: &ident.Map[*Source]{},
+			Targets: &ident.TableMap[*Target]{},
 		}, nil
 	}
 
@@ -134,8 +134,8 @@ func ProvideUserScript(
 	}
 
 	ret := &UserScript{
-		Sources: map[ident.Ident]*Source{},
-		Targets: map[ident.Table]*Target{},
+		Sources: &ident.Map[*Source]{},
+		Targets: &ident.TableMap[*Target]{},
 		rt:      boot.rt,
 		target:  target.AsSchema(),
 		watcher: watcher,
@@ -151,10 +151,11 @@ func ProvideUserScript(
 			return errors.WithStack(err)
 		}
 		defer func() { _ = tx.Rollback(ctx) }()
-		for tbl, tblCfg := range ret.Targets {
-			if err := applyConfigs.Store(ctx, tx, tbl, &tblCfg.Config); err != nil {
-				return errors.WithStack(err)
-			}
+		if err := ret.Targets.Range(func(tbl ident.Table, tblCfg *Target) error {
+			err := applyConfigs.Store(ctx, tx, tbl, &tblCfg.Config)
+			return errors.WithStack(err)
+		}); err != nil {
+			return err
 		}
 		if err := tx.Commit(ctx); err != nil {
 			return errors.WithStack(err)
--- a/internal/script/script.go
+++ b/internal/script/script.go
@@ -35,15 +35,15 @@ import (
 // to some number of downstream tables. Dispatch functions are
 // internally synchronized to ensure single-threaded access to the
 // underlying JS VM.
-type Dispatch func(ctx context.Context, mutation types.Mutation) (map[ident.Table][]types.Mutation, error)
+type Dispatch func(ctx context.Context, mutation types.Mutation) (*ident.TableMap[[]types.Mutation], error)
 
 // dispatchTo returns a Dispatch which assigns all mutations to the
 // given target.
 func dispatchTo(target ident.Table) Dispatch {
-	return func(_ context.Context, mut types.Mutation) (map[ident.Table][]types.Mutation, error) {
-		return map[ident.Table][]types.Mutation{
-			target: {mut},
-		}, nil
+	return func(_ context.Context, mut types.Mutation) (*ident.TableMap[[]types.Mutation], error) {
+		ret := &ident.TableMap[[]types.Mutation]{}
+		ret.Put(target, []types.Mutation{mut})
+		return ret, nil
 	}
 }
 
@@ -82,8 +82,8 @@ type Target struct {
 // UserScript encapsulates a user-provided configuration expressed as a
 // JavaScript program.
 type UserScript struct {
-	Sources map[ident.Ident]*Source
-	Targets map[ident.Table]*Target
+	Sources *ident.Map[*Source]
+	Targets *ident.TableMap[*Target]
 
 	rt      *goja.Runtime // The JavaScript VM. See execJS.
 	rtMu    sync.Mutex    // Serialize access to the VM.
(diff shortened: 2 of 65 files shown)


BramGruneir (approved):
:lgtm:

This was fairly big.  I did my best to make sure it all made sense.

> Reviewed 65 of 65 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/418#-:-NbZkQvN4V-LO-vfPLbw:b1n5r4w)* status: all files reviewed, 8 unresolved discussions (waiting on @bobvawter and @sravotto)

___
*[`internal/source/cdc/ndjson_query.go` line 61 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/418#-NbVCwpG54hAsYfgC4gh:-NbVCwpG54hAsYfgC4gi:b-5h827i) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/db276798ece31a041029958445ef1a688bcc413b/internal/source/cdc/ndjson_query.go#L61)):*
> ```Go
> 	table, ok := req.target.(ident.Table)
> 	if !ok {
> 		return req.keys, errors.Errorf("expecting ident.Table, got %T", req.target)
> ```

why are you returning req.keys when you know it's going to be nil?  why not just return nil?
___
*[`internal/source/fslogical/provider.go` line 110 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/418#-NbVFZwQ1Q-YNCu4tP0g:-NbVFZwQ1Q-YNCu4tP0h:biwf287) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/db276798ece31a041029958445ef1a688bcc413b/internal/source/fslogical/provider.go#L110)):*
> ```Go
> 	})
> 
> 	return ret, loops.Close, err
> ```

should ret and loops return if err isn't nil?  
I thought it's just best practice to return the rest as nil if there is an error.
___
*[`internal/staging/applycfg/conf.go` line 74 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/418#-NbVN-QS8Kji0ba25Ty4:-NbVN-QS8Kji0ba25Ty5:b-6h0570) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/db276798ece31a041029958445ef1a688bcc413b/internal/staging/applycfg/conf.go#L74)):*
> ```Go
> 
> // Equal returns true if the other Config is equivalent to the receiver.
> func (t *Config) Equal(o *Config) bool {
> ```

This whole thing could be one big return statement I think.  Might be easier to read.
___
*[`internal/types/types.go` line 206 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/418#-NbVNdorDArH7oVFsjJC:-NbVNdorDArH7oVFsjJD:b-t3pk1s) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/db276798ece31a041029958445ef1a688bcc413b/internal/types/types.go#L206)):*
> ```Go
> // Equal returns true if the two ColData are equivalent under
> // case-insensitivity.
> func (d ColData) Equal(o ColData) bool {
> ```

This could just be a return statement.  Not sure if it would be more readable.
___
*[`internal/util/ident/atom.go` line 26 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/418#-NbVbD2Q9UmL3uWDx66T:-NbVbD2Q9UmL3uWDx66U:b-jz6o5n) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/db276798ece31a041029958445ef1a688bcc413b/internal/util/ident/atom.go#L26)):*
> ```Go
> // An atom is a canonicalized instance of a quoted identifier. They
> // should only ever be constructed via [atoms]. An atom also contains a
> // pointer to a lower-cased representation of the atom (possibly itself)
> ```

We probably need to add this as a limitation in our docs.
___
*[`internal/util/ident/compare.go` line 22 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/418#-NbZdqFn1BF2L36m8TKW:-NbZdqFn1BF2L36m8TKX:b-pb9rjq) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/db276798ece31a041029958445ef1a688bcc413b/internal/util/ident/compare.go#L22)):*
> ```Go
> 
> // Compare returns a value less than, equal to, or greater than zero if
> // a should sort before, equal, or after b.
> ```

This compare is a bit weird, can you add some basic examples please.
___
*[`internal/util/ident/compare.go` line 24 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/418#-NbZdZJ-2CSzvc6L5T8M:-NbZdZJ-2CSzvc6L5T8N:b63ejtv) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/db276798ece31a041029958445ef1a688bcc413b/internal/util/ident/compare.go#L24)):*
> ```Go
> // a should sort before, equal, or after b.
> func Compare(a, b Identifier) int {
> 	aParts := a.Idents(make([]Ident, 0, maxArrayLength+1))
> ```

Why do you need to make a copy of them at all?  Can you not get the length and depth parts directly from Identifier?
___
*[`internal/util/ident/compare.go` line 30 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/418#-NbZcfVgBtONHv_KQGNJ:-NbZcfVgBtONHv_KQGNK:bi518my) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/db276798ece31a041029958445ef1a688bcc413b/internal/util/ident/compare.go#L30)):*
> ```Go
> 
> 	for i := 0; i < aLen && i < bLen; i++ {
> 		aPart := aParts[i]
> ```

Do you need these locals?    Can't you just access aParts[i].Canonical().Raw()?
___
*[`internal/util/ident/map.go` line 192 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/418#-NbZi-Nt-CI4DS2cJo6g:-NbZi-Nt-CI4DS2cJo6h:b-gsb6xu) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/db276798ece31a041029958445ef1a688bcc413b/internal/util/ident/map.go#L192)):*
> ```Go
> }
> 
> // PutErr can be passed to Range to copy the contents of one Map into
> ```

This description is incorrect.

bobvawter:
https://github.com/cockroachdb/cdc-sink/wiki/Case-Preservation

BramGruneir:
:lgtm:

PR #419, opened 2023-08-11
target/apply: Initial Oracle support

bobvawter:
This change adds initial support for applying mutations to an Oracle database.
The automated build tests against an Oracle 21 Express container. It has been
manually tested against v19 as well, but there's additional work needed to
produce a ready-to-run container for either v18 or v19.

Notable changes:
* Fixed a case-sensitivity issue in the user-script engine.
* Updated tests in script, cdc, and logical to use Oracle-compatible types
  and read-back queries.
* The apply package had some logic that was split and duplicated between
  templates.go and apply.go. All schema metadata calculations now live in
  column_mapping.go, which templates.go makes use of.
* Some schema computations were being run once per input row, they are now run
  once per schema-refresh cycle in column_mapping.go
* Postgres templates are moved to their own subdirectory and are unchanged.
* Oracle templates are added in their own subdirectory. The Oracle setup uses
  the same template for basic upserts and conditional situations, since it's
  really just a matter of adding additional CTE clauses.
* The construction of the templates cache now depends on the target product and
  has a few product-specific behaviors added, but is essentially identical to
  the previous version.
* The new templates are tested with golden-output tests that are copied from
  the PG tests.
* Oracle schema queries now also extract datatype lengths and precisions where
  they are a mandatory feature of the SQL grammar.
* The concept of a "parse helper" is added to ColData. This is a function that
  will take the input JSON string encoding and produce an in-memory
  representation of the data that can be sent to the target database. In this
  initial implementation, we do this to support temporal datatypes as well as a
  special case to convert incoming UUID-like strings to RAW(16) representations.
* stdpool gains an additional TestControls option, to make the pool setup more
  tolerant of cases where the target database is not yet ready to accept
  connections.

This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/419)


BramGruneir (approved):
:lgtm:

I did my best to read through all of it.  
But please get the Oracle queries looked at by @glennfawcett 

> Reviewed 39 of 39 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/419#-:-Nc2oRma1u3mJjTIr-I6:b-jtk5ws)* status: all files reviewed, 1 unresolved discussion (waiting on @bobvawter, @glennfawcett, and @sravotto)

___
*[`internal/target/apply/apply_test.go` line 595 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/419#-Nc2jgGv-s4CERU27Z-0:-Nc2jgGw0hL19fmIuuQz:b-92jkx7) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/7d54a1a213df2727bc8fd4a108e596874ef124dd/internal/target/apply/apply_test.go#L595)):*
> ```Go
> 		switch fixture.TargetPool.Product {
> 		case types.ProductCockroachDB, types.ProductPostgreSQL:
> 			q = "SELECT ver, ts FROM %s WHERE pk = $1"
> ```

If you create a test helper that converts a number to a placeholder, you could join all of these.  It's not required, but would work.
___
*[`internal/target/apply/queries/ora/common.tmpl` line 17 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/419#-Nbu-2iB82sLOXJ_3RNJ:-Nbu-2iB82sLOXJ_3RNK:b-qpcuh7) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/7d54a1a213df2727bc8fd4a108e596874ef124dd/internal/target/apply/queries/ora/common.tmpl#L17)):*
> ```
>     {{- $pairs := . -}}
>     {{- range $pairIdx, $pair := $pairs -}}
>         {{- if $pairIdx }}, {{ end -}}
> ```

space here, but no space in "names" or "join" was that intentional?

bobvawter (commented):
Rebased onto master to resolve conflicts.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/419#-:-Nc4RhfZ5xr4HZZjMVDO:b3y3tih)* status: 32 of 39 files reviewed, 1 unresolved discussion (waiting on @BramGruneir, @glennfawcett, and @sravotto)

___
*[`internal/target/apply/apply_test.go` line 595 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/419#-Nc2jgGv-s4CERU27Z-0:-Nc4JImY-CZzAMOpmCgy:b-c81jwo) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/7d54a1a213df2727bc8fd4a108e596874ef124dd/internal/target/apply/apply_test.go#L595)):*

If you create a test helper that converts a number to a placeholder, you could join all of these.  It's not required, but would work.
</blockquote></details>

I had that thought, too, but then you get into MySQL that can't use numbered parameters at all.  It only has `?` as the substitution marker, which means that things break down if you wanted to use e.g. `$1` twice in the same query.  Also, the test queries are all simple (unless they're not).
___
*[`internal/target/apply/queries/ora/common.tmpl` line 17 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/419#-Nbu-2iB82sLOXJ_3RNJ:-Nc4KPgV8SfcayPY9J6i:b-unq1uo) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/7d54a1a213df2727bc8fd4a108e596874ef124dd/internal/target/apply/queries/ora/common.tmpl#L17)):*

space here, but no space in "names" or "join" was that intentional?
</blockquote></details>

Not intentional.  Removed the space.

PR #428, opened 2023-08-16
ident: Schema.Contains fix for nested namespaces

bobvawter:
This change updates the behavior of Schema.Contains() so that a schema "db" will return true for a table "db.public.tbl".

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/428)

### DIFF
--- a/internal/util/ident/schema_test.go
+++ b/internal/util/ident/schema_test.go
@@ -93,6 +93,12 @@ func TestSchema(t *testing.T) {
 		a.NotEqual(s, s2)
 		a.Equal(s.Canonical(), s2.Canonical())
 		a.Same(s.array, s2.array.lowered)
+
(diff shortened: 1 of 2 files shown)


internal/util/ident/schema_test.go
```diff
@@ -93,6 +93,9 @@ func TestSchema(t *testing.T) {
 		a.NotEqual(s, s2)
 		a.Equal(s.Canonical(), s2.Canonical())
 		a.Same(s.array, s2.array.lowered)
+
+		// Schema "foo" should contain a table defined in a sub-schema.
+		a.True(s.Contains(NewTable(MustSchema(New("foo"), Public), New("bar"))))
```
otan, line 98: can you add a falsy test where `db_name` !contains `db_name.non_public_schema.table_name`?

otan (approved):

PR #429, opened 2023-08-16
source: Clean up references to staging and target db

### LINKED ISSUE #421: targetDB flag is mis-named
This flag should be called targetSchema.  Keep the existing flag, but mark it as hidden, with a warning if it's set.

bobvawter:
This change updates several field and CLI flag names to reflect the use of a schema, as opposed to a database, as the container for tables in cdc-sink. When performing logical replication into CockroachDB, a `--targetSchema foo` flag will be expanded into `foo.public`.

Closes #421


otan (approved):
if you don't mind leaving some breadcrumbs, can you mention `stagingdb/targetdb` flags are deprecated in the commit message?

PR #436, opened 2023-08-21
diag: Add diagnostics collector

bobvawter:
This change adds a global diagnostics collector to improve the supportability of cdc-sink. The diag package allows various components to register themselves in order to produce an arbitrary blob of JSON that we can use to support customers. The applycfg package had an initial implementation of exporting the current configuration via an HTTP endpoint, this change generalizes and expands on it.

Review notes:
* The diag package defines the collector and a basic interface for returning a json value.
* Components are registered with the diag package by their provider functions.
* Diagnostics are available at /_/diag, which is protected by the active authenticator when in C2X modes. The data returned here is not intended to represent a stable API.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/436)

### DIFF
--- a/internal/cmd/mylogical/mylogical.go
+++ b/internal/cmd/mylogical/mylogical.go
@@ -23,6 +23,8 @@ import (
 	"net/http"
 
 	"github.com/cockroachdb/cdc-sink/internal/source/mylogical"
+	"github.com/cockroachdb/cdc-sink/internal/staging/auth/trust"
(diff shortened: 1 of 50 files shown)


bobvawter:
I seem to have found a lifecycle bug in how we're starting logical loops. Going to close this PR and rebase it on a bugfix PR.

PR #437, opened 2023-08-22
apply: fix error message for missing target tables

otan:
Previously, `target` gets shadowed, meaning the error message prints a blank and unhelpful `target` database as it is missing. Fix this by removing the shadowing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/437)


otan:
test failures don't look like me - looks like the enterprise license expired?

otan:
#438 instead, using the main repo

PR #439, opened 2023-08-22
logical: Separate global vs. per-loop concerns

bobvawter:
The existing state of the logical loop package doesn't quite align with its
current usage. The code was originally built for use cases where there was
exactly one logical stream at any given time (e.g. XYlogical tools). The code
was updated somewhat to support multiple resolver loops when using the CDC
frontend, but there are a number of configuration and implementation details
that don't do a good job of separating global versus per-loop configuration and
state.

This change overhauls the configuration and lifecycle management of logical
loops to separate global (e.g. --flag) configuration from per-loop (e.g.
/target/schema) configuration.

* The `logical.BaseConfig` type has per-loop-instance settings extracted into a
  separate `logical.LoopConfig` type. This includes the name of the loop and
  the instance of the Dialect that the loop will drive.
* `logical.Factory` no longer provides any memoization or lifecycle management
  for the loops that it constructs.  This behavior either already exists in
  `cdc.Resolvers` or is irrelevant for single-loop (e.g. pglogical).
* The clean separation of concerns improves component instance reuse in
  multi-loop scenarios. Rather than calling `logical.Start()`, which would
  create duplicate instances of most of the component stack, the
  `logical.Factory.Start()` method constructs only components that depend on
  values in `LoopConfig`.
* Single-loop configurations (e.g. `pglogical.Config`) now also embed
  `logical.LoopConfig` in addition to the previous `logical.BaseConfig`. This
  improves CLI flag binding, since we don't have to hide irrelevant flags in
  multi-loop tools.
* The `logical.Option` type is removed, since it was only ever used as a hack
  to be able to set a custom loop name in multi-loop use cases.

A future change to the configuration may make it possible to override some
BaseConfig fields on a per-loop basis. This would be ideal if cdc-sink is
deployed as a centrally-configured service.

This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/439)


BramGruneir (approved):
:lgtm:

Just some small questions.

> Reviewed 37 of 37 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/439#-:-NdAzSe2DNaAh10Q1LG5:b-2kvnh9)* status: all files reviewed, 3 unresolved discussions (waiting on @bobvawter and @sravotto)

___
*[`internal/source/fslogical/provider.go` line 196 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/439#-NcWUzxC5mRN5ygK5nJV:-NcWUzxDCYEpRLAuM2gl:b-nw7tew) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/4f89c65c44b123b82c8f3e136f46b66c54d838f1/internal/source/fslogical/provider.go#L196)):*
> ```Go
> 	loopConfig.Dialect = ret
> 	loopConfig.LoopName = cfg.TombstoneCollection
> 	_, cancel, err := loops.Start(loopConfig)
> ```

If this returns an error, probably best to not do a full return.
___
*[`internal/source/logical/factory.go` line 52 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/439#-NdAyoZuD1FKOJssPtcw:-NdAyoZuD1FKOJssPtcx:b59hqo2) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/4f89c65c44b123b82c8f3e136f46b66c54d838f1/internal/source/logical/factory.go#L52)):*
> ```Go
> 	}
> 
> 	// Construct the new loop and start it.
> ```

Is there no longer a collection of named loops?
___
*[`internal/source/logical/factory.go` line 87 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/439#-NdAz8LgB50sKMtGwQX4:-NdAz8LgB50sKMtGwQX5:b-tli4xg) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/4f89c65c44b123b82c8f3e136f46b66c54d838f1/internal/source/logical/factory.go#L87)):*
> ```Go
> 
> // newLoop constructs a loop, but does not start it.
> func (f *Factory) newLoop(ctx *stopper.Context, config *LoopConfig) (*Loop, error) {
> ```

Ok, so you've got newLoop and Start.  Is newLoop for internal use one?  Is it called from anywhere besides Start?

bobvawter (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/439)* status: all files reviewed, 3 unresolved discussions (waiting on @BramGruneir and @sravotto)

___
*[`internal/source/fslogical/provider.go` line 196 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/439#-NcWUzxC5mRN5ygK5nJV:-NdGwubQ9q2-kmWD-jLu:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/4f89c65c44b123b82c8f3e136f46b66c54d838f1/internal/source/fslogical/provider.go#L196)):*

If this returns an error, probably best to not do a full return.
</blockquote></details>

Done.
___
*[`internal/source/logical/factory.go` line 52 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/439#-NdAyoZuD1FKOJssPtcw:-NdGxL3B7hx2FL3dWTef:b3vc7yo) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/4f89c65c44b123b82c8f3e136f46b66c54d838f1/internal/source/logical/factory.go#L52)):*

Is there no longer a collection of named loops?
</blockquote></details>

That's correct; it's called out in the commit message that the memoization in `logical.Factory` had become either unnecessary or redundant.
___
*[`internal/source/logical/factory.go` line 87 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/439#-NdAz8LgB50sKMtGwQX4:-NdGx7B57vspwC07pJNw:bk95xcs) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/4f89c65c44b123b82c8f3e136f46b66c54d838f1/internal/source/logical/factory.go#L87)):*

Ok, so you've got newLoop and Start.  Is newLoop for internal use one?  Is it called from anywhere besides Start?
</blockquote></details>

`newLoop()` is called from `Start()` as well as `doBackfill()`.

PR #448, opened 2023-08-31
target: Fix edge cases in UDT enum arrays

bobvawter:
Arrays of UDT enums were under-tested. Adding tests revealed a number of
deficiencies in how enum types were being represented in the schemawatch
package.

Until we decide to implement a proper typesystem model, this change adds an
array field to the ident.UDT type. It also changes ColData.Type to a string,
since we really only ever need to send the column type back to the target
database after introspecting it. This has a minor knock-on effect in the
mylogical package, although that code only ever looks at the ColData.Type field
for a sentinel value.

A small tweak has been made to the pg template to overcome a limitation in the
pgx driver, where the type information for the array-of-enum isn't always
loaded by the time we want to send the enum values. The substitution parameter
is first cast to `TEXT[]` and thence to the enum array type.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/448)

### DIFF
--- a/internal/source/mylogical/conn.go
+++ b/internal/source/mylogical/conn.go
@@ -318,7 +318,7 @@ func (c *conn) onDataTuple(
 				enc[targetCol.Name.Raw()] = string(s)
 			case int64:
 				// if it's a bit need to convert to a string representation
-				if targetCol.Type == mysql.MYSQL_TYPE_BIT {
+				if targetCol.Type == fmt.Sprintf("%d", mysql.MYSQL_TYPE_BIT) {
 					enc[targetCol.Name.Raw()] = strconv.FormatInt(s, 2)
 				} else {
 					enc[targetCol.Name.Raw()] = s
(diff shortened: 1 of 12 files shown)


BramGruneir (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/448#-:-NdCeUTM84BBPNPAvwKk:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sravotto)

___
*[`internal/target/schemawatch/coldata.go` line 197 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/448#-NdCeLPU1p4apaGB32wH:-NdCeLPU1p4apaGB32wI:b16j2hd) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/9999bbfa8e9b9be06ee29826079165aa76b2cfbf/internal/target/schemawatch/coldata.go#L197)):*
> ```Go
> 				parts = append(parts, part)
> 			}
> 			switch len(parts) {
> ```

This starting to get ugly... type system soon....  :(

PR #449, opened 2023-08-31
logical: Fix stall if apply fails in nested backfill mode

bobvawter:
The fslogical frontend will experience a stall if a nested collection is being processed in backfill mode and a mutation cannot be applied. This is because the Flush() method waits on a counter condition variable that would never be decremented.

This change ensures that the decrement always happens by hoisting the guts of the loop() method into a closure and calling the decrement API via defer.

Testing for this condition is performed by plumbing the choas probability into fanWorkers to simulate the applier failing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/449)

### DIFF
--- a/internal/source/logical/logical_test.go
+++ b/internal/source/logical/logical_test.go
@@ -76,13 +76,14 @@ func testLogicalSmoke(t *testing.T, allowBackfill, immediate, withChaos bool) {
 	const numEmits = 100
 	gen.emit(numEmits)
 
-	var dialect logical.Dialect = gen
+	var chaosProb float32
 	if withChaos {
-		dialect = logical.WithChaos(gen, 0.05)
+		chaosProb = 0.1
 	}
 
 	cfg := &logical.BaseConfig{
 		ApplyTimeout:   time.Second, // Increase to make using the debugger easier.
+		ChaosProb:      chaosProb,
 		Immediate:      immediate,
 		RetryDelay:     time.Nanosecond,
 		StagingConn:    fixture.StagingPool.ConnectionString,
@@ -103,7 +104,7 @@ func testLogicalSmoke(t *testing.T, allowBackfill, immediate, withChaos bool) {
 	defer cancelFactory()
 
 	loop, cancelLoop, err := factory.Start(&logical.LoopConfig{
-		Dialect:      dialect,
+		Dialect:      gen,
 		LoopName:     "generator",
 		TargetSchema: dbName,
 	})
--- a/internal/source/logical/fan_workers.go
+++ b/internal/source/logical/fan_workers.go
@@ -18,6 +18,7 @@ package logical
 
 import (
 	"context"
+	"math/rand"
 	"sync"
 
 	"github.com/bobvawter/latch"
@@ -34,6 +35,7 @@ import (
 type fanWorkers struct {
 	appliers     types.Appliers
 	batchSize    int
+	chaosProb    float32 // Set in tests via BaseConfig.ChaosProb.
 	pending      *latch.Counter
 	targetPool   *types.TargetPool
 	stamp        stamp.Stamp
@@ -53,6 +55,7 @@ func newFanWorkers(ctx context.Context, loop *loop, stamp stamp.Stamp) *fanWorke
 	ret := &fanWorkers{
 		appliers:   loop.factory.appliers,
 		batchSize:  batches.Size(),
+		chaosProb:  loop.factory.baseConfig.ChaosProb,
 		pending:    latch.New(),
 		targetPool: loop.factory.targetPool,
 		stamp:      stamp,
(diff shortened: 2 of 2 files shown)


internal/source/logical/logical_test.go
```diff
@@ -77,13 +77,14 @@ func testLogicalSmoke(t *testing.T, allowBackfill, immediate, withChaos bool) {
 	const numEmits = 100
 	gen.emit(numEmits)
 
-	var dialect logical.Dialect = gen
+	var chaosProb float32
 	if withChaos {
-		dialect = logical.WithChaos(gen, 0.05)
```
bobvawter, line 82: The logical factory calls the `WithChaos()` wrapper, so we're not losing any functionality here.

BramGruneir (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/449)* status: all files reviewed, 3 unresolved discussions (waiting on @bobvawter and @sravotto)

___
*[`internal/source/logical/fan_workers.go` line 138 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/449#-NdGF-GR43PwJTqEjWvu:-NdGF-GR43PwJTqEjWvv:bxcoagw) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/168eb517b6eda98f23016684347757f7c8f53193/internal/source/logical/fan_workers.go#L138)):*
> ```Go
> }
> 
> // chaos returns an error for testing.
> ```

chaos SOMETIMES returns an error for testing... 
Is there any reason to put this into fanWorkers and not just keep all the chaos code in the test code?
___
*[`internal/source/logical/fan_workers.go` line 172 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/449#-NdGENvqBOc-19MlPcKs:-NdGENvqBOc-19MlPcKt:b427ncn) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/168eb517b6eda98f23016684347757f7c8f53193/internal/source/logical/fan_workers.go#L172)):*
> ```Go
> 	}
> 
> 	for {
> ```

So this blocks on tryLoop, then tryLoop blocks on waitForWork.

Doesn't this make a lot more sense using channels and goroutines?  
(I can't believe I wrote that last line.)

bobvawter (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/449)* status: all files reviewed, 3 unresolved discussions (waiting on @BramGruneir and @sravotto)

___
*[`internal/source/logical/fan_workers.go` line 138 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/449#-NdGF-GR43PwJTqEjWvu:-NdGHR_A0hN8s0Hi_4jl:b-t4w9zi) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/168eb517b6eda98f23016684347757f7c8f53193/internal/source/logical/fan_workers.go#L138)):*

chaos SOMETIMES returns an error for testing... 
Is there any reason to put this into fanWorkers and not just keep all the chaos code in the test code?
</blockquote></details>

The chaos code isn't in the test code, since we need higher-level packages to be able to configure the chaos behavior for their own testing. Another option would be to pass around a pointer to a struct with a `Chaos() error` method.
___
*[`internal/source/logical/fan_workers.go` line 172 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/449#-NdGENvqBOc-19MlPcKs:-NdGI5yq1DIzpyj_owtd:b-9hzmyj) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/168eb517b6eda98f23016684347757f7c8f53193/internal/source/logical/fan_workers.go#L172)):*

So this blocks on tryLoop, then tryLoop blocks on waitForWork.

Doesn't this make a lot more sense using channels and goroutines?  
(I can't believe I wrote that last line.)
</blockquote></details>

That's how waitForWork operates.  There's a close-and-swap pattern that wakes up the worker routines.  The only reason for this hoist is that `defer` is the right way to express the post-attempt cleanup, but you can't meaningfully `defer` within a `for` loop.

BramGruneir (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/449#-:-NdGMbVP2AH313N4eLdc:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sravotto)

___
*[`internal/source/logical/fan_workers.go` line 138 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/449#-NdGF-GR43PwJTqEjWvu:-NdGLYcQ3HEyeXSWmbBa:b-ho4lli) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/168eb517b6eda98f23016684347757f7c8f53193/internal/source/logical/fan_workers.go#L138)):*

The chaos code isn't in the test code, since we need higher-level packages to be able to configure the chaos behavior for their own testing. Another option would be to pass around a pointer to a struct with a `Chaos() error` method.
</blockquote></details>

I really wish go would allow adding test only fields to structs.
 
Can you add a comment that it's for testing purposes only?
___
*[`internal/source/logical/fan_workers.go` line 172 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/449#-NdGENvqBOc-19MlPcKs:-NdGMU1v8ZupVPXdeMpj:b-rxxj8h) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/168eb517b6eda98f23016684347757f7c8f53193/internal/source/logical/fan_workers.go#L172)):*

That's how waitForWork operates.  There's a close-and-swap pattern that wakes up the worker routines.  The only reason for this hoist is that `defer` is the right way to express the post-attempt cleanup, but you can't meaningfully `defer` within a `for` loop.
</blockquote></details>

We should be able to specify at what level the defer occurs, not have to wrap everything in sub functions.

bobvawter (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/449)* status: 0 of 58 files reviewed, all discussions resolved (waiting on @BramGruneir and @sravotto)

___
*[`internal/source/logical/fan_workers.go` line 138 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/449#-NdGF-GR43PwJTqEjWvu:-NdHKeAg-Bzp45tNeIPE:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/168eb517b6eda98f23016684347757f7c8f53193/internal/source/logical/fan_workers.go#L138)):*

I really wish go would allow adding test only fields to structs.
 
Can you add a comment that it's for testing purposes only?
</blockquote></details>

Done.

BramGruneir (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/449)* status: 0 of 58 files reviewed, all discussions resolved (waiting on @sravotto)

PR #450, opened 2023-09-01
apply: Allow sparse payloads to use SQL DEFAULT expressions

bobvawter:
This change allows sparse payloads to interact nicely with columns that have a `DEFAULT` expression. That is, the `DEFAULT` expression for a column will be part of the upsert operation if no value for that column is provided by the incoming payload. This improves migration use-cases, where there may be a non-trivial amount of on-the-fly schema migration.

This behavior can already be accomplished by using an apply configuration or a userscript, but the cdc-sink operator would have to maintain the default value in both the SQL schema and the cdc-sink configuration.

Columns that support sparse operation add an extra parameter to their upsert expression that uses a `CASE` expression to select between the next parameter or the default expression.  It is not possible to use the `DEFAULT` keyword in this context, so we augment the schema introspection code to retrieve the appropriate expression.

The template test has been overhauled to use golden-output files, which can be regenerated by changing the `rewriteFiles` constant.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/450)

### DIFF
--- a/.github/workflows/golang.yaml
+++ b/.github/workflows/golang.yaml
@@ -59,7 +59,7 @@ jobs:
 
       - name: Copyright headers
         if: ${{ always () }}
-        run: go run github.com/google/addlicense -c "The Cockroach Authors" -l apache -s -v  -check .
+        run: go run github.com/google/addlicense -c "The Cockroach Authors" -l apache -s -v  -check  -ignore '**/testdata/**/*.sql' .
 
       - name: Lint
         if: ${{ always() }}
--- a/internal/target/apply/apply.go
+++ b/internal/target/apply/apply.go
@@ -327,7 +327,7 @@ func (a *apply) upsertLocked(
 			// in the input data, but that we don't actually want to
 			// make use of the provided value. For example, generated PK
 			// columns.
-			if targetColumn.UpsertPosition < 0 {
+			if targetColumn.UpsertIndex < 0 {
 				return nil
 			}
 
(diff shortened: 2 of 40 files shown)


BramGruneir (approved):
I wish you had separated the overhaul to golden output and this change.  The results are much easier to deal with.
But I'm not going to go over all of the sql as written, so I'm just going to assume it's still correct.

:lgtm: 

> Reviewed 41 of 41 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/450#-:-NdHJlVI3jGibDpmcr8J:b-92b6bq)* status: all files reviewed, 1 unresolved discussion (waiting on @bobvawter and @sravotto)

___
*[`internal/target/apply/queries/pg/delete.tmpl` line 7 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/450#-NdHL-KUDSOPDSvyZmbc:-NdHL-KUDSOPDSvyZmbd:bs3hq7q) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/2a49de1c719277df020440120c2867deae3ca789/internal/target/apply/queries/pg/delete.tmpl#L7)):*
> ```
> */ -}}
> DELETE FROM {{ .TableName }} WHERE (
>     {{- template "names" .PKDelete -}}
> ```

Why did you remove the tab?

bobvawter (commented):
FWIW, the template variants have full-up tests in `apply_test.go`, so there's redundant coverage.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/450#-:-NdHMs4EExYQb3xvwSJy:b-9kgf8w)* status: all files reviewed, 1 unresolved discussion (waiting on @BramGruneir and @sravotto)

___
*[`internal/target/apply/queries/pg/delete.tmpl` line 7 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/450#-NdHL-KUDSOPDSvyZmbc:-NdHMYgIAcwG6I-lb76i:b-ceg7gf) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/2a49de1c719277df020440120c2867deae3ca789/internal/target/apply/queries/pg/delete.tmpl#L7)):*

Why did you remove the tab?
</blockquote></details>

The template formatter in GoLand likes to de-indent tags that aren't nested within other tags.

PR #451, opened 2023-09-01
cdc: Fix flag panic due to bad merge

bobvawter:
PR #439 had a bad merge hunk which keeps the binary from starting. This change also adds a check to ensure that the main entry point starts.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/451)


BramGruneir:
:lgtm:

PR #458, opened 2023-09-06
fslogical: Fix backfill pagination

bobvawter:
The firestore.DocumentID sentinel usually works correctly with the values that one might expect to find in `doc.Ref.ID` when querying single collections. However, when querying collection groups, that value needs to be more (globally) unique, since the same document ID can, in theory, exist within multiple collection groups that are being queried.

This change removes a misguided attempt at solving for this, with a new understanding of how this sentinel behaves. Reference URLs are included in the docstring on `Dialect.backfillPoint()`.

The previous code was, in effect, randomly sampling swathes of document ids, while the test code only checked that we arrived at the desired number of rows in the target tables. In order to verify that ID-based pagination is working correctly, all documents to be backfilled are inserted at the same timestamp. A new test control is added to LoopConfig, which allows the test code to delay the loop switching from backfill to streaming mode. When deployed into production, the stochastic nature eventually sampled all rows in sufficiently small or sufficiently long-running attempts.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/458)

### DIFF
--- a/internal/source/fslogical/fslogical.go
+++ b/internal/source/fslogical/fslogical.go
@@ -22,6 +22,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
 	"time"
 
 	"cloud.google.com/go/firestore"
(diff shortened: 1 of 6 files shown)


BramGruneir (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/458)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sravotto)

BramGruneir (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/458#-:-Ne3shJx-Ly7Cf4F2PF2:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sravotto)

PR #463, opened 2023-09-08
workflow: Fix spontaneous cancellation of cache warming workflow

bobvawter:
We've been seeing unexpected cancellation of the cache-warming workflow. This
change moves the concurrency block from the job up to the workflow itself and
explicitly sets cancel-in-progress to false to see if that solves the problem.

This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/463)


BramGruneir:
Lgtm

Issue #430, opened 2023-08-17
Using PostgreSQL as a Changefeed Target in cdc-sink


farisam:
Hi All,

I have a question regarding the cdc-sink project and its compatibility with PostgreSQL as a changefeed target. I've been exploring the project and its capabilities, and I was wondering if there is support or plans to support using PostgreSQL as a target for changefeeds.

Could someone kindly provide some insight into whether this feature is currently available or if it's planned for future releases? I greatly appreciate any information or guidance you can offer on this matter.

Thank you for your time and consideration.

PR #469, opened 2023-09-11
schemawatch: Support vanilla PostgreSQL

bobvawter:
This change ports the schema introspection queries from using
CockroachDB-specific porcelain to using the information_schema virtual tables.
The existing table-dependency query is still used for older versions of
CockroachDB that enter an infinite loop when executing the revised CTE.

The test fixture code adds a workaround for the lack of cross-database
information_schema queries by swapping out the underlying database connection
after we've performed a `CREATE DATABASE`.

The schemawatch tests will pass against PostgreSQL v11-v15 by setting
`TEST_TARGET_CONNECT=postgres://postgres:SoupOrSecret@127.0.0.1:5432`

CI testing using PostgreSQL as a target will not be enabled until the apply
package is updated.

X-Ref: #430

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/469)

### DIFF
--- a/internal/sinktest/all/wire_gen.go
+++ b/internal/sinktest/all/wire_gen.go
@@ -67,7 +67,7 @@ func NewFixture() (*Fixture, func(), error) {
 		cleanup()
 		return nil, nil, err
 	}
-	targetSchema, cleanup8, err := base.ProvideTargetSchema(context, targetPool)
+	targetSchema, cleanup8, err := base.ProvideTargetSchema(context, diagnostics, targetPool)
 	if err != nil {
 		cleanup7()
 		cleanup6()
--- a/internal/sinktest/base/provider.go
+++ b/internal/sinktest/base/provider.go
@@ -251,11 +251,39 @@ func ProvideSourceSchema(
 // ProvideTargetSchema create a globally-unique container for tables in
 // the target database.
 func ProvideTargetSchema(
-	ctx context.Context, pool *types.TargetPool,
+	ctx context.Context, diags *diag.Diagnostics, pool *types.TargetPool,
 ) (sinktest.TargetSchema, func(), error) {
-	sch, cancel, err := provideSchema(ctx, pool, "tgt")
+	sch, dropSchema, err := provideSchema(ctx, pool, "tgt")
+	ret := sinktest.TargetSchema(sch)
+	if err != nil {
+		return ret, nil, err
+	}
 	log.Infof("target schema: %s", sch)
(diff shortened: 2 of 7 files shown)


BramGruneir (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/469#-:-Ne9uTHb3HVmGluAL_iL:bnfp4nl)* status: all files reviewed, 3 unresolved discussions (waiting on @bobvawter)

___
*[`internal/target/schemawatch/coldata.go` line 103 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/469#-Ne98WK04-wj0zFF5gPI:-Ne98WK04-wj0zFF5gPJ:b-rg517) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/099414614de78f53e9ad5f3c53f11953d7368090/internal/target/schemawatch/coldata.go#L103)):*
> ```Go
> // before the non-PK columns.
> const sqlColumnsQueryPg = `
>     WITH pk_name AS (
> ```

So this is no longer just constraint name, perhaps rename it pk_constraints
___
*[`internal/target/schemawatch/coldata.go` line 228 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/469#-Ne9osJm3quk8WH3GZsu:-Ne9osJm3quk8WH3GZsv:b-tk53ja) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/099414614de78f53e9ad5f3c53f11953d7368090/internal/target/schemawatch/coldata.go#L228)):*
> ```Go
> 				parts := parsed.Idents(make([]ident.Ident, 0, 3))
> 				if len(parts) != 3 {
> 					return errors.Errorf("unexpected number of parts when splitting ident: %s", parsed)
> ```

Add the expected 3, but got x in case something does go wrong.
___
*[`internal/target/schemawatch/coldata.go` line 241 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/469#-Ne9pkOy9LVCdGNZyiWe:-Ne9pkOy9LVCdGNZyiWf:becn3oq) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/099414614de78f53e9ad5f3c53f11953d7368090/internal/target/schemawatch/coldata.go#L241)):*
> ```Go
> 					lastPart = strings.ToUpper(lastPart)
> 					if isArray {
> 						lastPart = lastPart[1:] + "[]"
> ```

Why are you dropping the first character?

PR #470, opened 2023-09-12
workflow: Rearrange the deck chairs

bobvawter:
Using a concurrency group in a called workflow (i.e. to only warm the go build
cache once) is still showing spontaneous cancellation by the actions platform.
This change rolls all workflows up into a single dispatch event, triggered by a
top-level workflow. The top-level go.yaml then calls the existing, albeit
renamed, workflows after the cache warming has taken place.

An upside to this approach is that there's a relatively simple workflow that
provides a high-level view of the build, while the details of the build can
still be grouped into separate files.

Conditional behaviors, such as pushing to DockerHub or to the GCP bucket, have
been made inputs of their respective workflows, again to make it obvious at the
outset what happens when.

Where applicable, the `${{ !cancelled()}}` idiom is used in place of `${{
always() }}` for steps that don't provide diagnostic information that we may
want to capture out of a cancelled workflow.

This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/470)

### DIFF
--- a/.github/workflows/go-binaries.yaml
+++ b/.github/workflows/go-binaries.yaml
@@ -14,34 +14,31 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-# Binaries will build cdc-sink for a variety of common platforms.
-name: Binaries
+# Binaries will build cdc-sink for a variety of common platforms and
+# optionally push the results to a GCP bucket.
+name: Golang Binaries
 permissions:
   contents: read
-concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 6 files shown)


BramGruneir (approved):
:lgtm:

Let's see if this incantation works!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/470#-:-NeA8VGOAYhClgYKxX9p:bicgefd)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @bobvawter)

PR #474, opened 2023-09-13
workflow: Use larger machine types where appropriate

bobvawter:
This change uses a larger machine type when regenerating the golang build cache and when running the integration tests against Oracle DB.

The cache-warming workflow is split into two jobs. The first checks for a cache hit. If there is a cache miss, the second job will be scheduled on a large runners, since go compilation is highly parallel.


BramGruneir (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/474#-:-NeEzHQH0JhYxL1mkAd9:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @bobvawter and @sravotto)

___
*[`.github/workflows/go-build-cache.yaml` line 65 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/474#-NeEz7nt3eTjT5hVhppF:-NeEz7nt3eTjT5hVhppG:b-5h7eda) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/6893f2b4a4965ed828a013a1d146ec1cee2452a3/.github/workflows/go-build-cache.yaml#L65)):*
> ```YAML
>         # Keep options in sync with golang.yaml
>         run: |
>           go build . ./...
> ```

What is the extra `.` for?

PR #475, opened 2023-09-13
apply: Add support for PostgreSQL

bobvawter:
This change adds support to the apply package for applying mutations to a
PostgreSQL database. It also enables CI testing for PostgreSQL v11 through v15.

Review notes:
- The primary change required is to use INSERT ON CONFLICT DO UPDATE instead of
  the CockroachDB-specific UPSERT command. One note here is that CockroachDB
  does not allow the ROW keyword when performing a multi-column update, whereas
  PostgreSQL requires it.
- The golden-output template tests now compare CRDB vs. PG.
- ProvideTargetSchema() updates the TargetPool's connection string. This value
  is used by logical-loop tests to create their own connection pool.
- The idle_in_transaction_session_timeout connection variable must be set to an
  integer number of milliseconds in earlier version of PostgreSQL.

X-Ref: https://github.com/cockroachdb/cockroach/issues/79081

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/475)

### DIFF
--- a/.github/workflows/go-tests.yaml
+++ b/.github/workflows/go-tests.yaml
@@ -122,6 +122,23 @@ jobs:
             runs-on: ubuntu-latest-8-core
             target: oracle-v21.3
             targetConn: "oracle://system:SoupOrSecret@127.0.0.1:1521/XEPDB1"
+
+          # Test CRDB -> PostgreSQL for migration backfill use cases.
+          - cockroachdb: v23.1
+            target: postgresql-v11
+            targetConn: "postgres://postgres:SoupOrSecret@127.0.0.1:5432"
+          - cockroachdb: v23.1
+            target: postgresql-v12
+            targetConn: "postgres://postgres:SoupOrSecret@127.0.0.1:5432"
+          - cockroachdb: v23.1
+            target: postgresql-v13
+            targetConn: "postgres://postgres:SoupOrSecret@127.0.0.1:5432"
+          - cockroachdb: v23.1
+            target: postgresql-v14
+            targetConn: "postgres://postgres:SoupOrSecret@127.0.0.1:5432"
+          - cockroachdb: v23.1
+            target: postgresql-v15
+            targetConn: "postgres://postgres:SoupOrSecret@127.0.0.1:5432"
     env:
       COVER_OUT: coverage-${{ strategy.job-index }}.out
       DOCKER_LOGS_OUT: docker-${{ strategy.job-index }}.log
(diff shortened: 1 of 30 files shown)


BramGruneir:
I think it might be worthwhile keeping separate templates for each destination.  It means the templates have less extra logic in them and hence are more readable.  Also, weird bugs between versions (of say postgres) won't accidently creep into crdb.

Also, that way crdb can keep the upset, which is a much nicer query.

BramGruneir (changes requested):
> Reviewed 30 of 30 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/475)* status: all files reviewed, 1 unresolved discussion (waiting on @bobvawter and @sravotto)

___
*[`internal/target/apply/apply_test.go` line 590 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/475#-NeOGo89D4LSnMirBeQB:-NeOGo89D4LSnMirBeQC:b-spewjp) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/4616a5494e59db2e072e42af332307fd91bee313/internal/target/apply/apply_test.go#L590)):*
> ```Go
> }
> 
> func normalizeJSON(data string) (string, error) {
> ```

Can you just add a quick comment about why this is needed?

bobvawter (commented):
There are now `crdb` and `pg` template directories.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/475#-:-NePqqWeCFgTb1YpFZ9z:bqhynv5)* status: 22 of 33 files reviewed, 1 unresolved discussion (waiting on @BramGruneir and @sravotto)

___
*[`internal/target/apply/apply_test.go` line 590 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/475#-NeOGo89D4LSnMirBeQB:-NePqxzg1FJgwBYaxHXC:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/4616a5494e59db2e072e42af332307fd91bee313/internal/target/apply/apply_test.go#L590)):*

Can you just add a quick comment about why this is needed?
</blockquote></details>

Done.

BramGruneir (approved):
:lgtm:

> Reviewed 11 of 11 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/475#-:-NeUaKxg4xIf14dAQB01:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sravotto)

PR #477, opened 2023-09-15
logical: Overhaul API

### LINKED ISSUE #459: pglogical.lsnStamp is not well-ordered
The order in which we receive `BeginMessage` doesn't necessarily line up with their starting LSNs.  We need to investigate if the XID (with wraparound) is a better way to make `lsnStamp` well-ordered. There are additional LSN values associated with `CommitMessage` and should be researched. It would also be possible to add an extra method to `logical.Events` to allow the transaction stamp to be swizzled, but this may require replacing the use of stamps with handles to stamps in the logical package. See discussion in PR #458.

bobvawter:
The logical package API was built with a very specific use-case in mind: a serial source of replication events where the source controls replay. This design has been stretched to where it no longer makes sense for the variety of existing use cases (e.g. CDC) and is not suitable for others (e.g. Immedate-CDC mode). The hidden aspect of the logical loop managing the consistent point also makes flow control and behavior difficult to reason about.

This change makes two major changes to the API:
* The Events interface now constructs a Batch which receives data.
* The Dialect is now responsible for advancing the Loop's consistent point.

These changes will make it possible to build an Immediate-CDC pipeline that uses the same plumbing as the rest of the replication flows.

There is a common pattern of needing to know when some variable has changed that was addressed in several different ways. These concerns have been addressed with a new helper type that provides a notification mechanism when the enclosed value has changed.

As a side-effect of making the Dialect responsible for advancing the consistent point, the pglogical implementation can now use the LSN of the COMMIT record as its consistent point. This makes the LSN numbers well-ordered and we can implement an exactly-once behavior by ignoring incoming replication messages until we catch up to a previously-applied COMMIT.

Fixes #459

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/477)


BramGruneir (approved):
:lgtm:

This is so much cleaner.  
Might be worth moving notify to it's own repo.  I can see it being useful in a lot of projects.

> Reviewed 23 of 23 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/477#-:-Ned9CB4489A69UziEIh:b-tjwwo5)* status: all files reviewed, 2 unresolved discussions (waiting on @bobvawter and @sravotto)

___
*[`internal/source/cdc/resolver.go` line 187 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/477#-NecgrRN758JqYusJpJf:-NecgrRN758JqYusJpJg:b-vx24u5) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/990f32c07073fe0c7ed940f4cc3a8ad0e6e8cd2b/internal/source/cdc/resolver.go#L187)):*
> ```Go
> ) error {
> 	// Backup polling interval for work.
> 	const dbPollInterval = 1 * time.Second
> ```

Why declare this inside the function?
___
*[`internal/source/logical/fan_batch.go` line 226 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/477#-Ned6r6931yOLs3kfwXN:-Ned6r6931yOLs3kfwXO:b-e3hsgu) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/990f32c07073fe0c7ed940f4cc3a8ad0e6e8cd2b/internal/source/logical/fan_batch.go#L226)):*
> ```Go
> 	}
> 	close(b.mu.updated)
> 	b.mu.updated = make(chan struct{})
> ```

Any way to use notify itself for this instead of reimplementing it?  I know it's not a perfect fit.  So it it's too ugly, don't.

bobvawter (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/477)* status: 19 of 23 files reviewed, 2 unresolved discussions (waiting on @BramGruneir and @sravotto)

___
*[`internal/source/cdc/resolver.go` line 187 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/477#-NecgrRN758JqYusJpJf:-NedoFyW4M-RzMRzKgzG:bl0jdf4) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/990f32c07073fe0c7ed940f4cc3a8ad0e6e8cd2b/internal/source/cdc/resolver.go#L187)):*

Why declare this inside the function?
</blockquote></details>

Moved to the top-level of the file.
___
*[`internal/source/logical/fan_batch.go` line 226 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/477#-Ned6r6931yOLs3kfwXN:-NedoJDiC7IbqFF9Bxkp:b-jase6b) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/990f32c07073fe0c7ed940f4cc3a8ad0e6e8cd2b/internal/source/logical/fan_batch.go#L226)):*

Any way to use notify itself for this instead of reimplementing it?  I know it's not a perfect fit.  So it it's too ugly, don't.
</blockquote></details>

This did improve the code, since the entire `mu` can be moved into the `notify.Var`.  I added a `Var.Peek()` method to hold a read-lock while using the value to provide synchronization for the `data` map.

BramGruneir (approved):
> Reviewed 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/477)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sravotto)

PR #494, opened 2023-09-21
apply: Use prepared statements in the hot path

bobvawter:
This change adds prepared statements to the apply package to support targets
that don't use the pgx library. pgx automatically retains prepared statements
without user intervention, so this was never necessary before.

This measurably improves throughput when using Oracle Database as a target and
would be useful for other targets in the future. This change also identified a
dependency issue in some of the test rigs, where the Appliers and Watchers were
attached to the bootstrap stack instead of the per-test stack. This led to
cross sql.DB use between prepared statements.

Since we have a cache around prepared statement, the apply template code no
longer has its own cache of SQL expressions.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/494)

### DIFF
--- a/internal/sinktest/all/wire_gen.go
+++ b/internal/sinktest/all/wire_gen.go
@@ -67,8 +67,10 @@ func NewFixture() (*Fixture, func(), error) {
 		cleanup()
 		return nil, nil, err
 	}
-	targetSchema, cleanup8, err := base.ProvideTargetSchema(context, diagnostics, targetPool)
+	targetStatements, cleanup8 := base.ProvideTargetStatements(targetPool)
+	targetSchema, cleanup9, err := base.ProvideTargetSchema(context, diagnostics, targetPool, targetStatements)
 	if err != nil {
+		cleanup8()
 		cleanup7()
 		cleanup6()
 		cleanup5()
@@ -84,22 +86,11 @@ func NewFixture() (*Fixture, func(), error) {
 		SourceSchema: sourceSchema,
 		StagingPool:  stagingPool,
 		StagingDB:    stagingSchema,
+		TargetCache:  targetStatements,
 		TargetPool:   targetPool,
 		TargetSchema: targetSchema,
 	}
-	configs, cleanup9, err := applycfg.ProvideConfigs(context, diagnostics, stagingPool, stagingSchema)
-	if err != nil {
-		cleanup8()
-		cleanup7()
-		cleanup6()
-		cleanup5()
-		cleanup4()
-		cleanup3()
-		cleanup2()
-		cleanup()
-		return nil, nil, err
-	}
-	watchers, cleanup10, err := schemawatch.ProvideFactory(targetPool, diagnostics)
+	configs, cleanup10, err := applycfg.ProvideConfigs(context, diagnostics, stagingPool, stagingSchema)
 	if err != nil {
 		cleanup9()
 		cleanup8()
@@ -112,8 +103,23 @@ func NewFixture() (*Fixture, func(), error) {
 		cleanup()
 		return nil, nil, err
 	}
-	appliers, cleanup11, err := apply.ProvideFactory(configs, diagnostics, targetPool, watchers)
+	watchers, cleanup11, err := schemawatch.ProvideFactory(targetPool, diagnostics)
 	if err != nil {
+		cleanup10()
+		cleanup9()
+		cleanup8()
+		cleanup7()
+		cleanup6()
+		cleanup5()
+		cleanup4()
+		cleanup3()
+		cleanup2()
+		cleanup()
+		return nil, nil, err
+	}
+	appliers, cleanup12, err := apply.ProvideFactory(targetStatements, configs, diagnostics, targetPool, watchers)
+	if err != nil {
+		cleanup11()
 		cleanup10()
 		cleanup9()
 		cleanup8()
@@ -128,6 +134,7 @@ func NewFixture() (*Fixture, func(), error) {
 	}
 	memoMemo, err := memo.ProvideMemo(context, stagingPool, stagingSchema)
 	if err != nil {
+		cleanup12()
 		cleanup11()
 		cleanup10()
 		cleanup9()
@@ -145,6 +152,7 @@ func NewFixture() (*Fixture, func(), error) {
 	checker := version.ProvideChecker(stagingPool, memoMemo)
 	watcher, err := ProvideWatcher(context, targetSchema, watchers)
 	if err != nil {
+		cleanup12()
 		cleanup11()
 		cleanup10()
 		cleanup9()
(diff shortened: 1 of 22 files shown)


BramGruneir (changes requested):
> Reviewed 1 of 1 files at r1, 21 of 21 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/494)* status: all files reviewed, 4 unresolved discussions (waiting on @bobvawter and @sravotto)

___
*[`internal/sinktest/base/provider.go` line 292 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/494#-Nex9bf30rL76s22-VkU:-Nex9bf30rL76s22-VkV:b-f8fyqj) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/256848d7ad56e88c8a3509cce0f3d159bdf063b2/internal/sinktest/base/provider.go#L292)):*
> ```Go
> 			cancelNext()
> 		}
> 		stmts.Cache = stmtcache.New[string](pool.DB, 128)
> ```

Please move this 128 to a const above.
___
*[`internal/sinktest/base/provider.go` line 301 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/494#-Nex9gR68DZDfIZvSAJ6:-Nex9gR68DZDfIZvSAJ7:b-4s2760) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/256848d7ad56e88c8a3509cce0f3d159bdf063b2/internal/sinktest/base/provider.go#L301)):*
> ```Go
> // reused should also reuse the cache.
> func ProvideTargetStatements(pool *types.TargetPool) (*types.TargetStatements, func()) {
> 	ret := stmtcache.New[string](pool.DB, 128)
> ```

This 128 as well.
___
*[`internal/source/logical/config.go` line 34 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/494#-Nex9mGf9PM0CDs1dyXE:-Nex9mGf9PM0CDs1dyXF:bg1uajl) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/256848d7ad56e88c8a3509cce0f3d159bdf063b2/internal/source/logical/config.go#L34)):*
> ```Go
> 	defaultRetryDelay      = 10 * time.Second
> 	defaultStandbyTimeout  = 5 * time.Second
> 	defaultTargetCacheSize = 128
> ```

Why did you pick 128? Why not 1024?  
___
*[`internal/util/stmtcache/metrics.go` line 25 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/494#-NexAnBn8SU7JMVP31Bb:-NexAnBn8SU7JMVP31Bc:b-kcokjr) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/256848d7ad56e88c8a3509cce0f3d159bdf063b2/internal/util/stmtcache/metrics.go#L25)):*
> ```Go
> 
> var (
> 	stmtCacheHits = promauto.NewCounter(prometheus.CounterOpts{
> ```

Love that you added stats right away.  :)
___
*[`internal/util/stmtcache/stmtcache.go` line 89 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/494#-NexB3kBB57fdSy5wZUT:-NexB3kBB57fdSy5wZUU:bdnlq0q) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/256848d7ad56e88c8a3509cce0f3d159bdf063b2/internal/util/stmtcache/stmtcache.go#L89)):*
> ```Go
> 	ctx context.Context, db any, key T, gen func() (string, error),
> ) (*sql.Stmt, error) {
> 	stmt, err := c.get(ctx, key, gen)
> ```

This passing of the gen function in seems like a weird design choice.  I can see why you're doing it, but it seems like breaking of the separation of concerns.  And the fact that it can produce an error... maybe just panic on the error instead?  

I think you might have made this a bit too generic.  But then again, I could see it being useful this way.  So I'm torn.

Or instead of storing it a single level lookup table, would it be more efficient to have it multilevel?  cache[operation][target][generation][mut count]?  I don't know if that would be faster or slower.  Maybe slower... 

Let's chat outside of this review and discuss before you go changing anything.

BramGruneir (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/494#-:-NeySSV92z-4dLFG5Sr5:bnfp4nl)* status: all files reviewed, 3 unresolved discussions (waiting on @bobvawter and @sravotto)

bobvawter (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/494)* status: 19 of 22 files reviewed, 3 unresolved discussions (waiting on @BramGruneir and @sravotto)

___
*[`internal/sinktest/base/provider.go` line 292 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/494#-Nex9bf30rL76s22-VkU:-Neyn2wN9rBTpZMNu4O7:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/256848d7ad56e88c8a3509cce0f3d159bdf063b2/internal/sinktest/base/provider.go#L292)):*

Please move this 128 to a const above.
</blockquote></details>

Done.
___
*[`internal/sinktest/base/provider.go` line 301 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/494#-Nex9gR68DZDfIZvSAJ6:-Neyn3PLCyXjVgW0fjET:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/256848d7ad56e88c8a3509cce0f3d159bdf063b2/internal/sinktest/base/provider.go#L301)):*

This 128 as well.
</blockquote></details>

Done.
___
*[`internal/source/logical/config.go` line 34 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/494#-Nex9mGf9PM0CDs1dyXE:-NeynINyEENiSAC6uvl1:b-tltmvp) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/256848d7ad56e88c8a3509cce0f3d159bdf063b2/internal/source/logical/config.go#L34)):*

Why did you pick 128? Why not 1024?
</blockquote></details>

Added a clarifying comment.
___
*[`internal/util/stmtcache/stmtcache.go` line 89 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/494#-NexB3kBB57fdSy5wZUT:-NeynMIqCdtZJaGJk790:b-29iiww) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/256848d7ad56e88c8a3509cce0f3d159bdf063b2/internal/util/stmtcache/stmtcache.go#L89)):*

This passing of the gen function in seems like a weird design choice.  I can see why you're doing it, but it seems like breaking of the separation of concerns.  And the fact that it can produce an error... maybe just panic on the error instead?  

I think you might have made this a bit too generic.  But then again, I could see it being useful this way.  So I'm torn.

Or instead of storing it a single level lookup table, would it be more efficient to have it multilevel?  cache[operation][target][generation][mut count]?  I don't know if that would be faster or slower.  Maybe slower... 

Let's chat outside of this review and discuss before you go changing anything.
</blockquote></details>

Added some additional commentary to Cache as to why it's not built like a lazy map with a member factory function.

PR #495, opened 2023-09-22
apply: Use bulk inserts for Oracle

bobvawter:
This is stacked on #494 

This change alters the MERGE statement to use a bulk-transfer mechanism in the
go-ora driver. Instead of generating N-many SELECT FROM DUAL rows for each
input element, we pack the arguments into slices, one per column.

This is built in a fairly generic manner, so we should be able to use this for
other targets that can take advantage of bulk transfers through the sql.DB
interface.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/495)


BramGruneir (approved):
:lgtm:

> Reviewed 9 of 9 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/495#-:-NeyTfRN2gvx5e1q6a9m:bnfp4nl)* status: 9 of 29 files reviewed, all discussions resolved (waiting on @sravotto)

PR #502, opened 2023-09-27
schemawatch: fix dependency query for constraints using different schema

sravotto:
In the depOrderTemplatePg the where clause within the refs subquery has a redundant check for table_catalog, 
which should be a check for table_schema matching.

This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/502)

### DIFF
--- a/internal/target/schemawatch/dependencies.go
+++ b/internal/target/schemawatch/dependencies.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
-	"strings"
 
 	"github.com/cockroachdb/cdc-sink/internal/types"
 	"github.com/cockroachdb/cdc-sink/internal/util/ident"
@@ -80,9 +79,10 @@ WHERE tbl_owner = (:owner)
 GROUP BY tbl_name
 ORDER BY lvl, tbl_name`
 
-// depOrderLegacyCRDB supports versions of CRDB <= v21.2 which experience
-// an infinite loop when executing depOrderTemplatePg.
-const depOrderLegacyCRDB = `
+// depOrderTemplateCRDB uses the SHOW TABLES porcelain to calculate the referential depth.
+// Older version of CRDB (<= 21.2) experience an infinite loop when executing
+// depOrderTemplatePg.
+const depOrderTemplateCRDB = `
 WITH RECURSIVE
  tables AS (
    SELECT schema_name AS sch, table_name AS tbl
@@ -175,8 +175,8 @@ WITH RECURSIVE
             AND ref.unique_constraint_schema = parent.table_schema
             AND ref.unique_constraint_name = parent.constraint_name
       WHERE
-        (child.table_catalog, child.table_catalog, child.table_name)
-        != (parent.table_catalog, parent.table_catalog, parent.table_name)
+        (child.table_catalog, child.table_schema, child.table_name)
+        != (parent.table_catalog, parent.table_schema, parent.table_name)
     ),
   roots
     AS (
@@ -237,8 +237,15 @@ func getDependencyOrder(
 			return nil, errors.Errorf("expecting two schema parts, had %d", len(parts))
 		}
 
-		if tx.Product == types.ProductCockroachDB && strings.Contains(tx.Version, "v21.") {
-			stmt = fmt.Sprintf(depOrderLegacyCRDB, db, parts[0])
+		// We are using a different template for CRDB
+		// Older release (<= 21.2) may experience infinite loops using
+		// More recent releases may fail to report a correct depth
+		// when there are cross-schema dependencies.
+		// See https://github.com/cockroachdb/cockroach/issues/111419
+		// Once the issue above is fixed and we are not supporting 21.2
+		// we can use depOrderTemplatePg for CRDB as well.
+		if tx.Product == types.ProductCockroachDB {
+			stmt = fmt.Sprintf(depOrderTemplateCRDB, db, parts[0])
 		} else {
 			stmt = fmt.Sprintf(depOrderTemplatePg, parts[0])
 			args = []any{parts[0].Raw(), parts[1].Raw()}
@@ -275,9 +282,11 @@ func getDependencyOrder(
 				continue
 			}
 
-			if nextOrder > currentOrder {
-				currentOrder = nextOrder
+			// Allow skipping a level. This might happen if there
+			// are references across schemas.
+			for nextOrder > currentOrder {
 				depOrder = append(depOrder, nil)
+				currentOrder++
 			}
 			depOrder[currentOrder] = append(depOrder[currentOrder], tbl)
 		}
--- a/internal/target/schemawatch/dependencies_test.go
+++ b/internal/target/schemawatch/dependencies_test.go
@@ -197,3 +197,59 @@ func TestNoDeferrableConstraints(t *testing.T) {
 	a.ErrorContains(err, "deferrable")
 	a.ErrorContains(err, "42601")
 }
+
+// TestCrossSchemaTableReferencesPG verifies that we can handle
+// cross schema constraints, in case a table with the same name
+// is present in two different schemata.
+func TestCrossSchemaTableReferencesPG(t *testing.T) {
+	a := assert.New(t)
+	r := require.New(t)
+
+	fixture, cancel, err := all.NewFixture()
+	r.NoError(err)
+	defer cancel()
+
+	ctx := fixture.Context
+	pool := fixture.TargetPool
+	db, _ := fixture.TargetSchema.Split()
+	switch pool.Product {
+	case types.ProductPostgreSQL:
+		// In Postgres we can only create schemas in the current database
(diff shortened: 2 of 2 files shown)


bobvawter:
Please write a complete commit message and add a test that fails under the old behavior.

bobvawter:
https://cockroachlabs.atlassian.net/wiki/spaces/CRDB/pages/73072807/Git+Commit+Messages is what we should base our standards on.

sravotto (commented):
Done.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/502#-:-NfW93df1q8ll1NdXrSQ:b-896fix)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @bobvawter)

bobvawter (changes requested):
> Reviewed 1 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/502)* status: 1 of 2 files reviewed, 5 unresolved discussions (waiting on @sravotto)

___
*[`-- commits` line 4 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/502#-NfWGBwPDTTXf89vwoOD):*
Typo `redundant`

Word-wrap at 80 chars.
___
*[`internal/target/schemawatch/dependencies.go` line 279 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/502#-NfWFWZP7oOV3gcqjLVr:-NfWFWZP7oOV3gcqjLVs:bknpfxr) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/1dc5b72ec0c74c58de666820c1d2f27c2d171d47/internal/target/schemawatch/dependencies.go#L279)):*
> ```Go
> 
> 			// Allow skipping a level. This might happen if there
> 			// are references across schemas.
> ```

Makes sense. Empty levels won't break current consumers, but it is weird.
___
*[`internal/target/schemawatch/dependencies_test.go` line 201 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/502#-NfWFxb52tKoOIcNSBdb:-NfWFxb60ncte2DpRYX6:b7s3eiv) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/1dc5b72ec0c74c58de666820c1d2f27c2d171d47/internal/target/schemawatch/dependencies_test.go#L201)):*
> ```Go
> }
> 
> func TestDifferentSchema(t *testing.T) {
> ```

`TestCrossSchemaTableReferencesPG`
___
*[`internal/target/schemawatch/dependencies_test.go` line 212 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/502#-NfWF8Yp292kHhDjnm0Y:-NfWF8Yp292kHhDjnm0Z:b-z4uzvx) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/1dc5b72ec0c74c58de666820c1d2f27c2d171d47/internal/target/schemawatch/dependencies_test.go#L212)):*
> ```Go
> 	pool := fixture.TargetPool
> 	switch pool.Product {
> 	// TODO (Silvano) add CRDB when https://github.com/cockroachdb/cockroach/issues/111419 is fixed.
> ```

It would be worth writing this case, even though CRDB returns the wrong data, if only to document what cdc-sink will do in this case.  There are some other tests in cdc-sink that will break if/when certain bugs in CRDB are fixed.
___
*[`internal/target/schemawatch/dependencies_test.go` line 232 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/502#-NfWFf8I38n2xiZDXl_c:-NfWFf8I38n2xiZDXl_d:bqd3h0q) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/1dc5b72ec0c74c58de666820c1d2f27c2d171d47/internal/target/schemawatch/dependencies_test.go#L232)):*
> ```Go
> 		case 0:
> 			r.Equal(1, len(tables))
> 			a.Equal(fmt.Sprintf(`%[1]s."parent"`, fixture.TargetSchema), tables[0].Canonical().String())
> ```

Use `a.True(ident.Equal(...))` instead of this string manipulation.

sravotto (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/502)* status: 1 of 2 files reviewed, 4 unresolved discussions (waiting on @bobvawter)

___
*[`internal/target/schemawatch/dependencies.go` line 279 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/502#-NfWFWZP7oOV3gcqjLVr:-NfXU1lk8KT524yo26Yn:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/1dc5b72ec0c74c58de666820c1d2f27c2d171d47/internal/target/schemawatch/dependencies.go#L279)):*

Makes sense. Empty levels won't break current consumers, but it is weird.
</blockquote></details>

Done.
___
*[`internal/target/schemawatch/dependencies_test.go` line 201 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/502#-NfWFxb52tKoOIcNSBdb:-NfXU3pm5HXqnvqIYZou:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/1dc5b72ec0c74c58de666820c1d2f27c2d171d47/internal/target/schemawatch/dependencies_test.go#L201)):*

`TestCrossSchemaTableReferencesPG`
</blockquote></details>

Done.
___
*[`internal/target/schemawatch/dependencies_test.go` line 212 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/502#-NfWF8Yp292kHhDjnm0Y:-NfXU5dR3Kl9KTwR9Cm7:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/1dc5b72ec0c74c58de666820c1d2f27c2d171d47/internal/target/schemawatch/dependencies_test.go#L212)):*

It would be worth writing this case, even though CRDB returns the wrong data, if only to document what cdc-sink will do in this case.  There are some other tests in cdc-sink that will break if/when certain bugs in CRDB are fixed.
</blockquote></details>

Done.
___
*[`internal/target/schemawatch/dependencies_test.go` line 232 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/502#-NfWFf8I38n2xiZDXl_c:-NfXU7Kn7zwxzOn2t6CP:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/1dc5b72ec0c74c58de666820c1d2f27c2d171d47/internal/target/schemawatch/dependencies_test.go#L232)):*

Use `a.True(ident.Equal(...))` instead of this string manipulation.
</blockquote></details>

Done.

bobvawter (approved):
:lgtm: with the one comment about using the existing query that does return the correct information. There's no overwhelming need to use the `information_schema` query with CRDB, it just seemed like the thing to do if/when we retire support for v21.

> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/502#-:-NfkSLayCSYeiLxswCVU:bx7sl71)* status: all files reviewed, 1 unresolved discussion (waiting on @sravotto)

___
*[`internal/target/schemawatch/dependencies_test.go` line 241 at r3](https://reviewable.io/reviews/cockroachdb/cdc-sink/502#-NfkQYjq5RpNgbBXDzu6:-NfkQYjq5RpNgbBXDzu7:b-zcwaje) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/1865cb8fa9db324f12f2c77066707a59a828257c/internal/target/schemawatch/dependencies_test.go#L241)):*
> ```Go
> 	r.NoError(fixture.Watcher.Refresh(ctx, pool))
> 	snap := fixture.Watcher.Get()
> 	if pool.Product == types.ProductCockroachDB && !strings.Contains(pool.Version, "v21.") {
> ```

If the CRDB-specific query that uses the custom porcelain already returns the correct information in the cross-schema case, we should use it, with a note to try to move back to `information_schema` in a future release of CRDB.  Update the `case` statement in `getDependency()` order with a note.

sravotto (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/502)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @bobvawter)

___
*[`internal/target/schemawatch/dependencies_test.go` line 241 at r3](https://reviewable.io/reviews/cockroachdb/cdc-sink/502#-NfkQYjq5RpNgbBXDzu6:-Nfkh5ZL046-r6uoZAZ4:b321iu4) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/1865cb8fa9db324f12f2c77066707a59a828257c/internal/target/schemawatch/dependencies_test.go#L241)):*

If the CRDB-specific query that uses the custom porcelain already returns the correct information in the cross-schema case, we should use it, with a note to try to move back to `information_schema` in a future release of CRDB.  Update the `case` statement in `getDependency()` order with a note.
</blockquote></details>

Good point. I fixed it.

PR #506, opened 2023-09-29
mysql target: getting schema information from database

sravotto:
This is the initial work to support mysql as a target, focusing on getting the schema information. 
Subsequent PRs will address the logic to apply mutations and  completing the backed connection support (with TLS, and better error handling).

Summary of changes:

- Added mysql connection pool. Not supporting TLS yet.
- Added templates for querying column information and constraint dependencies.
- Updated dependencies test to be compatible with mysql syntax.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/506)

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -301,6 +301,7 @@ github.com/jackc/puddle/v2 v2.2.1/go.mod h1:vriiEXHvEE654aYKXXjOvZM39qJ0q+azkZFr
 github.com/jbenet/go-context v0.0.0-20150711004518-d14ea06fba99 h1:BQSFePA1RWJOlocH6Fxy8MmwDt+yVQYULKfN0RoTN8A=
 github.com/jbenet/go-context v0.0.0-20150711004518-d14ea06fba99/go.mod h1:1lJo3i6rXxKeerYnT8Nvf0QmHCRC1n8sfWVwXF2Frvo=
 github.com/jessevdk/go-flags v1.4.0/go.mod h1:4FA24M0QyGHXBuZZK/XkWh8h0e1EYbRYJSGM75WSRxI=
+github.com/jmoiron/sqlx v1.3.3 h1:j82X0bf7oQ27XeqxicSZsTU5suPwKElg3oyxNn43iTk=
 github.com/jmoiron/sqlx v1.3.3/go.mod h1:2BljVx/86SuTyjE+aPYlHCTNvZrnJXghYGpNiXLBMCQ=
 github.com/joonix/log v0.0.0-20200409080653-9c1d2ceb5f1d h1:k+SfYbN66Ev/GDVq39wYOXVW5RNd5kzzairbCe9dK5Q=
 github.com/joonix/log v0.0.0-20200409080653-9c1d2ceb5f1d/go.mod h1:fS54ONkjDV71zS9CDx3V9K21gJg7byKSvI4ajuWFNJw=
--- a/internal/sinktest/base/provider.go
+++ b/internal/sinktest/base/provider.go
@@ -314,7 +314,7 @@ func provideSchema[P types.AnyPool](
 	ctx context.Context, pool P, prefix string,
 ) (ident.Schema, func(), error) {
 	switch pool.Info().Product {
-	case types.ProductCockroachDB, types.ProductPostgreSQL:
+	case types.ProductCockroachDB, types.ProductMySQL, types.ProductPostgreSQL:
 		return CreateSchema(ctx, pool, prefix)
 
 	case types.ProductOracle:
@@ -388,8 +388,13 @@ func CreateSchema[P types.AnyPool](
 			return ident.Schema{}, cancel, errors.WithStack(err)
 		}
 	}
-
-	sch, err := ident.NewSchema(name, ident.Public)
+	var sch ident.Schema
+	var err error
+	if pool.Info().Product == types.ProductMySQL {
+		sch, err = ident.NewSchema(name)
+	} else {
+		sch, err = ident.NewSchema(name, ident.Public)
+	}
 	if err != nil {
 		return ident.Schema{}, cancel, err
 	}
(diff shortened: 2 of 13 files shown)


bobvawter (approved):
:lgtm: w/ nits

Once the mysql support lands, I'm thinking about extracting a `GetSqlCode(err error) string` function someplace to centralize pulling the error codes from the variety of error types that we may encounter.  The `IsStartupError()` function would also be hoisted out.

> Reviewed 7 of 15 files at r1, 8 of 8 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/506#-:-NfkWq6O7ILLkh7wTnW-:b-yjb54n)* status: all files reviewed, 8 unresolved discussions (waiting on @BramGruneir and @sravotto)

___
*[`internal/sinktest/base/provider.go` line 393 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/506#-NfkT9WzAAT4LUSirbCT:-NfkT9WzAAT4LUSirbCU:bkofr6) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/2af64958e6fb6641395cb0e4c6eba813f124e538/internal/sinktest/base/provider.go#L393)):*
> ```Go
> 	var sch ident.Schema
> 	var err error
> 	if pool.Info().Product == types.ProductMySQL {
> ```

It seems like this `if` statement isn't getting called with Oracle?  Can you verify coverage?
___
*[`internal/target/schemawatch/coldata.go` line 327 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/506#-NfkU12Z5y6Te1hROGJW:-NfkU12Z5y6Te1hROGJX:b-eercl) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/2af64958e6fb6641395cb0e4c6eba813f124e538/internal/target/schemawatch/coldata.go#L327)):*
> ```Go
> 				if defaultExpr.Valid {
> 					switch column.Type {
> 					// wrapping default values in single quotes for strings
> ```

Can you expand this comment a bit more as to what the data coming back from mysql looks like.  If I'm reading this correctly, if you declare some column as `name DEFAULT 'foo'` the contents of defaultExpr are just `foo`?
___
*[`internal/target/schemawatch/coldata.go` line 336 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/506#-NfkUJFG6EKsdOkVtDsC:-NfkUJFG6EKsdOkVtDsD:b-x252zq) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/2af64958e6fb6641395cb0e4c6eba813f124e538/internal/target/schemawatch/coldata.go#L336)):*
> ```Go
> 			case types.ProductOracle:
> 				if defaultExpr.Valid {
> 					column.DefaultExpr = strings.TrimSpace(defaultExpr.String)
> ```

Retain original comment about Oracle returning extraneous whitespace.
___
*[`internal/target/schemawatch/coldata_test.go` line 282 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/506#-NfkUVs03mO44oOG-EhN:-NfkUVs03mO44oOG-EhO:bllu1la) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/2af64958e6fb6641395cb0e4c6eba813f124e538/internal/target/schemawatch/coldata_test.go#L282)):*
> ```Go
> 	switch fixture.TargetPool.Product {
> 	case types.ProductMySQL:
> 		// MySQL does not support TYPES
> ```

But it does support enums, which is what this test case wants to test.  Hopefully, enums just look like boring strings and don't have custom types behind them.  https://dev.mysql.com/doc/refman/8.0/en/enum.html
___
*[`internal/target/schemawatch/dependencies.go` line 41 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/506#-NfkUrPQ1Afhbt9LM0Ev:-NfkUrPQ1Afhbt9LM0Ew:b-r75wfx) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/2af64958e6fb6641395cb0e4c6eba813f124e538/internal/target/schemawatch/dependencies.go#L41)):*
> ```Go
> //   - depths: Recursively computes the depth of each table from the root, walking
> //     the foreign keys constraints.
> //   - cycle_detect: insure that all tables have a depth, by injecting a default depth.
> ```

ensure
___
*[`internal/target/schemawatch/dependencies.go` line 338 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/506#-NfkVSiyDdpa45Si6zoQ:-NfkVSiyDdpa45Si6zoR:b6zrrkh) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/2af64958e6fb6641395cb0e4c6eba813f124e538/internal/target/schemawatch/dependencies.go#L338)):*
> ```Go
> 		args = []any{sql.Named("owner", db.Raw())}
> 	default:
> 		return nil, errors.Errorf("getDependencyOrder unimplemented product: %s", tx.Product)
> ```

Good catch.
___
*[`internal/target/schemawatch/dependencies_test.go` line 128 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/506#-NfkVgknAJ5-EAeBI2Mf:-NfkVgknAJ5-EAeBI2Mg:b6hxepj) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/2af64958e6fb6641395cb0e4c6eba813f124e538/internal/target/schemawatch/dependencies_test.go#L128)):*
> ```Go
> 	for idx, tc := range tcs {
> 		sql := fmt.Sprintf(tc.schema, fixture.TargetSchema.Schema())
> 		log.Print(sql)
> ```

Lower to `log.Trace()`. It's useful, but I'm going to take a pass over the tests to reduce the default logging output.
___
*[`go.mod` line 13 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/506#-NfWIgrK8nRLRzQMpJTM:-NfWIgrK8nRLRzQMpJTN:bh0pls3) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/7d8fd6db74872d88c7060ba459bd91b7acd7d6ca/go.mod#L13)):*
> ```XML
> 	github.com/evanw/esbuild v0.19.3
> 	github.com/go-mysql-org/go-mysql v1.6.0
> 	github.com/go-sql-driver/mysql v1.7.1
> ```

Why add a second mysql driver?  The package that we're already using supports the `sql.DB` interface https://github.com/go-mysql-org/go-mysql#driver

PR #513, opened 2023-10-02
logical: Fix deletes sent to userscript map function

### LINKED ISSUE #507: userscript: Delete sent to table with mapper and no dispatcher results in json parse error
The call to `json.Unmarshal` will fail at
https://github.com/cockroachdb/cdc-sink/blob/7988ff8081bb2ecbc63776e04a8934a0cd438cfc/internal/script/script.go#L274

if `scriptBatch.OnData` takes one of the early-out branches here:
https://github.com/cockroachdb/cdc-sink/blob/7988ff8081bb2ecbc63776e04a8934a0cd438cfc/internal/source/logical/script_events.go#L58-L66

This can be triggered in the fslogical frontend if there's a tombstone collection that drives a destination table with a map function.

API design question: Should the user-defined mapper function be called for deletes (e.g. to be able to filter them?) What should the input data look like?

bobvawter:
There is a bug in the current flow where a delete mutation can be sent to a userscript table's map function if there is no dispatch function defined on the source. There is no data in a deletion to map, so the operation doesn't make sense and the function binding returns an error as there is no payload.

This change reworks scriptEvents.OnData() so that any configuration of dispatch, map, etc. have the same flow.

Fixes #507

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/513)

### DIFF
--- a/internal/source/logical/logical_test.go
+++ b/internal/source/logical/logical_test.go
@@ -405,3 +405,60 @@ api.configureTable("t_2", {
 	a.Equal(100, count)
 
 }
+
+// https://github.com/cockroachdb/cdc-sink/issues/507
+func TestDeletesWithNoDispatch(t *testing.T) {
+	r := require.New(t)
+
+	// Create a basic test fixture.
+	fixture, cancel, err := base.NewFixture()
+	r.NoError(err)
+	defer cancel()
+
+	ctx := fixture.Context
+	targetSchema := fixture.TargetSchema.Schema()
+	pool := fixture.TargetPool
+
+	// Create some tables.
+	tgt := ident.NewTable(targetSchema, ident.New("tgt"))
+	var schema = fmt.Sprintf(`CREATE TABLE %s (k INT PRIMARY KEY, v VARCHAR(2048), ref INT)`, tgt)
+	_, err = pool.ExecContext(ctx, schema)
+	r.NoError(err)
+
+	cfg := &logical.BaseConfig{
+		ApplyTimeout:   2 * time.Minute, // Increase to make using the debugger easier.
+		Immediate:      false,
+		StagingConn:    fixture.StagingPool.ConnectionString,
+		StagingSchema:  fixture.StagingDB.Schema(),
+		StandbyTimeout: 5 * time.Millisecond,
+		TargetConn:     pool.ConnectionString,
+
+		ScriptConfig: script.Config{
+			MainPath: "/main.ts",
+			FS: &fstest.MapFS{
+				"main.ts": &fstest.MapFile{Data: []byte(`
+import * as api from "cdc-sink@v1";
+api.configureTable("tgt", {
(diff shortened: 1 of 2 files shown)


sravotto (approved):
:lgtm:

> Reviewed 1 of 2 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/513#-:-NfvL9t3C_bn6yASEQfW:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @BramGruneir)

BramGruneir (approved):
:lgtm:

> Reviewed 1 of 2 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/513#-:-Ng-dDqQ14tFbSh2Puts:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @bobvawter)

___
*[`internal/source/logical/script_events.go` line 67 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/513#-Ng-L2rh7A8p8ILTJ1eb:-Ng-L2rh7A8p8ILTJ1ec:bt6cx4m) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/d011778350dbbf499179be522784c5f8750f9d36/internal/source/logical/script_events.go#L67)):*
> ```Go
> 	var routing ident.TableMap[[]types.Mutation]
> 
> 	if cfg, ok := e.Script.Sources.Get(source); !ok {
> ```

The logic here is a bit hard to read.  Maybe just add some comments to make it clear.    Just because you're doing two lookups and on either not existing you do the default behaviour.

I'm trying to think of how to make it more readable and I can't quite nail it.

bobvawter (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/513)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @BramGruneir and @sravotto)

___
*[`internal/source/logical/script_events.go` line 67 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/513#-Ng-L2rh7A8p8ILTJ1eb:-Ng079Hn9SOa01XhHt5l:b-o4c0vm) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/d011778350dbbf499179be522784c5f8750f9d36/internal/source/logical/script_events.go#L67)):*

The logic here is a bit hard to read.  Maybe just add some comments to make it clear.    Just because you're doing two lookups and on either not existing you do the default behaviour.

I'm trying to think of how to make it more readable and I can't quite nail it.
</blockquote></details>

I added more commentary throughout the method.  I suspect that it feels weird because the else block depends on both `cfg` and `dispatch`, which isn't typical.

PR #515, opened 2023-10-05
apply: Allow templates to be loaded from disk

bobvawter:
This change adds an environment variable that allows the golang templates to be loaded from the filesystem, instead of using the ones embedded in the binary. A hidden `cdc-sink dumptemplates` command is added to copy the templates to disk.

The indended use-case for this change is to support in-the-field adjuments to the query templates, without the need to recompile the binary. No long-term compatibility guarantees can be made about the contents and structure of the template files, since they are tightly coupled with internal cdc-sink datastructures.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/515)

### DIFF
--- a/internal/cmd/dumptemplates/dumptemplates.go
+++ b/internal/cmd/dumptemplates/dumptemplates.go
@@ -1,0 +1,54 @@
+// Copyright 2023 The Cockroach Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+// Package dumptemplates contains a hidden command that will write
+// the apply templates to disk. It is for in-the-field debugging use
+// only and does not constitute a stable API.
+package dumptemplates
+
+import (
+	"fmt"
+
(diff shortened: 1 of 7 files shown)


BramGruneir (changes requested):
I know it's only for debugging, but a test would be helpful.

And instead of going whole hog, why not only load templates that are present and use the embedded ones otherwise.  (And add a log message to say which is which)

> Reviewed 3 of 3 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/515#-:-Ng07V-Z0-B7iuaocVg_:b-5jbd7d)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sravotto)

bobvawter (commented):
Added tests. PTAL.

It would be possible to write an `overlayFS` that merges two `fs.FS` together, but that seems like a lot of over-engineering when the total number of files is pretty minimal.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/515#-:-Ng0S8BIELW8XJacehkV:b-8y3eb1)* status: 1 of 7 files reviewed, all discussions resolved (waiting on @BramGruneir and @sravotto)

sravotto (approved):
:lgtm:

> Reviewed 6 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/515#-:-Ng2-0Y4-14GWYzAFRQJ:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @bobvawter)

BramGruneir (approved):
:lgtm:

> Reviewed 6 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/515#-:-Ng52dZC0ErqVMPoe-eM:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @bobvawter)

Issue #487, opened 2023-09-18
Three-way merge


bobvawter:
If a changefeed is created with the `diff` option, we should be able to compare the `before` data to the `target` row. If they differ, we can call the userscript with `before`, `after`, and `target` data to allow for custom merge operations. We can also tie into the existing CAS behaviors by using the time/version field(s) to detect conflicts.

PR #517, opened 2023-10-06
stage: Support storing a Mutation.Before field

bobvawter:
This change is part of #487 to support three-way merges. It adds a before field to types.Mutation and allows this data to be persisted.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/517)

### DIFF
--- a/internal/staging/stage/factory.go
+++ b/internal/staging/stage/factory.go
@@ -17,11 +17,8 @@
 package stage
 
 import (
-	"bytes"
-	"compress/gzip"
 	"context"
 	"fmt"
-	"io"
 	"strings"
 	"sync"
 	"time"
@@ -163,7 +160,7 @@ data AS (`)
 				if id == offsetTableIdx {
 					// We're resuming in the middle of reading a table.
 					_, _ = fmt.Fprintf(&sb,
-						"(SELECT %[1]d AS t, nanos, logical, key, mut "+
+						"(SELECT %[1]d AS t, nanos, logical, key, mut, before "+
 							"FROM %[2]s "+
 							"WHERE (nanos, logical, key) > ($6, $7, $8) "+
 							"AND (nanos, logical) <= ($3, $4) "+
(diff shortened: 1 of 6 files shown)


BramGruneir (approved):
:lgtm: if you add the tests.

But how much do we really gain by zipping the mutations?  
Because on disk, they're getting compressed right?  In pebble?  

I can see the argument that it reduces the size of the row, but I don't recall if we actually have a hard limit there.

The max size is going to be based on the originating row*2 with extra metadata... So perhaps we should consider column families for those two columns if it becomes a problem.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/517#-:-Ng56Mev5kklJkUBYINa:bpzgwfl)* status: all files reviewed, 2 unresolved discussions (waiting on @bobvawter and @sravotto)

___
*[`internal/staging/stage/stage.go` line 441 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/517#-Ng5660eDJr9BXyolcyE:-Ng5660eDJr9BXyolcyF:b-ud1lga) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/4fd98c3d6755560dfdc8014f58f2a6360282fd10/internal/staging/stage/stage.go#L441)):*
> ```Go
> // value. GZIP has about 1k of overhead for the dictionary, so zipping
> // small amounts of data isn't actually going to buy us anything.
> func maybeGZip(data []byte) ([]byte, error) {
> ```

Ugg... these probably need a few tests.  Sorry, I know they'll be tedious.
___
*[`internal/staging/stage/stage.go` line 463 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/517#-Ng55o5d7UIZBVU0GYxB:-Ng55o5d7UIZBVU0GYxC:b-henppz) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/4fd98c3d6755560dfdc8014f58f2a6360282fd10/internal/staging/stage/stage.go#L463)):*
> ```Go
> // they're present. The magic numbers would not be present in JSON.
> func maybeGunzip(data []byte) ([]byte, error) {
> 	if len(data) >= 2 && data[0] == 0x1f && data[1] == 0x8b {
> ```

Minor nit: Just negate the if and put the non-gziped exit first to reduce indentation.

bobvawter:
Added tests for the extracted helper functions.  Comparing the contents `Before` field is already handled in that test, since it checks the retrieved mutations against the originals via `assert.Equal()`.

The data's stored in a compressed form in Pebble, but it's really easy to wind up with 10k of JSON that compresses down to a much smaller amount of data to actually send over the network to CRDB.  IIRC, the pgwire protocol isn't compressed and we never try to process the mutation blob within the database.

PR #532, opened 2023-10-09
oracle: Use upstream go-ora

bobvawter:
The workaround that was introduced in PR #508 has been upstreamed. This change removes the local override and bumps the go-ora version.

X-Ref: https://github.com/sijms/go-ora/pull/445

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/532)


BramGruneir:
LGTM

PR #533, opened 2023-10-10
apply: Comparing the target and staging values in TestAllDataTypes

sravotto:
The TestAllDataTypes was incorrectly comparing the values stored in the target database with the values read from the same database.

The change verifies the values stored in the target database with the values stored in the staging database.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/533)


bobvawter (approved):
LGTM. Good catch.

PR #534, opened 2023-10-11
script: Allow user-defined merge functions

bobvawter:
This change is part of #487 to support three-way merges.

This change add supports for declaring a user-defind merge function within the userscript. The goja version is updated so that we can create a lightweight wrapper around the ident.Map that will store the reified mutation values.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/534)


sravotto (approved):
:lgtm: Just a minor comment.

> Reviewed 11 of 11 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/534#-:-NgV1J5mElH8O7ogDSm8:bwcg8ny)* status: all files reviewed, 1 unresolved discussion (waiting on @bobvawter and @BramGruneir)

___
*[`internal/script/ident_map.go` line 32 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/534#-NgUt1DF83eC9I3YGpdJ:-NgUt1DGCZjQJUZoMOc7:b3f8nn2) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/1c00c4a3586cc96ffe4d2ff3d812dafa6567ff8e/internal/script/ident_map.go#L32)):*
> ```Go
> var _ goja.DynamicObject = (*identMapWrapper)(nil)
> 
> func (m *identMapWrapper) Delete(key string) bool {
> ```

why returning the bool if it is always true?
Docs missing in these functions.

BramGruneir (changes requested):
> Reviewed 11 of 11 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/534)* status: all files reviewed, 7 unresolved discussions (waiting on @bobvawter)

___
*[`-- commits` line 8 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/534#-NgsnGjpEixmzANZqn8X):*
Could you add a bit of info about what is not yet implemented?
___
*[`internal/script/loader.go` line 63 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/534#-NgshAXrA2R3GXSwwecW:-NgshAXrA2R3GXSwwecX:bjr4npv) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/1c00c4a3586cc96ffe4d2ff3d812dafa6567ff8e/internal/script/loader.go#L63)):*
> ```Go
> type mergeResult struct {
> 	Apply goja.Value `goja:"apply"` // The data to return
> 	DLQ   string     `goja:"dlq"`   // Append to a dead-letter queue.
> ```

Where do these DLQs exist?  One table with the DQL name as part of thePK?

Also, what are the valid configurations?  Can you have a drop and a dlq and an apply?
___
*[`internal/script/script.go` line 350 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/534#-NgsjG9E9w2Y-4_gCUKu:-NgsjG9E9w2Y-4_gCUKv:b-550pe9) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/1c00c4a3586cc96ffe4d2ff3d812dafa6567ff8e/internal/script/script.go#L350)):*
> ```Go
> func (s *UserScript) bindMerge(table ident.Table, merger mergeJS) types.MergeFunc {
> 	return func(ctx context.Context, tx types.TargetQuerier, con *types.Conflict) (*types.Resolution, bool, error) {
> 		// con.Before will be nil in 2-way merge.
> ```

I think the order here is quite important.  (e.g. Why not put the drop at the top? or put the DQL up higher since we know the merge isn't happening? )

More comments are needed here to fully understand the logic and ordering of the check.  
___
*[`internal/script/testdata/cdc-sink@v1.d.ts` line 227 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/534#-NgslxPk8gZEbsnMGjJA:-NgslxPk8gZEbsnMGjJB:b-lbb9g) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/1c00c4a3586cc96ffe4d2ff3d812dafa6567ff8e/internal/script/testdata/cdc-sink%40v1.d.ts#L227)):*
> ```XML
>          * option will cause data loss and a DLQ may be a better option.
>          */
>         drop: true
> ```

Boolean?
___
*[`internal/script/testdata/main.ts` line 79 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/534#-Ngsm6vOChqIAFAnDVb1:-Ngsm6vOChqIAFAnDVb2:b-4qixv3) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/1c00c4a3586cc96ffe4d2ff3d812dafa6567ff8e/internal/script/testdata/main.ts#L79)):*
> ```XML
>     // This shows how a counter-like value can be merged.
>     merge: op => {
>         console.log("merge", JSON.stringify(op));
> ```

Is this leftover from testing?
___
*[`internal/types/merge.go` line 55 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/534#-Ngsn2WZ6HboW-Qq3I8O:-Ngsn2WZ6HboW-Qq3I8P:b95ykjv) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/1c00c4a3586cc96ffe4d2ff3d812dafa6567ff8e/internal/types/merge.go#L55)):*
> ```Go
> 	// of the Merger to ensure that the conflicted value arrived
> 	// somewhere.
> 	Merge(context.Context, TargetQuerier, *Conflict) (*Resolution, bool, error)
> ```

Why not just rely on the nil of Resolution to indicate that the conflict should be ignored?  Just to ensure that it was deliberate and not returned in error?

BramGruneir:
*[`-- commits` line 8 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/534#-NgsnGjpEixmzANZqn8X):*

Could you add a bit of info about what is not yet implemented?
</blockquote></details>

I'm thinking the DQL entirely.  And really anything post conflict script being run.

bobvawter:
Closing in favor of #540 which includes the base commit and already addresses several of the comments raised.

PR #538, opened 2023-10-12
sinktest: Generate names that must be quoted

bobvawter:
This change updates the CreateTable and CreateSchema functions to generate
names that contain dashes. This helps to ensure that all identifiers are
correctly quoted when we generate SQL commands.

This change additionally corrects some test failures around parsing ndjson URLs
that contain dashes. There is also a correction made to a schema introspection
query that was not correctly using a quoted ident type. There are a few
naming-related changes in the pglogical and mylogical packages, which are
explained by surrounding comments.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/538)

### DIFF
--- a/internal/sinktest/base/provider.go
+++ b/internal/sinktest/base/provider.go
@@ -362,7 +362,9 @@ func CreateSchema[P types.AnyPool](
 	// Each package tests run in a separate binary, so we need a
 	// "globally" unique ID.  While PIDs do recycle, they're highly
 	// unlikely to do so during a single run of the test suite.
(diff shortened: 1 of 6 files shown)


BramGruneir (approved):
:lgtm:

We really need better output form CDC.  A more formal spec for both the endpoints and some metadata if we can't get protos.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/538#-:-Ng_vdVkA6o0Rhw3A19a:b-btcg4j)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sravotto)

PR #539, opened 2023-10-13
mysql: conditional templates, completing test coverage for target.

sravotto:
Adding the template to support conditional update flow (compare-and-set, deadlines).

Adding template test data for mysql.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/539)

### DIFF
--- a/internal/target/apply/apply_test.go
+++ b/internal/target/apply/apply_test.go
@@ -70,6 +70,10 @@ func TestApply(t *testing.T) {
 		// control the exact bytes that come back.
 		tableSchema = "CREATE TABLE %s (pk0 INT, pk1 VARCHAR(2048), extras VARCHAR(2048), " +
 			"has_default VARCHAR(2048) NOT NULL DEFAULT 'Hello', PRIMARY KEY (pk0,pk1))"
+	case types.ProductMySQL:
+		// mysql has limitation on the size of the primary key (3072 bytes)
+		tableSchema = "CREATE TABLE %s (pk0 INT, pk1 VARCHAR(512), extras VARCHAR(2048), " +
+			"has_default VARCHAR(2048) NOT NULL DEFAULT 'Hello', PRIMARY KEY (pk0,pk1))"
 	case types.ProductOracle:
 		// The lower-case names in the payload will be matched to the
 		// upper-case names that exist within the database.
@@ -89,6 +93,8 @@ func TestApply(t *testing.T) {
 		switch fixture.TargetPool.Product {
 		case types.ProductCockroachDB, types.ProductPostgreSQL:
 			q = "SELECT count(*) FROM %s WHERE has_default=$1"
+		case types.ProductMySQL:
+			q = "SELECT count(*) FROM %s WHERE has_default=?"
 		case types.ProductOracle:
 			q = "SELECT count(*) FROM %s WHERE has_default=:p1"
 		default:
@@ -170,6 +176,8 @@ func TestApply(t *testing.T) {
 			switch fixture.TargetPool.Product {
 			case types.ProductCockroachDB, types.ProductPostgreSQL:
 				a.ErrorContains(err, "violates not-null constraint (SQLSTATE 23502)")
+			case types.ProductMySQL:
+				a.ErrorContains(err, "ERROR 1048 (23000)")
 			case types.ProductOracle:
 				a.ErrorContains(err, "ORA-01400: cannot insert NULL into")
 			default:
@@ -259,6 +267,8 @@ func TestApply(t *testing.T) {
 		switch fixture.TargetPool.Product {
 		case types.ProductCockroachDB, types.ProductPostgreSQL:
 			extrasQ = fmt.Sprintf("SELECT extras FROM %s WHERE pk0=$1 AND pk1=$2", tbl.Name())
+		case types.ProductMySQL:
+			extrasQ = fmt.Sprintf("SELECT extras FROM %s WHERE pk0=? AND pk1=?", tbl.Name())
 		case types.ProductOracle:
 			extrasQ = fmt.Sprintf(`SELECT extras FROM %s WHERE pk0=:p1 AND pk1=:p2`, tbl.Name())
 		default:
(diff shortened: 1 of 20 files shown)


bobvawter (changes requested):
This looks like a good start, but I think you should enable mysql as a target in the test cases to shake out edge cases.  Look in `.github/workflows/go-tests.yaml` around line 140 where CRDB -> PG is set up.

> Reviewed 1 of 23 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/539#-:-NgsWliw5YR4T3UWiOfo:bj7hcp4)* status: 1 of 23 files reviewed, 4 unresolved discussions (waiting on @BramGruneir and @sravotto)

___
*[`internal/target/apply/apply_test.go` line 1501 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/539#-NgsXArG9-yCrBlrcqpK:-NgsXArG9-yCrBlrcqpL:bz7xo6m) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/ad3d9ab5149f8f1d69242677223b3617c2cbe39b/internal/target/apply/apply_test.go#L1501)):*
> ```Go
> 	a := assert.New(t)
> 
> 	fixture, cancel, err := all.NewFixture()
> ```

You could pass the fixture into this helper.
___
*[`internal/target/apply/queries/my/conditional.tmpl` line 43 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/539#-NgsW1LtAQRbFjjaVo6y:-NgsW1LtAQRbFjjaVo6z:b9xangx) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/ad3d9ab5149f8f1d69242677223b3617c2cbe39b/internal/target/apply/queries/my/conditional.tmpl#L43)):*
> ```
> */ -}}
> {{- nl -}}
> WITH data AS (SELECT {{ template "pairs" .Columns -}} ) 
> ```

What happens if there's more than one row being upserted?

If the mysql driver will accept data in columnar format, you could enable `BulkUpdate` and/or `BulkDelete` in the `newTemplates()` function like we do for Oracle.
___
*[`internal/target/apply/testdata/my/cas.upsert.sql` line 3 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/539#-NgsXwFw1TrSpiVnGp-4:-NgsXwFx8mz5h7Arwfui:b2ixvr9) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/ad3d9ab5149f8f1d69242677223b3617c2cbe39b/internal/target/apply/testdata/my/cas.upsert.sql#L3)):*
> ```TSQL
> INSERT
> INTO"schema"."table"("pk0","pk1","val0","val1","has_default")
> WITH data AS (SELECT ? AS "pk0",? AS "pk1",? AS "val0",? AS "val1",? AS "has_default"),
> ```

MySQL appears to support naming the CTE columns directly, like the other templates do:

https://dev.mysql.com/doc/refman/8.0/en/with.html

`WITH cte (col1, col2) AS ( SELECT 1, 2 UNION ALL SELECT 3, 4 ) SELECT col1, col2 FROM cte;`
___
*[`internal/target/apply/testdata/my/expr.upsert.sql` line 4 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/539#-NgsYBByDpe0vbvRBeDi:-NgsYBByDpe0vbvRBeDj:b9t587w) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/ad3d9ab5149f8f1d69242677223b3617c2cbe39b/internal/target/apply/testdata/my/expr.upsert.sql#L4)):*
> ```TSQL
> ("pk0","pk1","val0","val1","has_default")
> VALUES
> (?,(?+?),('fixed'),(?||'foobar'),CASE WHEN ? THEN ? ELSE expr() END),
> ```

Is that going to do the correct thing if the user intends to insert an explicit NULL?  That's why the other templates have the default value guarded by a validity marker.

bobvawter (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/539)* status: 1 of 23 files reviewed, 3 unresolved discussions (waiting on @BramGruneir and @sravotto)

___
*[`internal/target/apply/testdata/my/cas.upsert.sql` line 3 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/539#-NgsXwFw1TrSpiVnGp-4:-NgsoXDq9_pPsiphvJg7:b-gmjc34) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/ad3d9ab5149f8f1d69242677223b3617c2cbe39b/internal/target/apply/testdata/my/cas.upsert.sql#L3)):*

MySQL appears to support naming the CTE columns directly, like the other templates do:

https://dev.mysql.com/doc/refman/8.0/en/with.html

`WITH cte (col1, col2) AS ( SELECT 1, 2 UNION ALL SELECT 3, 4 ) SELECT col1, col2 FROM cte;`
</blockquote></details>

(Moving comment to the correct line)

Also, is the has_default expression going to do the correct thing if the user intends to insert an explicit NULL? That's why the other templates have the default value guarded by a validity marker.

It seems like there's there's a common template expansion that's missing.

sravotto (commented):
Please take another look. Integration tests on a different PR.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/539#-:-NgySlLL-gPgie31wEUj:b-f2ufne)* status: 1 of 23 files reviewed, 3 unresolved discussions (waiting on @bobvawter and @BramGruneir)

___
*[`internal/target/apply/apply_test.go` line 1501 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/539#-NgsXArG9-yCrBlrcqpK:-NgyML1Q3ayVuW6t2xpe:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/ad3d9ab5149f8f1d69242677223b3617c2cbe39b/internal/target/apply/apply_test.go#L1501)):*

You could pass the fixture into this helper.
</blockquote></details>

Done.
___
*[`internal/target/apply/queries/my/conditional.tmpl` line 43 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/539#-NgsW1LtAQRbFjjaVo6y:-NgyMSN38QkZT6gV2nFw:bh52bv8) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/ad3d9ab5149f8f1d69242677223b3617c2cbe39b/internal/target/apply/queries/my/conditional.tmpl#L43)):*

What happens if there's more than one row being upserted?

If the mysql driver will accept data in columnar format, you could enable `BulkUpdate` and/or `BulkDelete` in the `newTemplates()` function like we do for Oracle.
</blockquote></details>

Changed the template to produce multiple rows. It doesn't look like this driver supports the columnar format.
___
*[`internal/target/apply/testdata/my/cas.upsert.sql` line 3 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/539#-NgsXwFw1TrSpiVnGp-4:-NgsfF7l5evVGXDt3Erc:buq3p2k) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/ad3d9ab5149f8f1d69242677223b3617c2cbe39b/internal/target/apply/testdata/my/cas.upsert.sql#L3)):*

(Moving comment to the correct line)

Also, is the has_default expression going to do the correct thing if the user intends to insert an explicit NULL? That's why the other templates have the default value guarded by a validity marker.

It seems like there's there's a common template expansion that's missing.
</blockquote></details>

It seems that when using `INSERT`, you have to use `INSERT ... WITH ... SELECT ...` (see the link you provided).
I also had to use `SELECT`.. `UNION SELECT,` as `VALUES` didn't seem to work.

PR #540, opened 2023-10-14
apply: Support merge function for CockroachDB and PG targets

bobvawter:
This PR includes the commit that was part of #534 and adds the wiring int the apply package.  Having a consumer of the API forced a few design revisions.  If it's easier, these two commits can be squashed before the final merge.

This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/540)

PR #543, opened 2023-10-16
dlq: Add dead-letter queue package

bobvawter:
Review notes: This is being sent in parallel to #540, which will only need a minor PR to call this API when they're both added. If we decide that cdc-sink should create schemas in the target data (why not is outlined below), we already have known-usable schemas in this PR for testing.

This change is part of #487 to support three-way merges.

This change adds, but does not integrate, support for a per-target-schema DLQ. cdc-sink, as a design rule, does not perform any schema changes in the target database. If the user wants to use the DLQ, they will need to create the destination table. The only requirements on the DLQ table are that it contain certain well-known column names with appropriate types. A basic schema is suggested by cdc-sink, and this suggested schema is used by the DLQ tests.

The justification for this hand-waving is that the DLQ becomes, in essence, a part of the user's application and will likely need to be part of a schema-management system. We cannot predict how the DLQ entries will be used, indexed, etc. so integration with a minimum number of well-known columns seems like it should give the user maximum flexibility.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/543)

### DIFF
--- a/internal/sinktest/all/fixture.go
+++ b/internal/sinktest/all/fixture.go
@@ -18,12 +18,16 @@ package all
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/cockroachdb/cdc-sink/internal/sinktest/base"
 	"github.com/cockroachdb/cdc-sink/internal/staging/version"
+	"github.com/cockroachdb/cdc-sink/internal/target/dlq"
 	"github.com/cockroachdb/cdc-sink/internal/types"
 	"github.com/cockroachdb/cdc-sink/internal/util/applycfg"
 	"github.com/cockroachdb/cdc-sink/internal/util/diag"
+	"github.com/cockroachdb/cdc-sink/internal/util/ident"
+	"github.com/pkg/errors"
 )
 
 // Fixture provides a complete set of database-backed services. One can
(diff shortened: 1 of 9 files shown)


bobvawter:
I'm open to being convinced that we, in fact, should create the DLQ table in the target schema.  This behavior could be hidden behind a flag.  It would DTRT in the simple case and still play nicely with folks using a schema management tool.  Let me know what you think.

sravotto (approved):
:lgtm:  
I think we should provide an option to create the DLQ table in the target (by default it won't be created).

> Reviewed 8 of 10 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/543#-:-Nh-Yfhx8p35GdqNSSLy:b-t2wi4q)* status: 8 of 10 files reviewed, all discussions resolved (waiting on @BramGruneir)

BramGruneir (approved):
:lgtm:

I'm against cdc-sink creating the DLQ tables.  As mentioned, they may be an integral part of the business logic.  So this should be a very deliberate choice by the user.  

> Reviewed 10 of 10 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/543#-:-Nh2tiDJ01BxaNudQtfN:bpez0xm)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @bobvawter)

___
*[`internal/target/dlq/dlq_schema.go` line 37 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/543#-Nh2tSVxBPOJOo4inMUp:-Nh2tSVxBPOJOo4inMUq:b6r59nl) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/aad561548e8d202211dc83f604eed31fac8e171a/internal/target/dlq/dlq_schema.go#L37)):*
> ```Go
> }
> 
> const (
> ```

You might eventually want to move these to templates.  But this is fine for now.

PR #545, opened 2023-10-18
Testing output to RedShift

BramGruneir:
This adds postgres v8 to the test suite to test compatibility with RedShift.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/545)


BramGruneir:
Moving data to redshift should be done via s3 instead of directly.

PR #547, opened 2023-10-18
merge: Implement a standard merge function

bobvawter:
This change adds merge.Standard and exposes it to the userscript. The merge function identifies the properties which have changed between the before and proposed bags. If the value of the property in the before bag equals the value of the property in the target, the change is applied.

Property equivalency is presently defined as "serializes to the same JSON bytes". The golang json serializer is deterministic and we have a rather fluid typesystem, so this seems like a reasonable initial implementation.

A fallback merge function can be composed with merge.Standard to handle properties with application-specific semantics. The script test shows the composition of the standard merge with a counter-like field that is only ever incremented. Properties which cannot be automatically merged are indicated by a new Conflict.Unmerged field and corresponding userscript binding. This fallback function can also be used to "merge or else" by using a trivial fallback that always returns the name of a dlq. This, too, is demonstrated in the test script.

The merge.Conflict.Existing field is renamed to Target. It either contains the existing state of the row in the target database, or it contains the data that merge.Standard determines should be stored in the target. The change in sort order improves readability in the tests: Before, Proposed, Target -> Expected.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/547)

### DIFF
--- a/internal/script/loader.go
+++ b/internal/script/loader.go
@@ -53,8 +53,9 @@ type mapJS func(
 type mergeOp struct {
 	Before   goja.Value     `goja:"before"`   // Backed by bagWrapper. Nil in 2-way case.
 	Meta     map[string]any `goja:"meta"`     // Equivalent to dispatch() or map() meta.
-	Existing goja.Value     `goja:"existing"` // Backed by bagWrapper.
 	Proposed goja.Value     `goja:"proposed"` // Backed by bagWrapper.
+	Target   goja.Value     `goja:"target"`   // Backed by bagWrapper.
+	Unmerged goja.Value     `goja:"unmerged"` // Intermediate state from standardMerge.
 }
 
 // A mergeResult is returned by the user-provided merge function.
@@ -66,6 +67,14 @@ type mergeResult struct {
 }
 
 type mergeJS func(*mergeOp) (*mergeResult, error)
+
+// These symbols allow bindMerge and standardMerge to collude. We'd
+// like to avoid an unnecessary round-trip through goja's value
+// interface if we can call [merge.Standard] directly.
+var (
+	symIsStandardMerge = goja.NewSymbol("standardMerge") // Boolean flag
+	symMergeFallback   = goja.NewSymbol("fallback")      // Reference to user-provided JS function.
+)
 
 // sourceJS is used in the API binding.
 type sourceJS struct {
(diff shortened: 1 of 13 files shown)


sravotto (approved):
:lgtm:

> Reviewed 10 of 12 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/547#-:-NhRpA1H25F7kl8XpI4T:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @BramGruneir)

bobvawter (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/547)* status: 9 of 13 files reviewed, 1 unresolved discussion (waiting on @BramGruneir and @sravotto)

___
*[`internal/util/ident/map.go` line 53 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/547#-NhgFzEi8oDx_qeNDVN6:-Nhm348K3-KLPNwq4GJu:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f7cd8b28e4f7c99352ff75fce065d72de489bf4d/internal/util/ident/map.go#L53)):*

Worth adding a quick test?
</blockquote></details>

Done.
___
*[`internal/util/merge/standard.go` line 115 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/547#-NhlMRSxESUNrRjCTJNV:-Nhm35Mr1dPnt_-pfrDK:blv43a5) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/f7cd8b28e4f7c99352ff75fce065d72de489bf4d/internal/util/merge/standard.go#L115)):*

This might be cleaner if you added a .GetOrDefault function that returns the value if found or the default if not.  So you could call it as .GetOrDefault(prop, undefined{})

I know I've seen this pattern before, just probably named something better than "orDefault".  Maybe that's not really the golang way.  Up to you.
</blockquote></details>

I'm not a fan of "GetOrDefault" in general.  It can inadvertently lead to wasted allocations.  Consider `GetOrDefault(key, &Foo{})`. There's always going to be a `*Foo` allocated and which will likely escape to the heap.  The way to fix this is to provide a lazy constructor, then your calls look like `GetOrDefault(key, func() *Foo { return &Foo{} })` and I'm honestly not sure that's any nicer than the `if !ok` pattern.

PR #549, opened 2023-10-20
mysql: supporting TLS transport

sravotto:
This change allows the mySQL pool to establish a secure connection with the target server.
Adding a secure utility package to parse postgres-like connection strings and instantiating the proper TLS configuration.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/549)


bobvawter (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/549)* status: 0 of 10 files reviewed, 10 unresolved discussions (waiting on @BramGruneir and @sravotto)

___
*[`internal/util/secure/tls.go` line 36 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/549#-Nhlk1-13iKfbz-0QLCs:-Nhlk1-13iKfbz-0QLCt:bf9w44c) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/a58f25595c50a0fe8376a7312595bbd5d29d1f62/internal/util/secure/tls.go#L36)):*
> ```Go
> // It allows a uniform way to specify connections to various backends
> // following the Postgres connection specifications.
> func ConfigureTransport(url *url.URL) ([]*tls.Config, error) {
> ```

This name implies that the function is active changing something, when it's really like `ParseTLSOptions()`.
___
*[`internal/util/secure/tls.go` line 47 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/549#-NhldAJ5CIpZzSOPOIUB:-NhldAJ5CIpZzSOPOIUC:b-qaa1i8) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/a58f25595c50a0fe8376a7312595bbd5d29d1f62/internal/util/secure/tls.go#L47)):*
> ```Go
> 		return []*tls.Config{nil}, nil
> 	case "allow":
> 		// try insecure first
> ```

It seems like `allow` and `prefer` should have the same behavior.  Why would it be preferable to connect insecurely when TLS is available?
___
*[`internal/util/secure/tls.go` line 121 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/549#-NhleePz2zrYTM68Xj5y:-NhleePz2zrYTM68Xj5z:b-3x734n) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/a58f25595c50a0fe8376a7312595bbd5d29d1f62/internal/util/secure/tls.go#L121)):*
> ```Go
> 	var certPem, keyPem []byte
> 	var err error
> 	if certPath != "" {
> ```

Why test for empty strings here?  It doesn't make sense to call this function without something to load in the first place.  If anything, this should return an error.
___
*[`internal/util/secure/tls.go` line 155 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/549#-NhlexA6EogzfRocZ5IM:-NhlexA6EogzfRocZ5IN:bvf4zx2) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/a58f25595c50a0fe8376a7312595bbd5d29d1f62/internal/util/secure/tls.go#L155)):*
> ```Go
> 			certs[i] = cert
> 		}
> 		// Leave DNSName empty to skip hostname verification.
> ```

On initial inspection, this seems dangerous. Can you expand the comment to say that this is only used in `verify-ca` mode where hostname verification is unwanted?
___
*[`internal/util/stdpool/my.go` line 39 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/549#-Nhlfb9z77Dqza-Rhfid:-Nhlfb9z77Dqza-Rhfie:b-3bx83n) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/a58f25595c50a0fe8376a7312595bbd5d29d1f62/internal/util/stdpool/my.go#L39)):*
> ```Go
> // tlsConfigName is the name of the tls configuration used
> // by the underlying driver.
> const tlsConfigName = "mysql_custom"
> ```

What happens if more than one mysql pool is opened?  Concurrently?  Run the associated test using `t.Parallel()` in the sub-tests and I'll bet you'll see a race condition pop out.  Some of the server integration tests do run in parallel mode and won't catch these race conditions unless the test container is set up to use TLS.
___
*[`internal/util/stdpool/my.go` line 72 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/549#-NhlfswW5zThANMJSJcP:-NhlfswW5zThANMJSJcQ:b-j4cec3) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/a58f25595c50a0fe8376a7312595bbd5d29d1f62/internal/util/stdpool/my.go#L72)):*
> ```Go
> 					return nil, errors.WithStack(err)
> 				}
> 				connector, err := sql.Open("mysql", mySQLString)
> ```

There are two mysql drivers in the package dependencies.  How do you know this returns the one that you want?

It's probably better to use https://pkg.go.dev/github.com/go-sql-driver/mysql#MySQLDriver.OpenConnector and then pass that to https://pkg.go.dev/database/sql#OpenDB
___
*[`internal/util/stdpool/my.go` line 73 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/549#-Nhlh5owC-d7pORPwWlB:-Nhlh5owC-d7pORPwWlC:b-53p6fq) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/a58f25595c50a0fe8376a7312595bbd5d29d1f62/internal/util/stdpool/my.go#L73)):*
> ```Go
> 				}
> 				connector, err := sql.Open("mysql", mySQLString)
> 				if err != nil {
> ```

Add a trace-level log entry in case it ever becomes necessary look at the errors resulting from an attempt. `log.WithError().Trace()`
___
*[`internal/util/stdpool/my.go` line 97 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/549#-NhlhoNh7pym8ieSHaHi:-NhlhoNh7pym8ieSHaHj:bnzvws7) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/e0a66692a7678e40dba7dadbeeb8dd60e145d7f5/internal/util/stdpool/my.go#L97)):*
> ```Go
> 		}
> 		log.Infof("Version %s. Mode %s", ret.Version, mode)
> 		if err := attachOptions(ctx, ret.DB, options); err != nil {
> ```

Keep these `if err :=` scoped assignments.  It's easier to think about error flow when the lexical scope exactly matches the duration of the utility of the variable.
___
*[`internal/util/stdpool/my.go` line 109 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/549#-Nhlj6LzCLjT7iUXzgIF:-Nhlj6LzCLjT7iUXzgIG:b-d07sco) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/a58f25595c50a0fe8376a7312595bbd5d29d1f62/internal/util/stdpool/my.go#L109)):*
> ```Go
> 					continue
> 				}
> 				// we got a connection, making sure that we can actually use it.
> ```

The CRL standard is that comments should be complete sentences with leading initial case.
___
*[`internal/util/stdpool/my.go` line 135 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/549#-Nhlj4C66o5OXcIzyZeG:-Nhlj4C7D2Ny8HckgxE7:bhk3bn1) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/a58f25595c50a0fe8376a7312595bbd5d29d1f62/internal/util/stdpool/my.go#L135)):*
> ```Go
> 				if err == nil {
> 					//Looks like we found a connection,
> 					//no need to try other transport options.
> ```

Rewrap comment.

sravotto (commented):
Please take another look.  I changed the integration test to use tls, since the docker image allows both secure and non-secure transport.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/549#-:-NhnBpyp3YPTA2p34g_d:b1aup8e)* status: 0 of 11 files reviewed, 10 unresolved discussions (waiting on @bobvawter and @BramGruneir)

___
*[`internal/util/secure/tls.go` line 36 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/549#-Nhlk1-13iKfbz-0QLCs:-NhmHlx13OTi1LrtKJpv:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/a58f25595c50a0fe8376a7312595bbd5d29d1f62/internal/util/secure/tls.go#L36)):*

This name implies that the function is active changing something, when it's really like `ParseTLSOptions()`.
</blockquote></details>

Done.
___
*[`internal/util/secure/tls.go` line 47 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/549#-NhldAJ5CIpZzSOPOIUB:-NhlnPyKC6cr6Iqtmlrm:b-hzvh8u) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/a58f25595c50a0fe8376a7312595bbd5d29d1f62/internal/util/secure/tls.go#L47)):*

It seems like `allow` and `prefer` should have the same behavior.  Why would it be preferable to connect insecurely when TLS is available?
</blockquote></details>

Just following the Postgres semantics: https://www.postgresql.org/docs/current/libpq-ssl.html
___
*[`internal/util/secure/tls.go` line 121 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/549#-NhleePz2zrYTM68Xj5y:-NhlnkQ3-RH0FTXb31ox:bqjact5) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/a58f25595c50a0fe8376a7312595bbd5d29d1f62/internal/util/secure/tls.go#L121)):*

Why test for empty strings here?  It doesn't make sense to call this function without something to load in the first place.  If anything, this should return an error.
</blockquote></details>

key pair is optional, but refactored the code to be more readable.
___
*[`internal/util/stdpool/my.go` line 39 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/549#-Nhlfb9z77Dqza-Rhfid:-NhnBO1J5wO4VwS_RknU:b-ipsfz9) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/a58f25595c50a0fe8376a7312595bbd5d29d1f62/internal/util/stdpool/my.go#L39)):*

What happens if more than one mysql pool is opened?  Concurrently?  Run the associated test using `t.Parallel()` in the sub-tests and I'll bet you'll see a race condition pop out.  Some of the server integration tests do run in parallel mode and won't catch these race conditions unless the test container is set up to use TLS.
</blockquote></details>

Let me know if this is what meant.
___
*[`internal/util/stdpool/my.go` line 72 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/549#-NhlfswW5zThANMJSJcP:-NhnBVm-BtYOIx7Njlcx:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/a58f25595c50a0fe8376a7312595bbd5d29d1f62/internal/util/stdpool/my.go#L72)):*

There are two mysql drivers in the package dependencies.  How do you know this returns the one that you want?

It's probably better to use https://pkg.go.dev/github.com/go-sql-driver/mysql#MySQLDriver.OpenConnector and then pass that to https://pkg.go.dev/database/sql#OpenDB
</blockquote></details>

Done.
___
*[`internal/util/stdpool/my.go` line 73 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/549#-Nhlh5owC-d7pORPwWlB:-NhnBWDeCiid6AH5VLrr:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/a58f25595c50a0fe8376a7312595bbd5d29d1f62/internal/util/stdpool/my.go#L73)):*

Add a trace-level log entry in case it ever becomes necessary look at the errors resulting from an attempt. `log.WithError().Trace()`
</blockquote></details>

Done.
___
*[`internal/util/stdpool/my.go` line 109 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/549#-Nhlj6LzCLjT7iUXzgIF:-NhnBa6g2cYFrq2tFPmp:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/a58f25595c50a0fe8376a7312595bbd5d29d1f62/internal/util/stdpool/my.go#L109)):*

The CRL standard is that comments should be complete sentences with leading initial case.
</blockquote></details>

Done.

bobvawter (approved):
LGTM w/ nits.

> Reviewed 6 of 10 files at r1, 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/549#-:-Ni-pk5XDZcL7sfdl36Y:b-ott0jx)* status: all files reviewed, 2 unresolved discussions (waiting on @BramGruneir and @sravotto)

___
*[`.github/workflows/go-tests.yaml` line 149 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/549#-Ni-p_SmBB2e7hAOf9CJ:-Ni-p_SmBB2e7hAOf9CK:blnaevd) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/185c4bd379ee20f72fe703522c7bc3d815212202/.github/workflows/go-tests.yaml#L149)):*
> ```YAML
>           - cockroachdb: v23.1
>             target: mysql-mariadb-v10
>             targetConn: "mysql://root:SoupOrSecret@127.0.0.1:3306/mysql"
> ```

Is it also worth testing TLS with MariaDB?
___
*[`internal/util/secure/tls.go` line 121 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/549#-NhleePz2zrYTM68Xj5y:-Ni-fr_yCZfSWT0CoLUt:b-w1xyjt) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/a58f25595c50a0fe8376a7312595bbd5d29d1f62/internal/util/secure/tls.go#L121)):*

key pair is optional, but refactored the code to be more readable.
</blockquote></details>

This `if` statement seems suspicious.  If you called this function with two empty strings, the call to `os.ReadFile()` is going to error out.  I think you want two checks to require that both `certPath` and `keyPath` are non-empty.
___
*[`internal/util/stdpool/my.go` line 39 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/549#-Nhlfb9z77Dqza-Rhfid:-Ni-l-UZ35DnBGLavyIK:b-8fx2u7) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/a58f25595c50a0fe8376a7312595bbd5d29d1f62/internal/util/stdpool/my.go#L39)):*

Let me know if this is what meant.
</blockquote></details>

:+1:

sravotto (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/549)* status: 10 of 11 files reviewed, 2 unresolved discussions (waiting on @bobvawter and @BramGruneir)

___
*[`.github/workflows/go-tests.yaml` line 149 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/549#-Ni-p_SmBB2e7hAOf9CJ:-Ni0lWI1AXLsMaSaIwED:b-37wvl4) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/185c4bd379ee20f72fe703522c7bc3d815212202/.github/workflows/go-tests.yaml#L149)):*

Is it also worth testing TLS with MariaDB?
</blockquote></details>

we could, however we would need to enable TLS on the server side (with mysql it's on by default). I did run a local test.
___
*[`internal/util/secure/tls.go` line 121 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/549#-NhleePz2zrYTM68Xj5y:-Ni0hZA-1iywL7B1-7Eb:b-j4xhlr) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/a58f25595c50a0fe8376a7312595bbd5d29d1f62/internal/util/secure/tls.go#L121)):*

This `if` statement seems suspicious.  If you called this function with two empty strings, the call to `os.ReadFile()` is going to error out.  I think you want two checks to require that both `certPath` and `keyPath` are non-empty.
</blockquote></details>

No - if the 2 are empty it will correctly return `nil, nil`. I'm going to flip the 2nd if to make it simpler.

PR #566, opened 2023-10-30
script: Fix race condition and missing Before in UserScript

bobvawter:
This change fixes two bugs in the script package. Firstly, if the `script.Evaluate()` function is called more than once (as it is when multiple logical loops are started), the UserScript instances that are created share a common goja.Runtime object, but have separate mutexes to guard access to the runtime. This changes the script loader (which constructs the Runtime) to also allocate a shared mutex which is passed to the derived UserScript instance(s).

There is also a small fix to `bindMap()` where a mutation's Before field was not being propagated. A test is added to verify the fix.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/566)

### DIFF
--- a/internal/script/loader.go
+++ b/internal/script/loader.go
@@ -24,6 +24,7 @@ import (
 	"net/url"
 	"path"
 	"strings"
+	"sync"
 
 	"github.com/dop251/goja"
 	esbuild "github.com/evanw/esbuild/pkg/api"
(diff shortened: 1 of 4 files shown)


BramGruneir (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/566#-:-Ni0TA-SAfpBYlj1n7JM:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sravotto)

Issue #565, opened 2023-10-27
Option to track re-applied mutations in the target


bobvawter:
We may want to add an option in future support of #487 and #504 to manage a manifest of applied mutations within the target database. By storing `(table, key, timestamp)` we could suppress any replayed mutations in use cases where an absolute guarantee of exactly-once behavior is required.

ryanluu12345:
I thought in the staging table we already track the `applied` state for each of these mutations. What would be different about this case?

Issue #504, opened 2023-09-27
RFC: Backfill, Immediate, and Initial-Fill can be unified


bobvawter:
In the present configuration, we have three major modes of operation:
* fully-transactional
* backfill
* immediate

Even with these modes, we don't have a great way to handle the initial-backfill case that doesn't involve user intervention to switch from immediate to transactional mode.  There are cases where one would like to use immediate mode, which are further thwarted by immediate mode's inability to support foreign-key relationships in the target schema.

Below is a proposal that might allow cdc-sink to have exactly two modes, transactional and non-transactional, that also manages to work with foreign-key dependencies. It should also be possible to implement this mode through the existing `logical.Dialect` API to enable temporary reductions in transactional consistency and, perhaps, eliminate the `logical.Backfiller` capability API altogether.

Transactional mode behaves as it does today: Data is staged in MVCC order and resolved timestamps from the source identify when we can safely apply staged data in a manner that is transactionally-consistent with respect to the source.

In non-transactional mode, when the CDC handler (or any other user of `logical.Batcher`) is presented with a mutation, the pipeline attempts to apply the mutation to the target table. This handles the no-FK case and the FK case where the parent key already exists. If a row cannot be applied to the target table due to a missing parent, the row will be written to the relevant staging table, which acts as a dead-letter queue.  Knowing that there's a DLQ, we'll revise the previous rule to read: "... it attempts to apply the mutation _if no un-applied data exists for the target key, otherwise the mutation is added to the DLQ_."  This happens within a single staging-database transaction to ensure that all received mutations have been stored somewhere before the `200 OK` is returned.

The above can be optimized further if we extend the schema introspection code to extract the columns that defines FK relationships. Mutations to child tables can proactively apply any mutations to their parent row; the fact that a mutation to a child row exists implies the existence of a parent row.  This on-demand behavior is applied recursively.  It would be infeasible to have parent mutations proactively look for child mutations which are unblocked without a mutation-indexing strategy.  Rather, we will have a background process that iterates over staging tables in their dependency order to look for mutations that can be applied.

If the `--retireOffset` value is zero, immediately-applied mutations need never be written to the staging table.  Similarly, demand-applied mutations can be immediately deleted from the staging table.  If `--retireOffset` is non-zero, the staging table will require an extra column to indicate that the mutation has been applied to a target table and should be ignored in future scans.

[...]

PR #572, opened 2023-10-30
stage: Track applied mutations in staging tables

bobvawter:
This change moves tracking of partially-applied resolved timestamp windows into
the staging tables by adding a new `applied` column. The goal of this change is
to move some state-tracking out of the cdc resolver loop into the stage
package. Tracking apply status on a per-mutation basis improves idempotency of
cdc-sink when the userscript has non-idempotent behaviors (e.g.: three-way
merge). It also allows us to export monitoring data around mutations which may
have slipped through the cracks or to detect when a migration process has
completely drained. Fine-grained tracking will also be useful for unifying the
non-transactional modes into a single behavior.

Many unused methods in the stage API have been deleted. The "unstaging" SQL
query is now generated with a golang template and is tested similarly to the
apply package.

The cdc package performs less work to track partial application of large
individual changes. It just persists the contents of the UnstageCursor as a
performance enhancement. Exactly-once behavior is provided by the applied
column.

The change to `server/integration_test.go` is due to the unstage processing
being a one-shot. The test being performed duplicates an existing test in
`cdc/handler_test.go`.

Breaking change: The `--selectBatchSize` flag is deprecated in favor of two
different flags `--largeTransactionLimit` and `--timestampWindowSize` which,
respectively, enable partial processing of a single, over-sized transaction and
a general limit on the total amount of data to be unstaged.

Breaking change: A staging schema migraion is required, this is documented in
the migrations directory.

X-Ref: #487
X-Ref: #504
X-Ref: #565

This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/572)

### DIFF
--- a/internal/sinktest/all/fixture.go
+++ b/internal/sinktest/all/fixture.go
@@ -26,7 +26,9 @@ import (
 	"github.com/cockroachdb/cdc-sink/internal/types"
 	"github.com/cockroachdb/cdc-sink/internal/util/applycfg"
 	"github.com/cockroachdb/cdc-sink/internal/util/diag"
+	"github.com/cockroachdb/cdc-sink/internal/util/hlc"
 	"github.com/cockroachdb/cdc-sink/internal/util/ident"
+	"github.com/jackc/pgx/v5"
 	"github.com/pkg/errors"
 )
 
@@ -75,4 +77,34 @@ func (f *Fixture) CreateTargetTable(
 		err = f.Watcher.Refresh(ctx, f.TargetPool)
 	}
 	return ti, err
+}
+
+// PeekStaged peeks at the data which has been staged for the target
+// table between the given timestamps.
+func (f *Fixture) PeekStaged(
+	ctx context.Context, tbl ident.Table, startAt, endBefore hlc.Time,
+) ([]types.Mutation, error) {
+	tx, err := f.StagingPool.BeginTx(ctx, pgx.TxOptions{})
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+	defer func() { _ = tx.Rollback(ctx) }()
+
+	q := &types.UnstageCursor{
+		StartAt:   startAt,
+		EndBefore: endBefore,
+		Targets:   []ident.Table{tbl},
+	}
+	var ret []types.Mutation
+	for selecting := true; selecting; {
+		q, selecting, err = f.Stagers.Unstage(ctx, tx, q,
+			func(ctx context.Context, tbl ident.Table, mut types.Mutation) error {
+				ret = append(ret, mut)
+				return nil
+			})
+		if err != nil {
+			return nil, err
+		}
+	}
+	return ret, nil
 }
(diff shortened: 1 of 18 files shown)


bobvawter:
I had to restack this PR on top of #575 so that the goroutine to report on dangling mutations in the staging table could be run in a context that is both durable and able to be shut down cleanly.

bobvawter:
Fixed after rebasing, ready for review.

bobvawter:
PTAL (part deux), after rebasing onto merged #577

BramGruneir (approved):
:lgtm:

The logic seems sound.  I couldn't find anything that stood out as incorrect.  

> Reviewed 13 of 18 files at r2, 36 of 36 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/572#-:-NiFcZ8c44nb-Ir_BCh-:b-fy1mkz)* status: all files reviewed, 3 unresolved discussions (waiting on @bobvawter and @sravotto)

___
*[`internal/source/cdc/config.go` line 33 at r3](https://reviewable.io/reviews/cockroachdb/cdc-sink/572#-NiFTpIN-5Re01CVQfYw:-NiFTpIN-5Re01CVQfYx:b6q8037) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/6d94bf3a3c0ef18d2244833be4e50a1429b614d2/internal/source/cdc/config.go#L33)):*
> ```Go
> 	defaultIdealBatchSize        = 1000
> 	defaultMetaTable             = "resolved_timestamps"
> 	defaultLargeTransactionLimit = 250_000                // Chosen arbitrarily.
> ```

Should this default be 0?
___
*[`internal/source/cdc/resolver.go` line 383 at r3](https://reviewable.io/reviews/cockroachdb/cdc-sink/572#-NiFVWGIBmYRsTCeYAO0:-NiFVWGJ1cwP9i3YqU_1:b-m68j3z) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/6d94bf3a3c0ef18d2244833be4e50a1429b614d2/internal/source/cdc/resolver.go#L383)):*
> ```Go
> 
> 		retries := -1
> 		if err := retry.Retry(ctx, func(ctx context.Context) error {
> ```

Could you move the retry counting into the retry.Retry?  Just have the Retry return retryCount, err

It might be worth having the number of retries attempted, even if it fails.
___
*[`internal/source/cdc/resolver.go` line 429 at r3](https://reviewable.io/reviews/cockroachdb/cdc-sink/572#-NiFVIcV4l3JPLaslrS8:-NiFVIcV4l3JPLaslrS9:b-nvmtb) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/6d94bf3a3c0ef18d2244833be4e50a1429b614d2/internal/source/cdc/resolver.go#L429)):*
> ```Go
> 	}
> 
> 	total := 0 // Metrics.
> ```

Metrics?

Issue #560, opened 2023-10-24
logical_commit_latency_seconds doesn't tick


bobvawter:
This metric is only updated when a resolved timestamp is processed.  If resolved timestamps aren't being received or processing is blocked for any reason, this metric does not update.  This metric should be updated by an independent goroutine that monitors e.g. a `notify.Var`.

bobvawter:
The information that we want to display really comes from the `*cdc.resolver`. PR #579 will add a new metric `cdc_resolver_committed_age_seconds` that ticks once per second.

PR #575, opened 2023-10-31
wire: Make stopper.Context globally available

bobvawter:
Review note: Start with the top-level `main.go` to see how we can now explicitly control graceful shutdown independently of tearing down the entire stack.  Several types like `cdc.Resolvers` no longer need to manage the lifecyle of things that they create, since it can be delegated to the stack-global stopper.

This change cleans up some longstanding lifecycle issues around graceful draining. In the past, we relied on the ability to return a cancel function to Wire to shut down background goroutines that might be started by some service or another. This works, but it would be nice to standardize on the stopper.Context utility type for running goroutines that can benefit from a soft-exit condition.

The crux of this change is to make `*stopper.Context` available from all injectors. The `context.Context` type remains injectable, but it becomes an alias for the `*stopper.Context`. The top-level injectors now require a `*stopper.Context`, which gives the callers the ability to trigger a soft shutdown of background processes before tearing down the entire stack. This should reduce the amount of log-spam during shutdown.

This change will also make it far more straightforward to implement metrics that tick (e.g. #560) since a service or object that needs to produce a ticking metric has the option to use the stack-global `*stopper.Context.Go()`. Increase use of our `Go()` method also provides a future opportunity to improve observabiity around background processes within a cdc-sink binary.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/575)

### DIFF
--- a/internal/cmd/fslogical/fslogical.go
+++ b/internal/cmd/fslogical/fslogical.go
@@ -21,6 +21,7 @@ package fslogical
 import (
 	"github.com/cockroachdb/cdc-sink/internal/source/fslogical"
 	"github.com/cockroachdb/cdc-sink/internal/util/stdlogical"
+	"github.com/cockroachdb/cdc-sink/internal/util/stopper"
 	"github.com/spf13/cobra"
 )
 
@@ -31,7 +32,8 @@ func Command() *cobra.Command {
 		Bind:  cfg.Bind,
 		Short: "start a Google Cloud Firestore logical replication feed",
 		Start: func(cmd *cobra.Command) (any, func(), error) {
-			return fslogical.Start(cmd.Context(), cfg)
+			// main.go provides this stopper.
+			return fslogical.Start(stopper.From(cmd.Context()), cfg)
 		},
 		Use: "fslogical",
 	})
--- a/internal/cmd/mylogical/mylogical.go
+++ b/internal/cmd/mylogical/mylogical.go
@@ -21,6 +21,7 @@ package mylogical
 import (
 	"github.com/cockroachdb/cdc-sink/internal/source/mylogical"
 	"github.com/cockroachdb/cdc-sink/internal/util/stdlogical"
+	"github.com/cockroachdb/cdc-sink/internal/util/stopper"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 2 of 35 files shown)


BramGruneir (approved):
This is a lot cleaner a shutdown, nice to see.

:lgtm:

> Reviewed 35 of 35 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/575#-:-Ni9yKK9E-JUE4597CqN:b-m6uvul)* status: all files reviewed, 1 unresolved discussion (waiting on @bobvawter and @sravotto)

___
*[`internal/staging/stage/factory.go` line 49 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/575#-Ni9xhDSDHq_L4VPwZn2:-Ni9xhDT847D8yFjihYF:b-vpzh2x) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/c16a4b65b9d23725c2540ac355beb90ca0725167/internal/staging/stage/factory.go#L49)):*
> ```Go
> 
> // Get returns a memoized instance of a stage for the given table.
> func (f *factory) Get(_ context.Context, target ident.Table) (types.Stager, error) {
> ```

Is this context required to conform to Stagers?  Otherwise, you can just remove it.

bobvawter (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/575)* status: all files reviewed, 1 unresolved discussion (waiting on @BramGruneir and @sravotto)

___
*[`internal/staging/stage/factory.go` line 49 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/575#-Ni9xhDSDHq_L4VPwZn2:-NiA-LRFBXMnkWpf9YCx:b-bb1pbj) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/c16a4b65b9d23725c2540ac355beb90ca0725167/internal/staging/stage/factory.go#L49)):*

Is this context required to conform to Stagers?  Otherwise, you can just remove it.
</blockquote></details>

The context is part of the `Stagers` interface.  I'd rather keep passing contexts around just in case we wind up needing it in the future.

PR #577, opened 2023-11-01
wire: Remove all cleanup functions

bobvawter:
The cleanup function in Wire is quite useful, but it's not the right pattern when we have background processes in cdc-sink. It's often the case that, in tests, we tear down the database pool before some goroutine has had a chance to stop. This creates log-spam in the tests, which is also sometimes seen when shutting down a real cdc-sink process.

The existing `stopper.Context` gains a new `Defer()` method. Like its golang namesake, `Defer()` accepts a function to be called in LIFO order after the stopper has stopped. This means that cleanup of, say, a database pool can be deferred until after any goroutines that use it have exited or been cancelled.

The remainder of the changes are essentially mechanical:
* Every top-level Wire injector function that used to return a cancel function now only return the constructed object and an error.
* The Wire code-generator complains about any provider method that returns a cancel function, since there would be no way to call it.
* Provider functions that previously had a cancel behavior are changed to accept a `*stopper.Context` and to use the new `Defer()` method.
* All cases where a long-lived, background goroutine is started must now use `stopper.Context.Go()`, which coordinates the lifecycle of the goroutine versus the services that it uses in the injector graph.
* Factory types that accumulated cleanup functions for long-lived service objects delegate cleanup to the top-level stopper. This naturally falls out of the previous point.
* The test fixture injector methods `all.NewFixture()` and `base.NewFixture()` require the associated `*testing.T` and register the top-level stopper with the `TB.Cleanup()` method.

The net result is that, for any given injector graph, a consistent way to tear down long-lived objects and services in a wholly coordinated fashion.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/577)

### DIFF
--- a/internal/cmd/fslogical/fslogical.go
+++ b/internal/cmd/fslogical/fslogical.go
@@ -31,9 +31,9 @@ func Command() *cobra.Command {
 	return stdlogical.New(&stdlogical.Template{
 		Bind:  cfg.Bind,
 		Short: "start a Google Cloud Firestore logical replication feed",
-		Start: func(cmd *cobra.Command) (any, func(), error) {
+		Start: func(ctx *stopper.Context, cmd *cobra.Command) (any, error) {
 			// main.go provides this stopper.
-			return fslogical.Start(stopper.From(cmd.Context()), cfg)
+			return fslogical.Start(ctx, cfg)
 		},
 		Use: "fslogical",
 	})
--- a/internal/cmd/mylogical/mylogical.go
+++ b/internal/cmd/mylogical/mylogical.go
@@ -31,9 +31,8 @@ func Command() *cobra.Command {
 	return stdlogical.New(&stdlogical.Template{
 		Bind:  cfg.Bind,
 		Short: "start a mySQL replication feed",
-		Start: func(cmd *cobra.Command) (any, func(), error) {
-			// main.go provides a stopper.
-			return mylogical.Start(stopper.From(cmd.Context()), cfg)
+		Start: func(ctx *stopper.Context, cmd *cobra.Command) (any, error) {
+			return mylogical.Start(ctx, cfg)
 		},
 		Use: "mylogical",
 	})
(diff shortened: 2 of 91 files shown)


BramGruneir (commented):
So much code cleanup!  And it's easier to reason about.  

:lgtm: if you can get the tests passing

> Reviewed 90 of 90 files at r1, 5 of 8 files at r2, 2 of 3 files at r3, 6 of 6 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/577#-:-NiBc5-jCqhw1Jjpw5tM:b-ny4r0c)* status: all files reviewed, 1 unresolved discussion (waiting on @bobvawter)

___
*[`internal/util/stopper/stopper.go` line 304 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/577#-NiB_ciR1z7lK4fjBWRy:-NiB_ciR1z7lK4fjBWRz:b92oysm) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/2a8c2390297420dff70b53eb85ccddb5a6df6e7a/internal/util/stopper/stopper.go#L304)):*
> ```Go
> // cancelLocked invokes the context-cancellation function and then
> // executes any deferred callbacks.
> func (s *Context) cancelLocked(err error) {
> ```

Why is this a separate func?

bobvawter (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/577)* status: 87 of 91 files reviewed, 1 unresolved discussion (waiting on @BramGruneir)

___
*[`internal/util/stopper/stopper.go` line 304 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/577#-NiB_ciR1z7lK4fjBWRy:-NiBjfBq4ygT3qiWnivR:b-s3ff6l) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/2a8c2390297420dff70b53eb85ccddb5a6df6e7a/internal/util/stopper/stopper.go#L304)):*

Why is this a separate func?
</blockquote></details>

It's called from a couple of places.

PR #579, opened 2023-11-02
cdc: Add metrics to resolver

bobvawter:
We have been monitoring the progress of a CDC feed via the metrics exported from the logical package, but that doesn't have visibility into the state of the CDC resolver. For instance the `logical_commit_latency_seconds` metric can only tell an observer about the latest timestamp that was committed and not about the one being processed. There are also some performance counters that will help to identify where, in the logical+cdc stack, time is being spent.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/579)

### DIFF
--- a/internal/source/cdc/metrics.go
+++ b/internal/source/cdc/metrics.go
@@ -1,0 +1,77 @@
+// Copyright 2023 The Cockroach Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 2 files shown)


BramGruneir (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/579#-:-NiGKOXeBVikDnXLKrK5:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @glennfawcett and @sravotto)

PR #582, opened 2023-11-03
pglogical: Add fix and test for UDT enum types

### LINKED ISSUE #578: Verify pglogical vs UDT in source schema
May need to process TypeMessages in `pglogical/conn.go` to unpack UDTs in the input stream.

bobvawter:
This change adds a test for an enum type in the source and target databases. When a UDT is used, a TypeMessage will be sent over the pglogical feed. We can ignore this message, since we interpret incoming data via the target typesystem.

Fixes #578

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/582)


BramGruneir:
Lgtm

PR #600, opened 2023-11-20
support for a mutation specific custom upsert template.

sravotto:
This change allows source dialect to request custom upsert templates for specific mutations. The logical apply loop will use the custom template if available.

For instance, this is used to handle Postgres unchanged toasted columns. If a custom colum is detected, then a custom upsert template is used to query the current value of the toasted column in target database before inserting the mutation.

This change also adds a new configuration flag '--enableToastedColumns' which allows cdc-sink to support toasted columns when using Postres as a source.

Support is limited for logical replication with basic upserts into CRDB targets via a custom template (no user scripts or merge functionality).

When the experimental support for toasted columns is enabled, and cdc-sink receives a TupleDataTypeToast, it marks the column with a placeholder.

The placeholder is recognized by the logical apply loop which uses the custom upsert template for any mutations with toasted columns. The templates retrieves the current value for the toasted column in the target database before applying the mutation.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/600)

### DIFF
--- a/internal/source/pglogical/config.go
+++ b/internal/source/pglogical/config.go
@@ -40,6 +40,8 @@ type Config struct {
 	Slot string
 	// Connection string for the source db.
 	SourceConn string
+	// Enable support for toasted columns
+	ToastedColumns bool
 }
 
 // Bind adds flags to the set.
@@ -55,6 +57,8 @@ func (c *Config) Bind(f *pflag.FlagSet) {
 	f.StringVar(&c.SourceConn, "sourceConn", "", "the source database's connection string")
 	f.StringVar(&c.Publication, "publicationName", "",
 		"the publication within the source database to replicate")
+	f.BoolVar(&c.ToastedColumns, "enableToastedColumns", false,
+		"Enable support for toasted columns")
 }
 
 // Preflight updates the configuration with sane defaults or returns an
--- a/internal/source/pglogical/conn.go
+++ b/internal/source/pglogical/conn.go
@@ -69,6 +69,8 @@ type conn struct {
 	slotName string
 	// The configuration for opening replication connections.
 	sourceConfig *pgconn.Config
+	// Support for toasted columns
+	toastedColumns bool
 }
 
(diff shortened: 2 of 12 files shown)


BramGruneir (commented):
So this looks good.  but I'm not convinced the generic template makes sense.  We're going to need to support toasted columns moving forward, and I like the sql to support them.  

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/600#-:-NjhhdoI4s64ySSg60GF:b-cns742)* status: all files reviewed, 2 unresolved discussions (waiting on @sravotto)

___
*[`internal/source/pglogical/config.go` line 44 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/600#-NjhhLhQ6k1g9lohSwky:-NjhaJ8O7wAzrX5w-YsR:b33qoxn) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/2c77ad1bcdc74d5d1f208dccfb34f9af500d78d8/internal/source/pglogical/config.go#L44)):*
> ```Go
> 	SourceConn string
> 	// Enable experimental support for toasted columns
> 	ExperimentalToastedColumns bool
> ```

Please remove Experimental.
___
*[`internal/source/pglogical/provider.go` line 98 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/600#-Njhakqf-6fqGO_ms-qr:-Njhakqf-6fqGO_ms-qs:b-oyyzyo) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/2c77ad1bcdc74d5d1f208dccfb34f9af500d78d8/internal/source/pglogical/provider.go#L98)):*
> ```Go
> 		slotName:                   config.Slot,
> 		sourceConfig:               sourceConfig,
> 		experimentalToastedColumns: config.ExperimentalToastedColumns,
> ```

Remove the experimental everywhere.

PR #601, opened 2023-11-20
pglogical: reworking toasted columns support.

sravotto:
This change adds a new configuration flag '--enableToastedColumns' which allows cdc-sink to support toasted columns when using Postres as a source.

Support is limited for logical replication with basic upserts into CRDB targets (no user scripts or merge functionality).

When the experimental support for toasted columns is enabled, and cdc-sink receives a TupleDataTypeToast, it marks the column with a placeholder.

The placeholder is recognized by the logical apply loop which uses a special upsert template for any mutations with toasted columns. The templates retrieves the current value for the toasted column in the target database before applying the mutation.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/601)

### DIFF
--- a/internal/source/pglogical/config.go
+++ b/internal/source/pglogical/config.go
@@ -40,6 +40,8 @@ type Config struct {
 	Slot string
 	// Connection string for the source db.
 	SourceConn string
+	// Enable experimental support for toasted columns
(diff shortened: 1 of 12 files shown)


sravotto:
Closed in favor of #600

PR #616, opened 2023-12-07
script: Allow userscript to execute SQL

bobvawter:
This change allows the userscript to replace the default upsert and delete
behaviors in the apply package by supplying JS functions via
`api.configureTable()`. These functions are provided with a facade around a
target database transaction and the documents to be upserted or the primary
keys to be deleted.

Database access is asynchronous from the perspective of the userscript. The
database API is modeled using Promises, which allows us to release the
exclusive lock on the JS runtime while the database commands are executed in
another goroutine.  Results of database queries are provided using the
iteration protocal that allows the `for..of` syntax and which can respond to a
`break` statement by releasing the underlying database resources. The
background databases accesses are performed using a limited-size goroutine
pool.

The `UserScript.execJS()` method now provides a signal, via a `notify.Var` that
JS code has finished executing. This allows the new `UserScript.await()` method
to see if its Promise argument has resolved. The JS runtime is left in an
interrupted state with a more useful error message to indicate coding errors
where `execJS()` was not used.

The ESBuild loader configuration now targets ES2022 since our newer goja
dependency natively supports promises and other syntaxes that the original
dependency did not.

Review notes:
* Start with the changes to the `.d.ts` and `main.ts` files. This will show the API being implemented.
* The `applycfg.ApplyDelegate` override may also be useful for simplifying tests that want to verify other transformations in the apply pipeline.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/616)

### DIFF
--- a/internal/script/applier.go
+++ b/internal/script/applier.go
@@ -1,0 +1,304 @@
+// Copyright 2023 The Cockroach Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package script
+
+import (
+	"context"
+	"database/sql"
+	"sync"
+
+	"github.com/cockroachdb/cdc-sink/internal/types"
+	"github.com/cockroachdb/cdc-sink/internal/util/ident"
+	"github.com/cockroachdb/cdc-sink/internal/util/pjson"
+	"github.com/dop251/goja"
+	"github.com/pkg/errors"
+)
+
+// A JS function which is provided with an array of applyOp.
+type applyJS func(opData []*applyOp) *goja.Promise
+
+type applyOp struct {
+	Action string         `goja:"action"` // Always populated. Type must be string.
+	Data   map[string]any `goja:"data"`   // Present in upsert mode.
+	Meta   map[string]any `goja:"meta"`   // Present in upsert mode. Equivalent to map() meta.
+	PK     []any          `goja:"pk"`     // Always populated.
+}
+
+type applyAction string
+
+const (
+	actionDelete applyAction = "delete"
+	actionUpsert applyAction = "upsert"
+)
+
+// notInTransaction is used to provide a helpful error message if
+// api.getTX() is called when no transaction is available.
+func notInTransaction() error {
+	return errors.New("no transaction is currently open")
+}
+
+// applier implements [types.Applier] to allow user-defined functions to
+// be used to interact with the database, rather than using cdc-sink's
+// built-in SQL.
+type applier struct {
+	apply  applyJS
+	parent *UserScript
+	table  ident.Table
+}
+
+var _ types.Applier = (*applier)(nil)
+
+func newApplier(parent *UserScript, table ident.Table, apply applyJS) *applier {
+	return &applier{
+		apply:  apply,
+		parent: parent,
+		table:  table,
+	}
+}
+
+// Apply implements [types.Applier].
+func (a *applier) Apply(ctx context.Context, tq types.TargetQuerier, muts []types.Mutation) error {
+	ops := make([]*applyOp, len(muts))
+	pks := make([]*[]any, len(muts))
+	data := make([]*map[string]any, len(muts))
+	for idx, mut := range muts {
+		mode := actionUpsert
+		if mut.IsDelete() {
+			mode = actionDelete
+		}
+		ops[idx] = &applyOp{
+			Action: string(mode),
+			Meta:   mut.Meta,
+		}
+		pks[idx] = &ops[idx].PK
+		data[idx] = &ops[idx].Data
+	}
+
+	if err := pjson.Decode(ctx, pks, func(i int) []byte {
+		return muts[i].Key
+	}); err != nil {
+		return err
+	}
+
+	if err := pjson.Decode(ctx, data, func(i int) []byte {
+		if data := muts[i].Data; len(data) > 0 {
+			return data
+		}
+		return []byte("null")
+	}); err != nil {
+		return err
+	}
+
+	tx := &targetTX{
+		ctx:     ctx,
+		applier: a,
+		tq:      tq,
+	}
+
+	var promise *goja.Promise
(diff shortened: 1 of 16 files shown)


BramGruneir (approved):
:lgtm:

But I think you probably need to add a number of new test cases that do weird things with the userscript.
- one that writes to two tables
- one that performs a delete than insert (I'm thinking redshift here)

> Reviewed 16 of 16 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/616#-:-NlPAyv32OVZExp4E5p3:b-wnzo3j)* status: all files reviewed, 1 unresolved discussion (waiting on @bobvawter and @sravotto)

___
*[`internal/script/applier.go` line 89 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/616#-NlP9c2N-bbLenoCBJS9:-NlP9c2N-bbLenoCBJSA:b-4lg7bw) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/bc23df0816bda54741c084bff516e79937ed8938/internal/script/applier.go#L89)):*
> ```Go
> 	}
> 
> 	if err := d.callDelete(ctx, tx, deleteData); err != nil {
> ```

Anyway to call these at the same time?

Or, even worse, what happens if you get multiple deletes and upserts for the same pks?
___
*[`internal/script/applier.go` line 170 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/616#-NlPACV_9fcaY1pshpEY:-NlPACVa9WKDfq-GnmOI:brzblql) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/bc23df0816bda54741c084bff516e79937ed8938/internal/script/applier.go#L170)):*
> ```Go
> 		_, err := tx.tq.ExecContext(tx.ctx, q, args...)
> 		tx.mu.Unlock()
> 		err = errors.Wrap(err, q)
> ```

Can you add a comment here on how this err is returned.  I found it a little hard to follow.

-- I see you use this pattern a few times.  

If you wrap a nil error, just it stay nil?
___
*[`internal/script/script_test.go` line 242 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/616#-NlP8QWX6_b8VVusIRoD:-NlP8QWX6_b8VVusIRoE:b2vedxb) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/bc23df0816bda54741c084bff516e79937ed8938/internal/script/script_test.go#L242)):*
> ```Go
> 	t.Run("sql_test", func(t *testing.T) {
> 		if fixture.TargetPool.Product != types.ProductCockroachDB {
> 			t.Skip("user script has opinionated SQL")
> ```

Opinionated?  :)

bobvawter (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/616)* status: all files reviewed, 1 unresolved discussion (waiting on @BramGruneir and @sravotto)

___
*[`internal/script/applier.go` line 89 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/616#-NlP9c2N-bbLenoCBJS9:-NlPd9A92VpE4P2fA8SH:b-spbnho) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/bc23df0816bda54741c084bff516e79937ed8938/internal/script/applier.go#L89)):*

Anyway to call these at the same time?

Or, even worse, what happens if you get multiple deletes and upserts for the same pks?
</blockquote></details>

The regular apply code calls [`msort.UniqueByKey()`](https://github.com/cockroachdb/cdc-sink/blob/master/internal/util/msort/msort.go#L38) to avoid duplicate updates to the same key within a single batch of updates.  I think you could make a reasonable argument, either way, as to wether or not this should be applied to the mutations passed to the script runtime.

Rather than defining a separate `delete` and `upsert` function, the API could be changed to receive a data type that looks like `( pk ,  hlc , operation, data)`.  The operation would be an enumerated value of `"upsert"` or `"delete"`.  This way, the callback would have access to the entire set of mutations at once.
___
*[`internal/script/applier.go` line 170 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/616#-NlPACV_9fcaY1pshpEY:-NlPljY72ytQmArgMMCq:b7lzmzb) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/bc23df0816bda54741c084bff516e79937ed8938/internal/script/applier.go#L170)):*

Can you add a comment here on how this err is returned.  I found it a little hard to follow.

-- I see you use this pattern a few times.  

If you wrap a nil error, just it stay nil?
</blockquote></details>

The errors package documents that any of the wrapper functions will return `nil` if a `nil` error is passed in.

bobvawter (commented):
PTAL. I changed the API so that there's only a single `apply` callback instead of the two methods.  The `ApplyOp` is a union type that indicates if it's a delete or an update, so the unified `apply` can inspect all operations to be performed on any given table.  I'm also taking advantage of a Goja feature that allows asynchronous promise chains to have a context-like value automatically pushed into the runtime. This allows me to have an `api.getTX()` function that's safe to call in a `Promise.then()` or in an `async` function instead of passing the transaction as an argument to the `apply` function.

Given the current API design where the apply package operates on a single table at a time, it would require additional internal work to hoist the JS `apply` function from being single-table to being multi-table.  That behavior would have to be hoisted up to the resolver loop.  We could also consider some kind of continuation-passing mechanism where the `apply` callback is called for each table, but there's another function in the api which allows it to safely accumulate into some object (re-using the async tracker) and we add a `preCommit()` callback to operate on the accumulated result.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/616#-:-NlUej3PAIszVx0JJ6FM:b-yok4uo)* status: 9 of 16 files reviewed, 1 unresolved discussion (waiting on @BramGruneir and @sravotto)

BramGruneir (approved):
:lgtm:

> Reviewed 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/616#-:-NlZdyZI76PwY1kdMsNB:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @bobvawter and @sravotto)

___
*[`internal/script/script.go` line 208 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/616#-NlZcYdTD2ybya_neT9H:-NlZcYdTD2ybya_neT9I:b-lq43oo) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/01daf5fcc5ea850a084676d2b59fcaa93b279334/internal/script/script.go#L208)):*
> ```Go
> 		if bag.Apply != nil {
> 			tgt.Delegate = newApplier(s, table, bag.Apply)
> 			// If we're using a Delete & Upsert, no other options are valid.
> ```

This comment seems incorrect now.

bobvawter:
https://github.com/cockroachdb/cdc-sink/wiki/User-Scripts#custom-sql-interactions

Issue #617, opened 2023-12-08
Views aren't excluded from schemawatch, leading to nuisance staging-table-not-found messages


sjuls:
Problem;
We have database in cockroach we would like to migrate to a new cluster, however this database contains a view in its schema leading to errors in cdc-sink when the view is present in the target database. This also results in data from the changefeed not being applied to the target tables.

Error message
```
"relation \"movr.cdcsink.movr_public_sneaky_view\" does not exist", \"file\": \"errors.go\",\n \"line\": 171,\n \"routine\": \"NewUndefinedRelationError\"
```

Script to reproduce: [demo_up.sh](https://github.com/sjuls/cdc-sink/blob/view-erroring/scripts/shell_demo/demo_up.sh)

bobvawter:
It looks like views are not being excluded from the `SchemaData.Order` mapping, which is what drives the resolver.

sjuls:
> In `demo-up.sh`, permissions on the `leases` tables are being reassigned to `root`; that yields the following error: `user cdcsink does not have UPDATE privilege on relation leases`. This is unrelated to the views not being excluded from the schema data.

Yeah, that was testing a different failure mode, forgot to clean that up.

sjuls:
I see you changed the title, just want to confirm that it is not just a nuisance, but actually blocking data from being synced? Even removing the leases grant fiddling I still see data missing after the workload completes and cdc-sink has time to flush.

bobvawter:
It should have been fixed in PR #618.  Did that not work for you?

sjuls:
Hi @bobvawter ,

Yeah, it is fixed after your merge, thanks! Just wanted to make sure I understand the root of the missing data.

PR #618, opened 2023-12-08
schemawatch: Exclude views from SchemaData.Order

bobvawter:
The queries used to compute table dependency order does not exclude views. This
change updates the queries and adds a test.

X-Ref: #617

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/618)

### DIFF
--- a/internal/target/schemawatch/dependencies.go
+++ b/internal/target/schemawatch/dependencies.go
@@ -48,6 +48,7 @@ WITH RECURSIVE
         table_catalog, table_schema, table_name
       FROM
         information_schema.tables
+      WHERE table_type = 'BASE TABLE'
     ),
   refs
     AS (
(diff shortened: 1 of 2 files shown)


BramGruneir (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/618#-:-Nl9YjMX7eiJs-4PqAe9:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @sravotto)

internal/target/schemawatch/dependencies.go
```diff
@@ -165,7 +166,8 @@ const depOrderTemplateCRDB = `
 WITH RECURSIVE
  tables AS (
    SELECT schema_name AS sch, table_name AS tbl
-   FROM [SHOW TABLES FROM %[1]s]),
+   FROM [SHOW TABLES FROM %[1]s]
```
rafiss, line 169: just a drive-by comment, but SHOW statements are not considered stable APIs: https://www.cockroachlabs.com/docs/stable/api-support-policy#apis

so ideally this query would use pg_catalog or information_schema, so that it is less likely to break across different versions of CRDBl

bobvawter: We can remove this query and use the vanilla-PG one in its place once sufficiently-old CRDB versions are phased out of what we want cdc-sink to support as a target.

PR #651, opened 2024-01-17
logical: Disable the backfillWindow flag

bobvawter:
Backfill mode had indirectly dependended on the use of SELECT FOR UPDATE to ensure that mutations for two keys could not be in flight at once. When we removed the use of SFU because the locks weren't reliable during range transfers, backfill mode no longer had any particular guarantees about execution order.

The work to unify the non-transaction modes per issue #504 will provide a suitable replacement for the intended behavior of backfill mode. It will provide the option of allowing non-overlapping transactions to be applied concurrently in order to improve throughput.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/651)

### DIFF
--- a/internal/source/logical/config.go
+++ b/internal/source/logical/config.go
@@ -115,8 +115,13 @@ func (c *BaseConfig) Bind(f *pflag.FlagSet) {
 
 	f.DurationVar(&c.ApplyTimeout, "applyTimeout", defaultApplyTimeout,
 		"the maximum amount of time to wait for an update to be applied")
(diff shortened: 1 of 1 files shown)


BramGruneir (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/651#-:-NoN68OvE3kok7b4vvoN:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @bobvawter)

PR #652, opened 2024-01-18
script: Allow PKs used for deletion to be rewritten

bobvawter:
This change allows the userscript to provide a mapping function for the primary key in a deletion. This supports use cases where the target schema has a different PK structure than the source. It is also possible to elide all deletes to a table by having the deleteKey function return null.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/652)

### DIFF
--- a/internal/script/loader.go
+++ b/internal/script/loader.go
@@ -32,6 +32,16 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+// Analogous to mapJS, save that it operates on a primary-key array. A
+// null or empty slice will be interpreted as a request to elide the
+// delete operation.
+//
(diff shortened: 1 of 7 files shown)


BramGruneir (approved):
:lgtm: 

Would be good to update the wiki too with the examples to added to the tests.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/652#-:-NoT0P0E3YhM9L1PwYPX:b-yyd9q3)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @sravotto)

bobvawter:
Updated https://github.com/cockroachdb/cdc-sink/wiki/User-Scripts#table-mappers with the examples.

PR #683, opened 2024-02-08
stopvar: Fix timing bug in test

bobvawter:
The test for DoWhenChangedOrInterval has a timing bug that was only visible on heavily-loaded test runners where the timer fired before the call to Set(). This change explicitly tests this flow by using a short timeout and a channel to control when the call to Set() is made.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/683)

### DIFF
--- a/internal/util/stopvar/stopvar_test.go
+++ b/internal/util/stopvar/stopvar_test.go
@@ -63,33 +63,41 @@ func TestDoWhenChangedOrInterval(t *testing.T) {
 	defer cancel()
 
 	var called atomic.Bool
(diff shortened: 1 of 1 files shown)


BramGruneir (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/683#-:-NqDg6k9Dy2RSpbbqwLi:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @sravotto)

PR #697, opened 2024-02-15
besteffort: Revise logic

bobvawter:
This change fixes a bug in the BestEffort.Sweep implementation where the resolved timestamp window could incorrectly advance under certain timing conditions, even though there were still unapplied mutations. This would leave dangling mutations in the staging table. The problem is a logic bug in how Unstage is being used.

As previously written, the sweepOnce method would retry all staged mutations within the open resolved-timestamp window. If mutations had to be deferred a second time, they would have been marked with a deferral lease via the call to Unstage. In any a subsequent call to sweepOnce, there is no signal from Unstage that a mutation had been suppressed due to an existing lease, which should prevent the minimum bound of the window from advancing.

This change keeps the existng LeaseExpiry to avoid implicitly marking mutations as having been applied, but it does set the IgnoreLeases flag. In addition, the minimum bound of the resolving window only advances if every attempted mutation was actually applied.

The concept of a "deferrable" error isn't as robust as it could be. Instead, we have a way to filter uninteresting errors from the log. Any error encountered while attempting to apply mutations to the target table causes the mutations to be staged. This enables a source to better isolated from any malfunction in the target. As long as the mutation is written somewhere, either to the target or to staging, then BestEffort has fulfilled its contract.

This change should resolve the flakiness seen in the CI system when testing against Oracle.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/697)

### DIFF
--- a/internal/sequencer/besteffort/acceptor.go
+++ b/internal/sequencer/besteffort/acceptor.go
@@ -23,6 +23,7 @@ import (
 	"github.com/cockroachdb/cdc-sink/internal/types"
 	"github.com/cockroachdb/cdc-sink/internal/util/metrics"
 	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
 	"golang.org/x/sync/errgroup"
 )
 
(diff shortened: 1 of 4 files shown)


bobvawter:
I have a similar concern that staging can grow without bounds and was thinking that we need some kind of lockout mechanism to push back on a given target schema based on conditions like timestamps going backwards, excessive number of unapplied mutations, etc.

PR #699, opened 2024-02-15
sinktest: Improve robustness of integration tests

bobvawter:
We have observed cases of the go-ora driver getting stuck while tearing down
the prepared-statement cache and subsequently closing the database pool. This
will cause the tests to be delayed long enough to trigger the golang test
and/or GitHub runner timeouts.

This change removes the teardown of the statement cache, since we're going to
close the connection pool anyway. The call to close the Oracle db connection
pool is initiated by a call to `stopper.Context.Close()`, but is made
asynchronous to allow the context to stop, even if the call to `sql.DB.Close()`
blocks.

As an aid to future debugging, the workflow timeout is incresed to 20m to allow
the go test timeout of 10m to be able to trigger a stack dump if things get
stuck.

Overall go test parallelism is limited to prevent the workers from being
overloaded, causing the 2m per-test-fixture timeout to be hit.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/699)


sravotto (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/699#-:-Nqi31cz-Nnuuw88AmTZ:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @BramGruneir)

PR #708, opened 2024-02-18
besteffort: Make sweepOnce parallel

bobvawter:
The implementation of sweepOnce should have been embarrassingly parallel. Instead, it was embarrassingly serial, which limited catch-up throughput.

This change makes sweepOnce act similarly to the best-effort acceptor type, where each single-mutation batch is attempted in parallel.

An extra test focusing on the catch-up case is added. It can be used an an ersatz performance-tuning exercise by increasing the number of backfilled mutations.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/708)

### DIFF
--- a/internal/sequencer/besteffort/best_effort.go
+++ b/internal/sequencer/besteffort/best_effort.go
@@ -19,6 +19,7 @@ package besteffort
 
 import (
 	"context"
+	"sync/atomic"
 	"time"
 
 	"github.com/cockroachdb/cdc-sink/internal/sequencer"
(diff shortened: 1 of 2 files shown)


BramGruneir (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/708#-:-Nr7jPtw-layKTObKojr:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @sravotto)

PR #715, opened 2024-02-20
cdc: Decouple from HTTP request context

bobvawter:
This change avoids using the HTTP request context and instead uses a context with a configurable timeout. This will help to insulate cdc-sink from HTTP requests that are abandonded by the client (or which have a network blip). It is likely the case that the client will re-deliver the payload, but cdc-sink is already prepared to deal with the at-least-once behavior of CDC changefeeds.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/715)

### DIFF
--- a/internal/source/cdc/config.go
+++ b/internal/source/cdc/config.go
@@ -27,8 +27,9 @@ import (
 )
 
 const (
-	defaultBackfillWindow = time.Hour
-	defaultNDJsonBuffer   = bufio.MaxScanTokenSize // 64k
+	defaultBackfillWindow  = time.Hour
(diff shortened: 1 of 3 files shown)


BramGruneir (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/715#-:-Nr7V__01K8RGqnfKz35:bg7igkv)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @bobvawter)

PR #721, opened 2024-02-24
script: Present database values as reified JSON values

### LINKED ISSUE #664: userscripts not handling INT8 correctly
When a userscript is used to process rows that contain INT8 values, the values are converted to Javascript `number` which are double precision floats. They can only represent numbers up to 2^53 - 1 (instead of  2^63-1).

To reproduce, set up a C2C environment and run the KV workload, using the following userscript:

```
import * as api from "cdc-sink@v1";
import {Document, Table} from "cdc-sink@v1";

api.configureSource("kv.public", {
    dispatch: (doc: Document, meta: Document): Record<Table, Document[]> => {
        let ret: Record<Table, Document> = {};
        ret[meta.table] = [doc];
        return ret
    }
})
```

bobvawter:
This change corrects an inconsistency in how values loaded from the target database versus those from incoming changefeed values are presented to the userscript. Previously, types such as time.Time would be presented as a basic string when received from a changefeed, while the values loaded from the database were presented as a proxied time.Time. Large magnitude numbers were also impacted by JavaScript's 53-bit mantissa limitation.

All values presented to the userscript are passed through a utility function which ensures a consistent, JSON-like presentation of the value to the script.

The apply.IsMergeSupported() function is added to improve test readability.

Breaking change: Performing math using basic JavaScript numeric types should not be undertaken lightly, due to potential loss of precision, so numeric values are now always presented as strings, which the user may choose to coerce to a numeric type.

Fixes #664

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/721)

### DIFF
--- a/internal/script/bag_wrapper.go
+++ b/internal/script/bag_wrapper.go
@@ -43,7 +43,12 @@ func (m *bagWrapper) Get(key string) goja.Value {
 	if !ok {
 		return goja.Undefined()
 	}
-	return m.rt.ToValue(v)
+	ret, err := safeValue(m.rt, v)
+	if err != nil {
+		// This will be presented as an exception.
+		panic(m.rt.NewGoError(err))
+	}
+	return ret
 }
 
 // Has implements goja.DynamicObject.
--- a/internal/script/bag_wrapper_test.go
+++ b/internal/script/bag_wrapper_test.go
@@ -1,0 +1,70 @@
+// Copyright 2024 The Cockroach Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package script
+
+import (
+	"testing"
+	"time"
+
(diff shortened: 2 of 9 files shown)


BramGruneir (approved):
:lgtm:

With some more test cases.

> Reviewed 9 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/721#-:-NraHxQS2tSKxNjRiHwj:b3253gx)* status: all files reviewed, 1 unresolved discussion (waiting on @bobvawter and @sravotto)

___
*[`internal/script/safe_value_test.go` line 29 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/721#-NraGA0gCebxax5_Gl_Z:-NraGA0gCebxax5_Gl__:b-bezplv) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/52d2e5f200147a1f157143a5e14bd637622d98b7/internal/script/safe_value_test.go#L29)):*
> ```Go
> )
> 
> func TestSafeValue(t *testing.T) {
> ```

This doesn't seem to test all the code paths of the safe value function.

Can you add a bunch more test cases please?

bobvawter (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/721)* status: 7 of 9 files reviewed, 1 unresolved discussion (waiting on @BramGruneir and @sravotto)

___
*[`internal/script/safe_value_test.go` line 29 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/721#-NraGA0gCebxax5_Gl_Z:-Nrf7lHO3eYh1j0pURO4:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/52d2e5f200147a1f157143a5e14bd637622d98b7/internal/script/safe_value_test.go#L29)):*

This doesn't seem to test all the code paths of the safe value function.

Can you add a bunch more test cases please?
</blockquote></details>

Done.

PR #740, opened 2024-03-04
sequencer: Use a common lockset for concurrent sequencers

### LINKED ISSUE #726: shingle acceptor limits throughput due to too many small transactions
The shingle acceptor type has over-rotated a bit and creates too many small transactions to be performant.  It should instead break the incoming MultiBatch up into small-enough transactions (on the order of, say ~1000 mutations) to avoid an excessive number of database round-trips.

bobvawter:
This PR consists of three commits to optimize throughput and improve correctness.

---

This change revises the staging and target pool sizes and timeouts to better
align with operational best practices. Additional points of configuration have
been added to set the maximum idle lifetime and jitter.

---

This change allows invocation of the callback functions to be delegated to
another type, instead of using the go keyword. This will allow, for example,
the use of a workgroup to limit overall concurrency.

----

This change adds a new `sequencer/scheduler` package that provides a
centralized way for BestEffort and Shingle to ensure that their various
goroutines preserve the order of updates on individual keys. This scheduler is
built around the existing `lockset.Set`, configured with a `workgroup.Group`
that enforces the Parallelism configuration option.

This commit fixes an occasional race condition in BestEffort when a mutation
contains multiple updates for the same key. It was possible for the mutations
to be applied in an out-of-order sequence. This was causing occasional test
failures and could be reproduced locally by adding a random delay while
processing batches.

The shingle implementation was creating an excessive number of tasks, so it now
operates on larger segments. This segmentation code is driven by a new
configuration field, `FlushSize`.

The serial sequencer does not use the scheduler since it operates on a single
goroutine.

There were a few tests that were not calling the configuration Preflight
method. This has been corrected.

Fixes #726

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/740)

### DIFF
--- a/internal/cmd/preflight/preflight.go
+++ b/internal/cmd/preflight/preflight.go
@@ -77,7 +77,7 @@ func testTargetConnection(ctx *stopper.Context, connString string) error {
 	pool, err := stdpool.OpenTarget(
 		ctx,
 		connString,
-		stdpool.WithConnectionLifetime(5*time.Minute),
+		stdpool.WithConnectionLifetime(5*time.Minute, time.Minute, 15*time.Second),
 		stdpool.WithTransactionTimeout(time.Minute),
 	)
 	if err != nil {
@@ -133,7 +133,7 @@ func testStagingConnection(ctx *stopper.Context, connString string) error {
 	pool, err := stdpool.OpenPgxAsStaging(
 		ctx,
 		connString,
-		stdpool.WithConnectionLifetime(5*time.Minute),
+		stdpool.WithConnectionLifetime(5*time.Minute, time.Minute, 15*time.Second),
 		stdpool.WithTransactionTimeout(time.Minute),
 	)
 	if err != nil {
--- a/internal/sequencer/besteffort/acceptor.go
+++ b/internal/sequencer/besteffort/acceptor.go
@@ -21,10 +21,11 @@ import (
 	"time"
 
 	"github.com/cockroachdb/cdc-sink/internal/types"
+	"github.com/cockroachdb/cdc-sink/internal/util/lockset"
 	"github.com/cockroachdb/cdc-sink/internal/util/metrics"
+	"github.com/cockroachdb/cdc-sink/internal/util/notify"
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
-	"golang.org/x/sync/errgroup"
 )
 
 type acceptor struct {
@@ -34,33 +35,10 @@ type acceptor struct {
 
 var _ types.TableAcceptor = (*acceptor)(nil)
 
-// AcceptTableBatch implements [types.TableAcceptor]. This
-// implementation ignores the options, since we always want to use the
-// staging or target database pools.
-func (a *acceptor) AcceptTableBatch(
-	ctx context.Context, batch *types.TableBatch, _ *types.AcceptOptions,
-) error {
-	ctx, cancel := context.WithCancel(ctx)
-	defer cancel()
-
-	eg, egCtx := errgroup.WithContext(ctx)
-	if p := a.cfg.Parallelism; p > 0 {
-		eg.SetLimit(p)
-	}
-
-	if err := a.applyOrStage(egCtx, eg, batch, a.delegate); err != nil {
-		return err
-	}
-	return eg.Wait()
-}
-
-// Unwrap is an informal protocol to return the delegate.
-func (a *acceptor) Unwrap() types.TableAcceptor {
-	return a.delegate
-}
-
-// applyOrStage will write mutations directly into the destination
-// table or write them into a staging table.
+// AcceptTableBatch implements [types.TableAcceptor]. It will write
+// mutations directly into the destination table or write them into a
+// staging table. This implementation ignores the options, since we
+// always want to use the staging or target database pools.
 //
 // This method will stage mutations if a staged mutation for the key
 // already exists. This ensures that per-key data flows in an ordered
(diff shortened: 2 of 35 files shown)


BramGruneir (approved):
:lgtm:

With a few minor comments.

> Reviewed 12 of 12 files at r1, 2 of 2 files at r2, 21 of 21 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/740#-:-NsH8Jbf2mZknyDxUuxh:btbwjz3)* status: all files reviewed, 3 unresolved discussions (waiting on @bobvawter and @sravotto)

___
*[`internal/sequencer/scheduler/scheduler.go` line 19 at r3](https://reviewable.io/reviews/cockroachdb/cdc-sink/740#-NsH6zQy5AH3KqiNLqxX:-NsH6zQy5AH3KqiNLqxY:b-ltacb7) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/ff1685d086248a4a015900199a6ee2eec76d8b24/internal/sequencer/scheduler/scheduler.go#L19)):*
> ```Go
> // Package scheduler contains Sequencer-specific utilities for ensuring
> // ordered access to rows.
> package scheduler
> ```

While the logic here is pretty simple, a simple test to ensure things don't rot is probably a good idea.
___
*[`internal/sequencer/shingle/acceptor.go` line 75 at r3](https://reviewable.io/reviews/cockroachdb/cdc-sink/740#-NsH7DdP0r1qBzlPPDT9:-NsH7DdP0r1qBzlPPDTA:b2nigjw) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/ff1685d086248a4a015900199a6ee2eec76d8b24/internal/sequencer/shingle/acceptor.go#L75)):*
> ```Go
> 	// to avoid sending hundreds of trivially-small transactions to the
> 	// target.
> 	segments := segmentMultiBatch(batch, a.cfg.FlushSize)
> ```

yay
___
*[`internal/sequencer/shingle/acceptor.go` line 117 at r3](https://reviewable.io/reviews/cockroachdb/cdc-sink/740#-NsH7QHSBmHyiA7eYexG:-NsH7QHSBmHyiA7eYexH:boabhg) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/ff1685d086248a4a015900199a6ee2eec76d8b24/internal/sequencer/shingle/acceptor.go#L117)):*
> ```Go
> 			}
> 
> 			// Mark the mutations as having been applied.
> ```

Could you do this async?  

If we've got only a single instance applying, but multithreaded, we could update it in memory and start the next apply while async mark the mutations as applied?  As long as there is only a single markapplied thread, if things go funky, they won't get out of order.  (I think)
___
*[`internal/util/stdpool/size.go` line 45 at r4](https://reviewable.io/reviews/cockroachdb/cdc-sink/740#-NsH86eS94QLuwYGciZm:-NsH86eS94QLuwYGciZn:b6v2yn1) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/5b2e543eaf6acd64b8f4e2ab9231f89f9a78f802/internal/util/stdpool/size.go#L45)):*
> ```Go
> }
> func (o *withPoolSize) sqlDB(_ context.Context, impl *sql.DB) error {
> 	impl.SetMaxIdleConns(2)
> ```

why the magic number?  Didn't you specify this as a overridable default earlier?

bobvawter (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/740)* status: 34 of 35 files reviewed, 1 unresolved discussion (waiting on @BramGruneir and @sravotto)

___
*[`internal/sequencer/shingle/acceptor.go` line 117 at r3](https://reviewable.io/reviews/cockroachdb/cdc-sink/740#-NsH7QHSBmHyiA7eYexG:-NsJHCSO-3CSYh_bVAyJ:b-fklgnj) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/ff1685d086248a4a015900199a6ee2eec76d8b24/internal/sequencer/shingle/acceptor.go#L117)):*

"Time might reverse" is a warning that should be on every software product.

Any way to test that this doesn't happen?  Purposefully block s1 and let s2 in?  (you might have already tested this case.
</blockquote></details>

The new, unified. sequencer test package injects a chaos shim that exercises the error paths when sending data to the `apply` package. I verified that a case where an early segment fails out while a latter one succeeds through some printf tracing.

sravotto (approved):
:lgtm: Just a minor comment

> Reviewed 10 of 12 files at r1, 2 of 2 files at r2, 19 of 21 files at r3, 1 of 1 files at r4, 1 of 3 files at r5.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/740#-:-NsJJH5v5VbGX3W8IzJO:b-hac4vk)* status: 34 of 35 files reviewed, 2 unresolved discussions (waiting on @bobvawter and @BramGruneir)

___
*[`internal/util/lockset/lockset.go` line 36 at r4](https://reviewable.io/reviews/cockroachdb/cdc-sink/740#-NsJI-bn-mVobonIWk3x:-NsJI-bn-mVobonIWk3y:bmofaaw) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/5b2e543eaf6acd64b8f4e2ab9231f89f9a78f802/internal/util/lockset/lockset.go#L36)):*
> ```Go
> type Runner interface {
> 	// Go should execute the function in a non-blocking fashion.
> 	Go(func(context.Context)) error
> ```

Could you add to workgroup.go
`
var \_ lockset.Runner \= (\*Group)(nil)
`
Perhaps I'd move the interface definition there.

bobvawter (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/740)* status: 34 of 35 files reviewed, 2 unresolved discussions (waiting on @BramGruneir and @sravotto)

___
*[`internal/util/lockset/lockset.go` line 36 at r4](https://reviewable.io/reviews/cockroachdb/cdc-sink/740#-NsJI-bn-mVobonIWk3x:-NsJQm6xC6h_D448WDE-:b-3n9g3c) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/5b2e543eaf6acd64b8f4e2ab9231f89f9a78f802/internal/util/lockset/lockset.go#L36)):*

Could you add to workgroup.go
`
var \_ lockset.Runner \= (\*Group)(nil)
`
Perhaps I'd move the interface definition there.
</blockquote></details>

That would create an import cycle.  It's also idiomatic for consumers to declare an interface that happens to be implemented by some other type.

BramGruneir (commented):
:lgtm:

> Reviewed 1 of 3 files at r5.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/740#-:-NsJ_ZNr73f55kh1n0o_:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @sravotto)

BramGruneir (approved):
:lgtm: with the new fix.

> Reviewed 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/740#-:-NsKp2-56nsuVyk0xOuT:b4c2nwz)* status: all files reviewed, 1 unresolved discussion (waiting on @sravotto)

PR #742, opened 2024-03-04
types: Fix noCopy API leakage

bobvawter:
We use the noCopy linter trick to ensure that batch types are never copied. The current embedding makes the no-op `Lock()` method visible to potential callers, which is confusing and incorrect. This change makes the noCopy field anonymous, so these methods no longer pollute the exported API.

The TableBatch types was missing a noCopy marker, and this is now fixed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/742)

### DIFF
--- a/internal/types/batches.go
+++ b/internal/types/batches.go
@@ -56,7 +56,7 @@ type Batch[B any] interface {
 // A well-formed MultiBatch will have its Data field sorted by time.
 // MultiBatch implements [sort.Interface] to make this simple.
 type MultiBatch struct {
-	noCopy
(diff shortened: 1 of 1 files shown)


BramGruneir (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/742#-:-Ns96TYM2v8GIiPWfPhd:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @sravotto)

PR #750, opened 2024-03-06
shingle: Add fast retry path for FK constraint violations

### LINKED ISSUE #748: Shingle + parallelism + FK needs a fast-retry for constraint violations
Shingle mode can encounter constraint-violation errors when transaction T2 refers to rows in transaction T1. We need to detect this case, using the logic from BestEffort, and retry that transaction when it reaches the ultimate head-of-line in the scheduler.

The `lockset` API should have a sentinel `ErrRetryAtHead` so that a callback will be re-executed when all other callbacks that were enqueued ahead of the problematic one have been dispatched. This can be accomplished by making the waiters have a `next` pointer and the `lockset.Set` type have `head` and `tail` pointers.

bobvawter:
This PR consists of two commits which enable `lockset.Set` to retry tasks when they have
reached the head of a global queue and for the `shingle` sequencer to make use of this functionality.

This changes the shingle acceptor to automatically retry FK constraint
violations or other deferrable errors if there were potentially-blocking
segments already in flight. Previously, these FK errors would cause the entire
shingled batch to be retried after the quiescent period elapsed, leading to
throughput hiccups. The existing sequencer test suite already exercised the
top-level FK-ordering retry behavior.

The is-deferrable-error function has been moved out of the besteffort package
into the sequtil package.

Fixes #748

This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/750)

### DIFF
--- a/internal/sequencer/besteffort/acceptor.go
+++ b/internal/sequencer/besteffort/acceptor.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"time"
 
+	"github.com/cockroachdb/cdc-sink/internal/sequencer/sequtil"
 	"github.com/cockroachdb/cdc-sink/internal/types"
 	"github.com/cockroachdb/cdc-sink/internal/util/lockset"
 	"github.com/cockroachdb/cdc-sink/internal/util/metrics"
@@ -121,7 +122,7 @@ func (a *acceptor) AcceptTableBatch(
 			}
 			deferredCount.Inc()
 			// We'll suppress errors like FK constraint violations.
-			if !isNormalError(err) {
+			if !sequtil.IsDeferrableError(err) {
 				log.WithError(err).Warnf(
 					"staging mutation instead of target table %s key %s",
 					singleBatch.Table, string(singleBatch.Data[0].Key))
--- a/internal/sequencer/besteffort/best_effort.go
+++ b/internal/sequencer/besteffort/best_effort.go
@@ -34,11 +34,7 @@ import (
 	"github.com/cockroachdb/cdc-sink/internal/util/notify"
 	"github.com/cockroachdb/cdc-sink/internal/util/stopper"
 	"github.com/cockroachdb/cdc-sink/internal/util/stopvar"
-	"github.com/go-sql-driver/mysql"
-	"github.com/jackc/pgx/v5/pgconn"
-	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
-	"github.com/sijms/go-ora/v2/network"
 	log "github.com/sirupsen/logrus"
 )
 
@@ -316,7 +312,7 @@ func (s *BestEffort) sweepOnce(
 				// Ignore expected errors, we'll try again later. We'll
 				// leave the lease intact so that we might skip the
 				// mutation until a later cycle.
-				if isNormalError(err) {
+				if sequtil.IsDeferrableError(err) {
 					deferrals.Inc()
 					return nil
 				}
@@ -406,34 +402,4 @@ func (s *BestEffort) sweepOnce(
 		}
 	}
 	return nil
-}
-
-// isNormalError returns true if the error represents an error that's
-// likely to go away on its own in the future (e.g. FK constraints).
-// These are errors that we're ok with reducing log levels for.
-//
-// https://github.com/cockroachdb/cdc-sink/issues/688
-func isNormalError(err error) bool {
-	if pgErr := (*pgconn.PgError)(nil); errors.As(err, &pgErr) {
-		return pgErr.Code == "23503" // foreign_key_violation
-	} else if myErr := (*mysql.MySQLError)(nil); errors.As(err, &myErr) {
-		// Cannot add or update a child row: a foreign key constraint fails
-		return myErr.Number == 1452
-	} else if oraErr := (*network.OracleError)(nil); errors.As(err, &oraErr) {
-		switch oraErr.ErrCode {
-		case 1: // ORA-0001 unique constraint violated
-			// The MERGE that we execute uses read-committed reads, so
-			// it's possible for two concurrent merges to attempt to
-			// insert the same row.
-			return true
-		case 60: // ORA-00060: Deadlock detected
-		// Our attempt to insert ran into another transaction, possibly
-		// from a different cdc-sink instance. This can happen since a
-		// MERGE operation reads before it starts writing and the order
-		// in which locks are acquired may vary.
-		case 2291: // ORA-02291: integrity constraint
-			return true
-		}
-	}
-	return false
 }
--- a/internal/sequencer/sequtil/deferrable.go
+++ b/internal/sequencer/sequtil/deferrable.go
@@ -1,0 +1,60 @@
+// Copyright 2024 The Cockroach Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package sequtil
+
+import (
+	"github.com/go-sql-driver/mysql"
+	"github.com/jackc/pgx/v5/pgconn"
+	"github.com/pkg/errors"
+	"github.com/sijms/go-ora/v2/network"
+)
+
+// IsDeferrableError returns true if the error represents an error
+// that's likely to go away on its own in the future (e.g. FK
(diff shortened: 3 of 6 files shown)


sravotto (approved):
:lgtm:

> Reviewed 2 of 2 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/750#-:-NsPzwi-4LLAnekPUF76:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @BramGruneir)

BramGruneir (approved):
:lgtm: 

With a few questions

> Reviewed 2 of 2 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/750#-:-NsTMAhh2M5Xxrr0Nnab:bfr7l2z)* status: all files reviewed, 4 unresolved discussions (waiting on @bobvawter)

___
*[`internal/sequencer/shingle/acceptor.go` line 152 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/750#-NsTLU7dCiF6nJdGbvhF:-NsTLU7dCiF6nJdGbvhG:b9h0lux) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/2d7e93209d70e4ec8d541bca84a4be11e7e0415f/internal/sequencer/shingle/acceptor.go#L152)):*
> ```Go
> 
> // tryBatch attempts to apply the given segment.
> func (a *acceptor) tryBatch(
> ```

Is this just pulled out for testing purposes?  Because acceptmultibatch seems quite small already
___
*[`internal/sequencer/shingle/acceptor.go` line 170 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/750#-NsTM4hm0D7SIrdhX3Pm:-NsTM4hm0D7SIrdhX3Pn:b-fomamh) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/2d7e93209d70e4ec8d541bca84a4be11e7e0415f/internal/sequencer/shingle/acceptor.go#L170)):*
> ```Go
> 	opts.TargetQuerier = tx
> 	if err := a.delegate.AcceptMultiBatch(ctx, segment, opts); err != nil {
> 		return err
> ```

no with stack?  I guess the acceptmultibatch will already do that
___
*[`internal/util/lockset/lockset.go` line 101 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/750#-NsTIfZNCApPLey7oY5p:-NsTIfZO0kiI8eauQM6B:b-rtfi7u) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/bde5aeae8cc74014cfc3ebdf5db6c0d96928aafe/internal/util/lockset/lockset.go#L101)):*
> ```Go
> func (s *Status) String() string {
> 	switch s {
> 	case executing:
> ```

So why are you doing this instead of just using an enum?  I'm sure there was a reason when I first looked at the lockset code.
___
*[`internal/util/lockset/lockset.go` line 330 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/750#-NsTJUuT7EKwGmRFZl9n:-NsTJUuT7EKwGmRFZl9o:b-so5w22) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/bde5aeae8cc74014cfc3ebdf5db6c0d96928aafe/internal/util/lockset/lockset.go#L330)):*
> ```Go
> 		// Once the waiter has been called, update its status and call
> 		// dequeue to find any tasks that have been unblocked.
> 		var finalErr error
> ```

why is this logic in a defer and not the normal codepath?
___
*[`internal/util/lockset/lockset.go` line 419 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/750#-NsTJmO56ndwjN81aEyR:-NsTJmO56ndwjN81aEyS:b-omxihh) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/bde5aeae8cc74014cfc3ebdf5db6c0d96928aafe/internal/util/lockset/lockset.go#L419)):*
> ```Go
> 
> 	// Insert the waiter into the global queue.
> 	if s.mu.tail == nil {
> ```

I really wish go had some of these generic so you don't have to re-write queue logic again and again chancing errors each time.

bobvawter (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/750)* status: all files reviewed, 4 unresolved discussions (waiting on @BramGruneir)

___
*[`internal/sequencer/shingle/acceptor.go` line 152 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/750#-NsTLU7dCiF6nJdGbvhF:-NsTQzBP5OMsGowNPc-Y:bnief74) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/2d7e93209d70e4ec8d541bca84a4be11e7e0415f/internal/sequencer/shingle/acceptor.go#L152)):*

Is this just pulled out for testing purposes?  Because acceptmultibatch seems quite small already
</blockquote></details>

`tryBatch` has multiple return paths.  This makes the task callback much shorter since it can be just

```
			err := a.tryBatch(stopCtx, segment, opts)
			if sequtil.IsDeferrableError(err) {
				return lockset.RetryAtHead(err)
			}
			return err
```
___
*[`internal/sequencer/shingle/acceptor.go` line 170 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/750#-NsTM4hm0D7SIrdhX3Pm:-NsTS7YiDnJi0-rTJ6_g:b-jhcqw4) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/2d7e93209d70e4ec8d541bca84a4be11e7e0415f/internal/sequencer/shingle/acceptor.go#L170)):*

no with stack?  I guess the acceptmultibatch will already do that
</blockquote></details>

The rubric that I've been using is to use `WithStack` when calling code that we don't own (e.g. DB access).  All cdc-sink internal APIs are then implicitly "stack ready" and we don't wind up with nested error stacks that say the same thing.
___
*[`internal/util/lockset/lockset.go` line 101 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/750#-NsTIfZNCApPLey7oY5p:-NsTSUHtCYJmHv3jyYG3:b-rb3syu) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/bde5aeae8cc74014cfc3ebdf5db6c0d96928aafe/internal/util/lockset/lockset.go#L101)):*

So why are you doing this instead of just using an enum?  I'm sure there was a reason when I first looked at the lockset code.
</blockquote></details>

When the task callback returns an error, we cook up a new `Status{err}` instance.  In all other cases, we use these sentinel instances.  So it's exactly like an enum, except when it isn't.
___
*[`internal/util/lockset/lockset.go` line 330 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/750#-NsTJUuT7EKwGmRFZl9n:-NsTXCDw6TVFNlLyEWxl:b-j9w2s6) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/bde5aeae8cc74014cfc3ebdf5db6c0d96928aafe/internal/util/lockset/lockset.go#L330)):*

why is this logic in a defer and not the normal codepath?
</blockquote></details>

The sheer number of different exit paths below that may need to pass an error back.

I tried setting this up in a more conventional way, but it turned into a hugely nested mess of conditional code.  Even though this is a little bit harder to grok on first examination, there's exactly one linear chunk of source code that describes how we execute the tasks.  We let the compiler deal with the fact that there's many different combinations of return paths.
___
*[`internal/util/lockset/lockset.go` line 419 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/750#-NsTJmO56ndwjN81aEyR:-NsT_PCSAedlkLCJaXxJ:b-gdei8h) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/bde5aeae8cc74014cfc3ebdf5db6c0d96928aafe/internal/util/lockset/lockset.go#L419)):*

I really wish go had some of these generic so you don't have to re-write queue logic again and again chancing errors each time.
</blockquote></details>

It's better in go 1.22, with generic collections now being part of the standard library.  I'll add a cleanup issue to see where adopting that code makes sense.

bobvawter (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/750)* status: 1 of 6 files reviewed, 4 unresolved discussions (waiting on @BramGruneir and @sravotto)

___
*[`internal/util/lockset/lockset.go` line 330 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/750#-NsTJUuT7EKwGmRFZl9n:-NsTddbH9VyDgJA56CHl:bw226x6) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/bde5aeae8cc74014cfc3ebdf5db6c0d96928aafe/internal/util/lockset/lockset.go#L330)):*

The sheer number of different exit paths below that may need to pass an error back.

I tried setting this up in a more conventional way, but it turned into a hugely nested mess of conditional code.  Even though this is a little bit harder to grok on first examination, there's exactly one linear chunk of source code that describes how we execute the tasks.  We let the compiler deal with the fact that there's many different combinations of return paths.
</blockquote></details>

I hoisted the panic handler out into a helper function and now this path is built in an entirely conventional way.

PR #760, opened 2024-03-08
besteffort: Increase batch size to improve throughput

bobvawter:
This change updates besteffort to use reasonably-sized segments, instead of
singleton updates, when sweeping. This is in the same spirit as the previous
update to the `shingle` package.

The segmentation helper functions are extracted from the `shingle` package into
the `sequtil` package.

The msort helper functions are updated to return new slices, instead of
operating in place. This corrects a subtle bug when trying to re-process
batches.

The `seqtest` package is updated to verify that rows arrive at an expected end
state. This was use to investigate the `msort` issue.

This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/760)


BramGruneir (approved):
:lgtm:

> Reviewed 11 of 11 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/760#-:-NsVAkoT5VQ1Sy3kaVRV:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @bobvawter and @sravotto)

___
*[`internal/sequencer/besteffort/best_effort.go` line 343 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/760#-NsV7jaV856vo5_w5g6N:-NsV7jaV856vo5_w5g6O:b-k1w5pk) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/6b3cce4114ba6b63e987e1825ec329368773319a/internal/sequencer/besteffort/best_effort.go#L343)):*
> ```Go
> 					deferrals.Inc()
> 				} else {
> 					errCount.Inc()
> ```

When it's non-deferrable, this is just due to the an issue with the db right?
___
*[`internal/sequencer/sequtil/segment.go` line 95 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/760#-NsVAcA0A5QeLzkxp5_M:-NsVAcA0A5QeLzkxp5_N:byx0e0q) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/6b3cce4114ba6b63e987e1825ec329368773319a/internal/sequencer/sequtil/segment.go#L95)):*
> ```Go
> 				Data:  muts,
> 				Table: table,
> 				// Time is indefinite.
> ```

sage words

bobvawter (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/760)* status: all files reviewed, 1 unresolved discussion (waiting on @BramGruneir and @sravotto)

___
*[`internal/sequencer/besteffort/best_effort.go` line 343 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/760#-NsV7jaV856vo5_w5g6N:-NsVKBQyBhMQBtCZW0F-:bevt3hw) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/6b3cce4114ba6b63e987e1825ec329368773319a/internal/sequencer/besteffort/best_effort.go#L343)):*

When it's non-deferrable, this is just due to the an issue with the db right?
</blockquote></details>

Correct, incrementing that counter would indicate a somebody-do-something error when talking to the database.

PR #775, opened 2024-03-15
types: Add CopyInto methods to mutation batch types

bobvawter:
This change adds a general-pupose, ordered, traversal mechanism to the batch types. This supports an upcoming commit and a few existing batch traversals are update to use CopyInto.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/775)

### DIFF
--- a/internal/sequencer/scheduler/scheduler.go
+++ b/internal/sequencer/scheduler/scheduler.go
@@ -98,15 +98,10 @@ func keyForBatch(batch *types.MultiBatch) []string {
 	// The keys will be deduplicated and ordered by
 	// lockset.Set.Schedule, so we don't need to worry about that here.
 	var ret []string
-	for _, sub := range batch.Data {
-		// Ignoring error because callback only returns nil.
-		_ = sub.Data.Range(func(tbl ident.Table, tblData *types.TableBatch) error {
-			for _, mut := range tblData.Data {
-				ret = append(ret, keyForSingleton(tbl, mut)...)
-			}
-			return nil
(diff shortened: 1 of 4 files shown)


BramGruneir (approved):
:lgtm:

with one question

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/775#-:-Nt1rq3y48WDcX_yoOdn:b-hxealz)* status: all files reviewed, 1 unresolved discussion (waiting on @bobvawter and @sravotto)

___
*[`internal/types/batches.go` line 133 at r1](https://reviewable.io/reviews/cockroachdb/cdc-sink/775#-Nt1rQDQ8mVrZiAf7ZlT:-Nt1rQDQ8mVrZiAf7ZlU:blk6b6t) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/72617c742f9c19e66720363d132325e66c5d0dd0/internal/types/batches.go#L133)):*
> ```Go
> func (b *MultiBatch) Copy() *MultiBatch {
> 	ret := &MultiBatch{
> 		ByTime: make(map[hlc.Time]*TemporalBatch, len(b.Data)),
> ```

Why are you adding bytime if it's never referenced anywhere besides where you set it below?

bobvawter:
It was an error to have `ByTime` missing from the `Copy` method. It's also used to make `MultiBatch.Accumulator` go faster, since we can insert directly into the necessary `TemporalBatch`.

PR #784, opened 2024-03-19
apply: Report original value if ColData.Parse fails

bobvawter:
The existing code reports a meaningless value if the call to Parse fails. The assignment to the value variable is deferred until after the error check.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/784)


BramGruneir (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/784)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @sravotto)

PR #804, opened 2024-03-29
stmtcache: Move statement destruction out of hot path

bobvawter:
The call to sql.Stmt.Close can involve some blocking on a mutex within pgx. This change moves the call into a separate goroutine. The metrics in this package aren't labeled, so the metrics fields in the Cache type are unnecessary.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/804)


BramGruneir (approved):
:lgtm:

> Reviewed 10 of 10 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/804#-:-Nu9NItW4792VTg3lXGV:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @sravotto)

PR #808, opened 2024-04-02
dashboard: removing hardcoded datasource uid from filters.

sravotto:
This change fixes the grafana replicator dashboard ad-hoc filtering by using the DS_PROMETHEUS variable in place of the hardcoded uid.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/808)


bobvawter (approved):
LGTM

I used the "export for sharing" setting, I wonder why that entry wasn't right.

PR #812, opened 2024-04-08
apply, stage: Add support for CockroachDB NO_FULL_SCAN hint

bobvawter:
This change adds an explicit hint for hot-path UPSERT statements in the apply and stage packages. We see cases where stale or missing table statistics in CockroachDB cause an UPSERT to execute with a full table scan, leading to performance decay. By adding an `@{NO_FULL_SCAN}` hint, we can force the optimizer to issue a lookup join.

The `ident.Hinted` wrapper associates an identifier with an arbitrary suffix that contains a table or index hint. This design is generalized to support other potential target database use cases that would benefit from hints.

In the specific case of CockroachDB, support for this hint is added in v23.1.17 and v23.2.3, so we need to conditionally inject the hinting behavior. This is accomplished by adding another helper function to PoolInfo that will decorate a table name if appropriate.

While it would be possible to set the CockroachDB `disallow_full_table_scans` session variable by way of the pgx connection parameters, there are some bootstrap and metrics cases that do require full scans. The number of performance-critical queries that require this hint is relatively small, hence the choice to selectively hint.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/812)

### DIFF
--- a/internal/source/cdc/server/integration_test.go
+++ b/internal/source/cdc/server/integration_test.go
@@ -35,6 +35,7 @@ import (
 	"github.com/cockroachdb/cdc-sink/internal/util/diag"
 	"github.com/cockroachdb/cdc-sink/internal/util/ident"
 	"github.com/cockroachdb/cdc-sink/internal/util/stdlogical"
+	"github.com/cockroachdb/cdc-sink/internal/util/stdpool"
 	"github.com/cockroachdb/cdc-sink/internal/util/stdserver"
 	"github.com/cockroachdb/cdc-sink/internal/util/stopper"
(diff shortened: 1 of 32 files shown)


sravotto (approved):
:lgtm:

> Reviewed 32 of 32 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/812#-:-NuzUkddD6uMzYUxkrS4:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @BramGruneir)

PR #817, opened 2024-04-09
Kafka - support for TLS/SASL

sravotto:
This PR contains 2 commits:
The first commit moves the TLSConfig function to util/secure so it can be used by
other packages.

The second commit adds new configuration parameters to allow a secure connection
with the Kafka cluster. It also provides support for SASL authentication.

On start up, the user may specify the location of certificates and private key
to connect to a Kafka cluster using the Transport Layer Security (TLS).

SASL support is provided by specifying either a user/password for plain text/SCRAM
verification or by specifying the URL of an third party identity provider(IdP)
that supports OATH2. OATH2 requires the user to provide a client id and client secret
to authenticate with the IdP.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/817)


sravotto (commented):
Thanks.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/817#-:-NvYCPS91EoVH3aWkBrC:b-ucfkn4)* status: 4 of 19 files reviewed, 3 unresolved discussions (waiting on @bobvawter and @BramGruneir)

___
*[`internal/source/kafka/config.go` line 108 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/817#-NvWxQhk8Z23UvHBYrtH:-NvY5kIb2lS6IubqgcWs:b-9kgxru) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/2ea02dc9263effed8ebfcbfaac8b6037ea6153fd/internal/source/kafka/config.go#L108)):*

Are these the same as other secure config settings?  If so, please pull it all into it's on library instead of including them twice.
</blockquote></details>

Moved to the util/secure package.
___
*[`internal/source/kafka/config.go` line 182 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/817#-NvWxz_U2NkTA6uN4CpJ:-NvY5iX6DYK2Y1enmec4:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/2ea02dc9263effed8ebfcbfaac8b6037ea6153fd/internal/source/kafka/config.go#L182)):*

Shouldn't this be part of preflight?  I don't recall seeing buildConfig in other places, but I may be wrong.
</blockquote></details>

Done.
___
*[`internal/source/kafka/config.go` line 226 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/817#-NvWyNQG1byfrzB4HHnn:-NvY-N5oDDXrsFGZ0wJE:b-2ded64) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/2ea02dc9263effed8ebfcbfaac8b6037ea6153fd/internal/source/kafka/config.go#L226)):*

What does a blank mechanism mean?
</blockquote></details>

Either no authentication or certificated based authentication.
___
*[`internal/source/kafka/conn_test.go` line 50 at r2](https://reviewable.io/reviews/cockroachdb/cdc-sink/817#-NvWzb8ICQmlZ50AxVxS:-NvY016G-oFgWeet6mHJ:b-xnkxpr) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/2ea02dc9263effed8ebfcbfaac8b6037ea6153fd/internal/source/kafka/conn_test.go#L50)):*

I can see this mock being useful for non-kafka tests.
</blockquote></details>

Agreed. However, we should make it more generic.

PR #831, opened 2024-04-22
apply: Add knob for per-statement row limit

bobvawter:
The pgwire protocol has a limit of 2^15 bind variables per statement. The number of bind variables in our generated SQL is rows*colums, which means that extremely wide tables can run into a practical limit on the number of rows that can be sent in a single statement. Our current default number of rows is 100, which is somewhat conservative for normal-width schemas.

This change adds a new field to apply.Config that controls the number of rows to send for non-bulk targets. It is exposed via the userscript and adjusts the capacity of the buffers used when assembling database payloads.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/831)

### DIFF
--- a/internal/script/loader.go
+++ b/internal/script/loader.go
@@ -125,6 +125,9 @@ type targetJS struct {
 	// Two- or three-way merge operator. The bindMerge method will
 	// validate the type of value.
 	Merge goja.Value `goja:"merge"`
+	// For targets without a bulk-transfer mechanism, the maximum number
+	// of rows to send in a single statement.
+	RowLimit int `goja:"rowLimit"`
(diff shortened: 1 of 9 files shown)


sravotto (approved):
:lgtm:

> Reviewed 9 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/831#-:-NwA4Wa082K5OcbuMgbc:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @BramGruneir)

PR #833, opened 2024-04-24
acceptor: Moving CountingAcceptor to types, so it can be reused.

sravotto:
Currently, the cdc source package uses a counting acceptor wrapper to record all the mutations that are pushed through the acceptor machinery, and exposes metrics to track outcomes.

This change moves the counting acceptor and relative metrics to the types package, so it can be reused by other sources.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/833)

### DIFF
--- a/internal/source/cdc/metrics.go
+++ b/internal/source/cdc/metrics.go
@@ -17,11 +17,6 @@
 package cdc
 
 import (
-	"context"
-
-	"github.com/cockroachdb/cdc-sink/internal/types"
-	"github.com/cockroachdb/cdc-sink/internal/util/ident"
-	"github.com/cockroachdb/cdc-sink/internal/util/metrics"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 )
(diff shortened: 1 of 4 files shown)


bobvawter (changes requested):
The existing metrics names need to be preserved in the `cdc` use-case, this is going to unnecessarily break some existing customer dashboards.  I would have the extracted type contains some function pointers to allow the "what to do with the count" to be arbitrarily composable. 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/833#-:-NwFhr5J7esdxitu2mBh:bkjplmu)* status: 0 of 5 files reviewed, all discussions resolved

sravotto (commented):
good point. I changed so we don't break the existing metrics.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/833#-:-NwG8fUh9_Q8NqlHWzo-:b-nasvxx)* status: 0 of 5 files reviewed, all discussions resolved

bobvawter (changes requested):
> Reviewed 1 of 5 files at r3.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/833)* status: 1 of 5 files reviewed, 2 unresolved discussions (waiting on @sravotto)

___
*[`internal/types/counter_test.go` line 37 at r3](https://reviewable.io/reviews/cockroachdb/cdc-sink/833#-NwKUVA36e9zlo90zzcb:-NwKUVA36e9zlo90zzcc:b-yywhtk) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/3e75aa769c9b440b4a1e3d61de730ed8373ef0f7/internal/types/counter_test.go#L37)):*
> ```Go
> 
> var (
> 	mutationsErrorCount = promauto.NewCounterVec(prometheus.CounterOpts{
> ```

There's no need to set up actual metrics here and go through the rigamarole of interfacing with the prometheus collector.  You have the `Counter` interface.

```go
type counter float64

func (c *counter) Add(f float64) {
  *c = *c+f
}

func (c *counter) Value() float64 {
  return *c
}
```
___
*[`internal/types/counter_test.go` line 69 at r3](https://reviewable.io/reviews/cockroachdb/cdc-sink/833#-NwKV0VUB9f11GgY5RnO:-NwKV0VUB9f11GgY5RnP:b6gt4dq) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/3e75aa769c9b440b4a1e3d61de730ed8373ef0f7/internal/types/counter_test.go#L69)):*
> ```Go
> 	r := require.New(t)
> 	a := assert.New(t)
> 	ctx, cancel := context.WithCancel(context.Background())
> ```

In general, prefer `context.WithTimeout()`.  The test runner has a default timeout of 10m, but that's a long time doing nothing for a test to hang a CI runner.

sravotto (commented):
Please take another look.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/833#-:-NwL2KBFEY4MPeaDd2y8:b1xog4q)* status: 1 of 5 files reviewed, 2 unresolved discussions (waiting on @bobvawter)

___
*[`internal/types/counter_test.go` line 37 at r3](https://reviewable.io/reviews/cockroachdb/cdc-sink/833#-NwKUVA36e9zlo90zzcb:-NwL2Gop2eYxfwU2aMzc:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/3e75aa769c9b440b4a1e3d61de730ed8373ef0f7/internal/types/counter_test.go#L37)):*

There's no need to set up actual metrics here and go through the rigamarole of interfacing with the prometheus collector.  You have the `Counter` interface.

```go
type counter float64

func (c *counter) Add(f float64) {
  *c = *c+f
}

func (c *counter) Value() float64 {
  return *c
}
```
</blockquote></details>

Done.
___
*[`internal/types/counter_test.go` line 69 at r3](https://reviewable.io/reviews/cockroachdb/cdc-sink/833#-NwKV0VUB9f11GgY5RnO:-NwL2EREBlTVR28o8XX2:b-896fix) ([raw file](https://github.com/cockroachdb/cdc-sink/blob/3e75aa769c9b440b4a1e3d61de730ed8373ef0f7/internal/types/counter_test.go#L69)):*

In general, prefer `context.WithTimeout()`.  The test runner has a default timeout of 10m, but that's a long time doing nothing for a test to hang a CI runner.
</blockquote></details>

Done.

bobvawter (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/833)* status: 1 of 5 files reviewed, 1 unresolved discussion

PR #841, opened 2024-04-30
schemawatch: Handle byte vs char length semantics in Oracle Database

bobvawter:
This change improves the schema introspection to ensure that the type `CHAR(1 BYTE)` and `CHAR(1 CHAR)` remain distinguishable at runtime. Previously, they would have been extracted as `CHAR(1)`, whose interpretation varies based on a variety of conditions.

The `NLS_LENGTH_SEMANTICS` session variable is set upon connection to ensure that any unspecified semantics always default to char type.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/841)


BramGruneir (approved):
:lgtm:

wow

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/cdc-sink/841#-:-NwoL3nt3KRAMAPbSjEu:batbgvi)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @sravotto)

PR #850, opened 2024-05-08
staging: Rename database objects

bobvawter:
This commit changes the default staging database name from _cdc_sink to _replicator. The provider function for the staging database name is updated to look for the legacy name and use that instead.

Various references to the _cdc_sink database have been updated to reflect the new name. Existing documents in the migrations directory are not updated, since they are historical documents only.

The default name for the DLQ table has been adjusted without automatic backward-compatibility, since this has only been used in limited testing scenarios.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/850)


BramGruneir (approved):
:lgtm:

> Reviewed 16 of 16 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/850#-:-NxOUOoY-nVV8fXpoAHo:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @bobvawter)

___
*[`internal/sinktest/base/provider.go` line 111 at r1](https://reviewable.io/reviews/cockroachdb/replicator/850#-NxOU6-f47b7cUngWTbK:-NxOU6-g8DET9LAAFYBX:baig5g2) ([raw file](https://github.com/cockroachdb/replicator/blob/0cc7add3c0a956d13777f9374a69ca95ee2f771e/internal/sinktest/base/provider.go#L111)):*
> ```Go
> 	SourcePool   *types.SourcePool       // Access to user-data tables and changefeed creation.
> 	SourceSchema sinktest.SourceSchema   // A container for tables within SourcePool.
> 	StagingPool  *types.StagingPool      // Access to _replicator database.
> ```

Why not just call these staging instead of _replicator?

PR #854, opened 2024-05-09
A collection of more renames

BramGruneir:
Just some more renaming cleanup.  
Only one that might affect people is renaming the default application name for pgx.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/854)


bobvawter (approved):
:lgtm:

PR #866, opened 2024-05-13
cmd/workload: Add synthetic workload generator command

bobvawter:
This change adds a workload generator that synthesizes webhook payloads for a
parent/child table model. It generates a sequence of data payloads,
interspersed with resolved timestamp messages. Message ordering across retries
are managed by re-using the `lockset` type to ensure per-key order is
maintained.

This re-uses the existing workload generator and checker code from the
sequencer tests. The `seqtest/generate.go` file is moved into `util/workload`
and split into a `GeneratorBase` and `Checker` type.

The workload command has both an all-inclusive "demo" subcommand as well as a
"run" command that presumes an existing Replicator instance.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/866)


BramGruneir (approved):
:lgtm:

With just a few comments.

> Reviewed 25 of 25 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/866#-:-NxrdbiwA5S47DzmUARi:b-ftdfso)* status: all files reviewed, 6 unresolved discussions (waiting on @bobvawter, @MattWhelan, and @sravotto)

___
*[`internal/cmd/workload/runner.go` line 161 at r1](https://reviewable.io/reviews/cockroachdb/replicator/866#-NxraSRCAqeGFb1xwpWe:-NxraSRDE5LAQ2kuVqCz:biska7a) ([raw file](https://github.com/cockroachdb/replicator/blob/716211fc92720e4174d5678b60829080a910cf75/internal/cmd/workload/runner.go#L161)):*
> ```Go
> // returned previously. It will increment the logical counter if the
> // wall time has not advanced.
> func (r *runner) hlcNow() hlc.Time {
> ```

This isn't threadsafe right?  It's only called once at a time?  If so, what's the chance of a collision on now()?
___
*[`internal/sinktest/all/workload.go` line 34 at r1](https://reviewable.io/reviews/cockroachdb/replicator/866#-NxrQP6j4BYdWTYC525h:-NxrQP6j4BYdWTYC525i:bja6x9n) ([raw file](https://github.com/cockroachdb/replicator/blob/716211fc92720e4174d5678b60829080a910cf75/internal/sinktest/all/workload.go#L34)):*
> ```Go
> )
> 
> // Workload combines [workload.Workload]
> ```

Buffalo buffalo Buffalo buffalo buffalo buffalo Buffalo buffalo.
___
*[`internal/staging/stage/metrics.go` line 28 at r1](https://reviewable.io/reviews/cockroachdb/replicator/866#-NxrLANM0ihBeycd1nCo:-NxrLANM0ihBeycd1nCp:bl0m05p) ([raw file](https://github.com/cockroachdb/replicator/blob/716211fc92720e4174d5678b60829080a910cf75/internal/staging/stage/metrics.go#L28)):*
> ```Go
> 	stageMarkDuration = promauto.NewHistogramVec(prometheus.HistogramOpts{
> 		Name:    "stage_mark_duration_seconds",
> 		Help:    "the amount of time it took to mark a mutation as applied",
> ```

Is this the delay between it being applied an getting marked or just the marking sql itself?

Looks like just the sql itself... please update the message to make that clear.
I can see a reason why you would want a metric on every sql statement, why not add it directly to Retry so  you can pass in the metric?  (in a follow up pr perhaps)
___
*[`internal/util/workload/base.go` line 49 at r1](https://reviewable.io/reviews/cockroachdb/replicator/866#-NxrTrs8AO0aodQICUlO:-NxrTrs8AO0aodQICUlP:bjg0jcq) ([raw file](https://github.com/cockroachdb/replicator/blob/716211fc92720e4174d5678b60829080a910cf75/internal/util/workload/base.go#L49)):*
> ```Go
> 
> 	// Compare data to ensure time order is maintained.
> 	ParentVals map[int]int64 `json:"parentVals"`
> ```

nit: swap order here to match the top.  (Or the other way around)
___
*[`internal/util/workload/base.go` line 55 at r1](https://reviewable.io/reviews/cockroachdb/replicator/866#-NxrU7yv-zBSXHJO7Ska:-NxrU7yv-zBSXHJO7Skb:b-j4u4mo) ([raw file](https://github.com/cockroachdb/replicator/blob/716211fc92720e4174d5678b60829080a910cf75/internal/util/workload/base.go#L55)):*
> ```Go
> 	ChildToParent map[int]int `json:"childToParent"`
> 
> 	ctr int
> ```

comment please
___
*[`internal/util/workload/base.go` line 74 at r1](https://reviewable.io/reviews/cockroachdb/replicator/866#-NxrUaDJEF0HXrq4NNFV:-NxrUaDJEF0HXrq4NNFW:b-95ime6) ([raw file](https://github.com/cockroachdb/replicator/blob/716211fc92720e4174d5678b60829080a910cf75/internal/util/workload/base.go#L74)):*
> ```Go
> func (g *GeneratorBase) GenerateInto(batch *types.MultiBatch, time hlc.Time) {
> 	val := time.Nanos()
> 	switch g.ctr % 5 {
> ```

Shouldn't this be based on some percentage instead of the repeated in order operations?
___
*[`internal/util/workload/base.go` line 139 at r1](https://reviewable.io/reviews/cockroachdb/replicator/866#-NxrZ2sUB_cOaSni34X9:-NxrZ2sUB_cOaSni34XA:b-vbut63) ([raw file](https://github.com/cockroachdb/replicator/blob/716211fc92720e4174d5678b60829080a910cf75/internal/util/workload/base.go#L139)):*
> ```Go
> // created data.
> func (g *GeneratorBase) Range() hlc.Range {
> 	return hlc.RangeIncluding(hlc.Zero(), g.MaxTime)
> ```

What the point of this if it just is always returning maxTime?  Plans to start a non-zero?
___
*[`internal/util/workload/base.go` line 168 at r1](https://reviewable.io/reviews/cockroachdb/replicator/866#-NxrZC7z4_lRE08aZAJr:-NxrZC7z4_lRE08aZAJs:b7kf3gp) ([raw file](https://github.com/cockroachdb/replicator/blob/716211fc92720e4174d5678b60829080a910cf75/internal/util/workload/base.go#L168)):*
> ```Go
> 		return child
> 	}
> 	panic("no children")
> ```

This is my life when it's suddenly quiet and they are not in the room.  
___
*[`internal/util/workload/base.go` line 176 at r1](https://reviewable.io/reviews/cockroachdb/replicator/866#-NxrZKhs1WcRcwe7A-Tq:-NxrZKhs1WcRcwe7A-Tr:b-e1u52w) ([raw file](https://github.com/cockroachdb/replicator/blob/716211fc92720e4174d5678b60829080a910cf75/internal/util/workload/base.go#L176)):*
> ```Go
> 		return parent
> 	}
> 	panic("no parents")
> ```

And when my children are alone, this is always what they do.
___
*[`internal/util/workload/base.go` line 182 at r1](https://reviewable.io/reviews/cockroachdb/replicator/866#-Nxr_0Ph-CQdmFOScyzq:-Nxr_0Ph-CQdmFOScyzr:b-8yu042) ([raw file](https://github.com/cockroachdb/replicator/blob/716211fc92720e4174d5678b60829080a910cf75/internal/util/workload/base.go#L182)):*
> ```Go
> 	for {
> 		child := int(rand.Int31())
> 		if _, exists := g.Children[child]; exists {
> ```

I'd love to see a counter on these for collisions... Just so you know should should have played the lottery that day.

bobvawter (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/866)* status: all files reviewed, 5 unresolved discussions (waiting on @BramGruneir, @MattWhelan, and @sravotto)

___
*[`internal/cmd/workload/runner.go` line 161 at r1](https://reviewable.io/reviews/cockroachdb/replicator/866#-NxraSRCAqeGFb1xwpWe:-NxrfJm10bZrgXQT2B9X:bweokos) ([raw file](https://github.com/cockroachdb/replicator/blob/716211fc92720e4174d5678b60829080a910cf75/internal/cmd/workload/runner.go#L161)):*

This isn't threadsafe right?  It's only called once at a time?  If so, what's the chance of a collision on now()?
</blockquote></details>

The chance of collision is zero, since the `runner.Run()` method is single-threaded. We're able to generate batches faster than we can ultimately process them, so we achieve concurrency by making network requests in parallel rather than by generating in parallel.
___
*[`internal/sinktest/all/workload.go` line 34 at r1](https://reviewable.io/reviews/cockroachdb/replicator/866#-NxrQP6j4BYdWTYC525h:-Nxrg0umBDmzmORGUA1R:bbz2djc) ([raw file](https://github.com/cockroachdb/replicator/blob/716211fc92720e4174d5678b60829080a910cf75/internal/sinktest/all/workload.go#L34)):*

Buffalo buffalo Buffalo buffalo buffalo buffalo Buffalo buffalo.
</blockquote></details>

Revised comment.
___
*[`internal/staging/stage/metrics.go` line 28 at r1](https://reviewable.io/reviews/cockroachdb/replicator/866#-NxrLANM0ihBeycd1nCo:-Nxrg8fCDsuawTCVDIYF:bw2l3g8) ([raw file](https://github.com/cockroachdb/replicator/blob/716211fc92720e4174d5678b60829080a910cf75/internal/staging/stage/metrics.go#L28)):*

Is this the delay between it being applied an getting marked or just the marking sql itself?

Looks like just the sql itself... please update the message to make that clear.
I can see a reason why you would want a metric on every sql statement, why not add it directly to Retry so  you can pass in the metric?  (in a follow up pr perhaps)
</blockquote></details>

Updated the comment to clarify that it's about a SQL statement.

A retry block may encapsulate any number of statements, but I do like the idea of naming retry blocks to make it easier to provide higher-level metrics around database-related tasks in the system.
___
*[`internal/util/workload/base.go` line 55 at r1](https://reviewable.io/reviews/cockroachdb/replicator/866#-NxrU7yv-zBSXHJO7Ska:-NxrgvvBA2CgbIPdm3LQ:b-d6cdyv) ([raw file](https://github.com/cockroachdb/replicator/blob/716211fc92720e4174d5678b60829080a910cf75/internal/util/workload/base.go#L55)):*

comment please
</blockquote></details>

Renamed to `generation` and added a comment.
___
*[`internal/util/workload/base.go` line 74 at r1](https://reviewable.io/reviews/cockroachdb/replicator/866#-NxrUaDJEF0HXrq4NNFV:-NxriNnlAQR7s-eTBahb:b-q8xjh2) ([raw file](https://github.com/cockroachdb/replicator/blob/716211fc92720e4174d5678b60829080a910cf75/internal/util/workload/base.go#L74)):*

Shouldn't this be based on some percentage instead of the repeated in order operations?
</blockquote></details>

It could be percentage based, but I like the deterministic selection of scenarios provided by this approach.  My goal is to create an ever-expanding population of parents and children.
___
*[`internal/util/workload/base.go` line 139 at r1](https://reviewable.io/reviews/cockroachdb/replicator/866#-NxrZ2sUB_cOaSni34X9:-NxrhPb0CGwm49_U5a7U:b23uz9z) ([raw file](https://github.com/cockroachdb/replicator/blob/716211fc92720e4174d5678b60829080a910cf75/internal/util/workload/base.go#L139)):*

What the point of this if it just is always returning maxTime?  Plans to start a non-zero?
</blockquote></details>

Quantities that represent making progress in doing something are represented by an `hlc.Range`. It had become very confusing to pass around plain-old `hlc.Time` values to communicate progress.  Is that specific `hlc.Time` inherently inclusive or exclusive?  The `hlc.Range` type is defined as a half-open range and the constructor functions make it obvious at the call-site what the intended use of the specific `hlc.Time` is.  At some point I'd like to introduce an `hlc.Progress` type that acts like a `[ 0, time )` range.
___
*[`internal/util/workload/base.go` line 168 at r1](https://reviewable.io/reviews/cockroachdb/replicator/866#-NxrZC7z4_lRE08aZAJr:-NxrikCf2JvPr-XeAT2k:b-vhca5t) ([raw file](https://github.com/cockroachdb/replicator/blob/716211fc92720e4174d5678b60829080a910cf75/internal/util/workload/base.go#L168)):*

This is my life when it's suddenly quiet and they are not in the room.
</blockquote></details>

https://a.co/d/2WogM0h
___
*[`internal/util/workload/base.go` line 176 at r1](https://reviewable.io/reviews/cockroachdb/replicator/866#-NxrZKhs1WcRcwe7A-Tq:-Nxrj9mS7KHmAkMcPRHV:b-b78g0c) ([raw file](https://github.com/cockroachdb/replicator/blob/716211fc92720e4174d5678b60829080a910cf75/internal/util/workload/base.go#L176)):*

And when my children are alone, this is always what they do.
</blockquote></details>

Inculcating object permanence is outside of my area of expertise.

Issue #898, opened 2024-06-04
Handle non-idempotent Changefeed redelivery


bobvawter:
CockroachDB changefeeds do not guarantee an idempotent ordering of messages under all checkpointing and/or re-delivery scenarios. For some row `R`, it is possible to see versions of the row `1, 2a, 3, 2b, 4` (the suffix is for clarity below). While this satisfies the at-least-once delivery guarantees, it is something of a novel discovery that the system behaves this way.

This is a non-issue in fully-consistent mode, since the staging tables would suppress the duplicate `2b` update.

In eventually-consistent mode (BestEffort), we bypass the staging tables whenever possible. This leads to a case where the row could roll back under the `2b` message (and perhaps there is no `4` message). This could lead to rows that are desynchronized.  Immediate mode is also impacted.

The proposed solution is to have BestEffort mode also write stub entries to the staging table which are applied, but that do not contain a payload. This will at least decrease the total number of bytes transferred, even if we do incur an additional put operation.

Immediate mode could also write stub entries to staging tables, but it may be more flexible if we were to add optional high-water tracking in the target database. We could then guarantee exactly-once behavior, at the cost of somewhat more expensive writes into the target database.

bobvawter:
Related #565

PR #901, opened 2024-06-06
seqtest: Make CheckSequencer work with Immediate

bobvawter:
This change adds some extra configuration plumbing so that the Immedate sequencer can be tested and verified with approximately the same tests used on other sequencers. The points of configuration are to disable foreign keys in the target schema and to skip any tests which would pre-populate staging tables.

X-Ref: #898

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/901)

### DIFF
--- a/internal/cmd/workload/workload_test.go
+++ b/internal/cmd/workload/workload_test.go
@@ -62,7 +62,7 @@ func TestWorkload(t *testing.T) {
 		},
 	}
 
-	workload, group, err := fixture.NewWorkload(ctx)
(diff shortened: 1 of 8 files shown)


sravotto (approved):
:lgtm:

> Reviewed 6 of 8 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/901#-:-NzjHSH78H_pITA4XDVX:bnfp4nl)* status: all files reviewed, 3 unresolved discussions (waiting on @bobvawter, @BramGruneir, and @MattWhelan)

PR #915, opened 2024-06-13
sequencer: Address non-idempotent redeliveries in changefeeds

bobvawter:
Under certain redelivery circumstances, it is possible for a CockroachDB
changefeed to only emit a subset of updates that could be redelivered. For
example, a sequence of updates (1, 2, 3, 2, 4) may be received. This satisfies
the at-least-once changefeed guarantees, but it is something of a novel
discovery that changefeeds do not redeliver all messages in an idempotent
manner.

This does not affect the fully-consistent mode in Replicator, since all
mutations are passed through the staging tables and applied only upon receipt
of a resolved timestamp. In eventually-consistent modes (BestEffort and
Immediate), we do not have a mechanism for suppressing already-processed
mutations. This can lead to rows rolling back under a scenario such as
(1, 2, 3, 2)

The solution in this change is to ensure that an applied staging entry exists
for all processed mutations, regardless of whether or not the staging table
would otherwise be involved in processing the mutation.

Summary of changes:
* Stager.MarkApplied will create "stub" rows for unstaged mutations.
* Stager.FilterApplied is added to perform an anti-join.
* A sequencer/decorators package is added to compose filtering and marking
  behaviors within the Sequencer stack.
* The seqtest.Checker has been updated to replay a subset of
  previously-processed mutations in order to simulate non-idempotent
  redelivery.
* A --assumeIdempotent flag is added to restore the previous behavior for
  immediate-mode usecases that prioritize throughput above consistency.

X-Ref: #565
Fixes: #898

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/915)


sravotto (approved):
:lgtm:

> Reviewed 30 of 30 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/915#-:-O-IToET4cRHoDEc8sTS:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @bobvawter, @BramGruneir, and @MattWhelan)

___
*[`internal/sequencer/immediate/immediate.go` line 33 at r1](https://reviewable.io/reviews/cockroachdb/replicator/915#-O-INdIW-XxNpw852Q-k:-O-INdIW-XxNpw852Q-l:b-aoekq7) ([raw file](https://github.com/cockroachdb/replicator/blob/1d5736bd4e896830362ae9c2ae304977cb42c4d5/internal/sequencer/immediate/immediate.go#L33)):*
> ```Go
> 
> // Immediate is a trivial implementation of [sequencer.Sequencer] that
> // writes through to the underlying acceptor.
> ```

You might want to update the comment.

PR #917, opened 2024-06-13
kafka: deduping messages

sravotto:
Under some conditions a changefeed will emit duplicate messages to ensure the sink is receiving each message at least once.
However, the duplicate messages may be sent after newer messages, possibly overriding or deleting rows.

This change adds buffering in the Kafka connector, keeping track of the timestamp for each key, ensuring that older values for a particular row do not override most recent ones.

The buffer is backed by LRU cache. When we receive a resolved timestamp, we evict from cache any entries that have a timestamp that is older than the resolved timestamp.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/917)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,6 +18,7 @@ require (
 	github.com/google/go-licenses v1.6.0
 	github.com/google/uuid v1.6.0
 	github.com/google/wire v0.6.0
+	github.com/hashicorp/golang-lru/v2 v2.0.7
 	github.com/jackc/pglogrepl v0.0.0-20240307033717-828fbfe908e9
 	github.com/jackc/pgx/v5 v5.6.0
 	github.com/joonix/log v0.0.0-20200409080653-9c1d2ceb5f1d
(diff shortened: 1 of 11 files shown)


BramGruneir (commented):
The use of the LRU is interesting.  I was thinking of building a purpose built cache.  Because you could make it significantly quicker.  The issue I have with the LRU on you're using is that it's tied to actual time.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/917#-:-O-fcroq3C7-ZVc5U2TA:b-2ra2pi)* status: 0 of 11 files reviewed, 1 unresolved discussion (waiting on @sravotto)

___
*[`internal/source/kafka/cache.go` line 132 at r1](https://reviewable.io/reviews/cockroachdb/replicator/917#-O-fcD5G5jhto-GXypAd:-O-fcD5G5jhto-GXypAe:bkvrgdf) ([raw file](https://github.com/cockroachdb/replicator/blob/49acd54029c53cc7926306daad9780fb4c1ee378/internal/source/kafka/cache.go#L132)):*
> ```Go
> // markResolved indicates that any entries older that the given timestamp can be evicted.
> // if there are no entries that have a newer timestamp, the cache is purged.
> func (c *cache) markResolved(timestamp hlc.Time) {
> ```

this should hold the lock the whole time, locking twice is dangerous

PR #951, opened 2024-08-08
c2Oracle: remove linux tag and add support for env for OIC path

ZhouXing19:
With Oracle finally supporting Oracle Instant Client on MacOS Arm machine now, we can have replicator support these new arch with looser restriction on tags.

Users now have to specify the path to the OIC library via `LD_LIBRARY_PATH` or `DYLD_LIBRARY_PATH` env var so that the godror driver can connect to the oracle server.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/951)


bobvawter (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/951)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @ZhouXing19)

___
*[`internal/util/stdpool/ora.go` line 27 at r1](https://reviewable.io/reviews/cockroachdb/replicator/951#-O3rY6MxDsJUo73dg8AU:-O3rY6MxDsJUo73dg8AV:brv2iu0) ([raw file](https://github.com/cockroachdb/replicator/blob/17ad30365ef6dfde25eda08b69c0cf34c99bd19f/internal/util/stdpool/ora.go#L27)):*
> ```Go
> 	"time"
> 
> 	"github.com/cockroachdb/errors"
> ```

Please reset to `pkg/errors` and undo dependency changes.
___
*[`internal/util/stdpool/ora.go` line 89 at r1](https://reviewable.io/reviews/cockroachdb/replicator/951#-O3rYTeP2A5xzLFauReL:-O3rYTeQ2KvBBLE8GAj2:bs0nvji) ([raw file](https://github.com/cockroachdb/replicator/blob/17ad30365ef6dfde25eda08b69c0cf34c99bd19f/internal/util/stdpool/ora.go#L89)):*
> ```Go
> 	}
> 
> 	if os.Getenv("LD_LIBRARY_PATH") == "" && os.Getenv("LD_LIBRARY_PATH") == "" {
> ```

Setting these environment variables should obviate the need to set `params.LibDir` at all; they add additional search paths when the linker is resolving short library names (e.g. `dlopen()` and friends).  It would be reasonable, instead, to use an environment variable like `OIC_LIBRARY_PATH` that is Replicator-specific.

bobvawter (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/951)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @ZhouXing19)

___
*[`.github/workflows/go-tests.yaml` line 213 at r1](https://reviewable.io/reviews/cockroachdb/replicator/951#-O3r_4baB2gOMgzG272Q:-O3r_4baB2gOMgzG272R:b-cjvi0a) ([raw file](https://github.com/cockroachdb/replicator/blob/17ad30365ef6dfde25eda08b69c0cf34c99bd19f/.github/workflows/go-tests.yaml#L213)):*
> ```YAML
>           sudo mv -f /tmp/instantclient_21_13 /usr/lib
>           sudo rm /tmp/ora-libs.zip
>           sudo readlink -f /usr/lib/* | grep instantclient_21_13
> ```

Why these changes?

ZhouXing19:
This actually might not be necessary so closing

Issue #902, opened 2024-06-06
Allow split-mode Replicator to be restarted if the target is down


bobvawter:
If Replicator is deployed in split-mode (i.e. separate staging and target databases), it should be possible for a Replicator binary to be restarted if the target database is down. Replicator would continue to stage data until the target database has been restored, at which point the staged data will be applied as per usual. We would likely need to cache information about the target schema, since that is used to sanity-check incoming mutations.

bobvawter:
PR #980 adds persistence to the schemawatch package.

ryanluu12345:
If I'm understanding this correctly: right now we require that both target and staging must be up for us to even start the replicator C2X service right? So we basically want to remove this requirement if staging is up but target is down

bobvawter:
Correct. Matt has been working in PR #1000 to add a test rig for this.

PR #952, opened 2024-08-09
schemawatch: Refresh Watch method to return a notify.Var

bobvawter:
This change updates the Watcher.Watch() method to return a notify.Var to make it consistent with other sources of asynchronous updates. This will make a follow-on change to cache the target's schema cleaner to implement.

X-Ref: #902

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/952)

### DIFF
--- a/internal/target/apply/apply.go
+++ b/internal/target/apply/apply.go
@@ -28,7 +28,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/cockroachdb/field-eng-powertools/notify"
 	"github.com/cockroachdb/field-eng-powertools/stopper"
 	"github.com/cockroachdb/replicator/internal/target/load"
 	"github.com/cockroachdb/replicator/internal/types"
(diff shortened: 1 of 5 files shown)


BramGruneir (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/952#-:-O46TgmX2wRJcC1-JuLm:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @bobvawter and @sravotto)

___
*[`internal/target/schemawatch/watcher.go` line 89 at r1](https://reviewable.io/reviews/cockroachdb/replicator/952#-O46T52i61ef_uz_PL6a:-O46T52i61ef_uz_PL6b:b-c8wbii) ([raw file](https://github.com/cockroachdb/replicator/blob/ebc8ba4b8545e7e817237b200613d2b36edc418a/internal/target/schemawatch/watcher.go#L89)):*
> ```Go
> // Get implements types.Watcher.
> func (w *watcher) Get() *types.SchemaData {
> 	ret, _ := w.data.Get()
> ```

please add comments as to why the return values are not needed

PR #957, opened 2024-08-13
stage: Add optional staging table consistency checks

bobvawter:
This change adds a sanity check to the staging package to ensure that the staging table entries are marked as applied in an order which is consistent with the order in which the entries are staged. This check is optionally enabled in Stager.MarkApplied() by an environment variable. The check is also used to drive a monitoring metric.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/957)

### DIFF
--- a/internal/sequencer/seqtest/check.go
+++ b/internal/sequencer/seqtest/check.go
@@ -160,6 +160,7 @@ type Check struct {
 }
 
 // Check generates data and verifies that it reaches the target tables.
+// It will also ensure that the staging tables are in a correct state.
 func (c *Check) Check(ctx *stopper.Context, t testing.TB, cfg *all.WorkloadConfig) {
 	r := require.New(t)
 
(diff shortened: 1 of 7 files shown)


BramGruneir (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/957#-:-O4HQPKP7kVS9CZIwTf_:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @bobvawter)

___
*[`internal/staging/stage/consistency.go` line 29 at r1](https://reviewable.io/reviews/cockroachdb/replicator/957#-O4H8bWE371-3y91Konk:-O4H8bWE371-3y91Konl:b-7r5pq0) ([raw file](https://github.com/cockroachdb/replicator/blob/2fb5bff2a02d0b77595ba065735d321f821e8783/internal/staging/stage/consistency.go#L29)):*
> ```Go
> )
> 
> var extraSanityChecks = os.Getenv("REPLICATOR_SANITY_CHECKS") != ""
> ```

Why via an environment variable and not a just a flag?

bobvawter (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/957)* status: all files reviewed, 1 unresolved discussion (waiting on @BramGruneir)

___
*[`internal/staging/stage/consistency.go` line 29 at r1](https://reviewable.io/reviews/cockroachdb/replicator/957#-O4H8bWE371-3y91Konk:-O4Hmh2vEmFkaiR-AB3l:b-cf5kpf) ([raw file](https://github.com/cockroachdb/replicator/blob/2fb5bff2a02d0b77595ba065735d321f821e8783/internal/staging/stage/consistency.go#L29)):*

Why via an environment variable and not a just a flag?
</blockquote></details>

As discussed, this package doesn't have any other points of configuration and these checks exist for diagnostic purposes.

PR #959, opened 2024-08-15
stage: Prevent table reader from jumping backwards

bobvawter:
It is possible for a reader to re-emit rows in subsequent cursors if it has reached an idle state and the bounds variable is updated without advancing the minimum value. This will happen if a slow target database delays the resolved timestamp bound from advancing before a new resolved timestamp is accepted for processing.

This change updates the behavior of the table reader such that it will not re-emit rows, regardless of how the bounds variable is driven.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/959)

### DIFF
--- a/internal/staging/stage/table_merger_test.go
+++ b/internal/staging/stage/table_merger_test.go
@@ -206,6 +206,9 @@ func TestTableMerger(t *testing.T) {
 		r.Equal(bigMuts, seen)
 	})
 
+	// Moving the lower or upper bounds backwards will have no effect.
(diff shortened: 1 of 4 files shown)


BramGruneir (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/959#-:-O4Lr4FOCIP7o_RHf6tZ:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @bobvawter)

PR #960, opened 2024-08-15
internal/util: support oracle parser

ZhouXing19:
So that we can parse UPDATE/INSERT/DELETE sql stmt provided by logminer.

The parsing logic is to convert the sql statement to a {column_name: new_value} kv struct.

Example:

```sql
INSERT INTO "C##MYADMIN"."EMPLOYEE" ("ID", "NAME_TARGET", "SALARY", "DATE_JOINED") 
VALUES (1008, 'Sam Smith', 75000, NULL);
----
{
  "DATE_JOINED": null,
  "ID": "1008",
  "NAME_TARGET": "Sam Smith",
  "SALARY": "75000"
}
```

```sql
UPDATE "C##MYADMIN"."EMPLOYEE RECORDS" SET "NAME_TARGET" = 'Pink White', "SALARY" = 75000 WHERE "NAME_TARGET" = 'John Doe';
----
{
  "NAME_TARGET": "Pink White",
  "SALARY": "75000"
}

```

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/960)


internal/oracleparser/parser_test.go
```diff
@@ -0,0 +1,157 @@
...
+
+func (s *mockListener) EnterUpdate_statement(ctx *Update_statementContext) {
+	set := ctx.Update_set_clause()
+	allColSets := set.AllColumn_based_update_set_clause()
+	for i, colSet := range allColSets {
+		colName := colSet.Column_name().GetText()
+		valStr := colSet.Expression().GetText()
+		children := colSet.GetChildren()
+		if strings.HasPrefix(colName, "\"") && strings.HasSuffix(colName, "\"") {
+			colName = strings.TrimPrefix(strings.TrimSuffix(colName, "\""), "\"")
```
Jeremyyang920, line 65: Maybe a small readability thing. Using backtick here could maybe make this easier to read `"` instead of escaped character with double quote.

bobvawter:
Split the package up into `internal/util/oracleparser` and `internal/util/oracleparser/third_party`. Then the filters in the workflow file can exclude `*/third_party/*` as necessary.  We'll probably wind up with more of these cases as more db frontends are added.

internal/oracleparser/parser_test.go
ZhouXing19: Done.

ZhouXing19:
> Split the package up into internal/util/oracleparser and internal/util/oracleparser/third_party. Then the filters in the workflow file can exclude */third_party/* as necessary. We'll probably wind up with more of these cases as more db frontends are added.

Done. 

I have to exclude the whole `oracleparser` for lint though, because we have to explicitly implement `EnterUpdate_statement` and `EnterInsert_statement`, whose function name is decided by antlr and is already breaking the lint rule 🤷

internal/oracleparser/README.md
```diff
@@ -0,0 +1,25 @@
...
+need to regenerate the files.
+
+* Install the [ANTLR4 complete jar and a java environment](install) that can run it.
+* Download `PlSqlLexer.g4` and `PlSqlParser.g4` from [the definitions](plsqldefs).
+* Run `java -jar /path/to/antlr4.jar -Dlanguage=Go -no-visitor -package parser *.g4` (replace `/path/to/antlr4.jar`).
+* Replace all uses of `self` in `plsql_lexer.go` and `plsql_parser.go` with `p.`.
+  See [here](golangissue) for details.
+* Re-run gofmt: `gofmt -w *.go`.
+
+tl;dr: `antlr -Dlanguage=Go -no-visitor -package parser *.g4 && gofmt -w *.go && sed -i '' -e 's/self\./p\./g' plsql_*.go`
```
ryanluu12345, line 21: When exactly does this need to be run?

internal/oracleparser/parser_test.go
```diff
@@ -0,0 +1,158 @@
...
+
+func (kv kvStruct) String() (string, error) {
+	byteRes, err := json.MarshalIndent(kv, "", "  ")
+	if err != nil {
+		return "", err
+	}
+	return string(byteRes), nil
+}
+
+func (s *mockListener) EnterUpdate_statement(ctx *orclantl.Update_statementContext) {
```
ryanluu12345, line 58: Is the reason that we have this odd naming convention here to implement the interface provided by a listener that `godror` or `antlr` provides?

internal/oracleparser/testdata/insert_antrl.ddt
```diff
@@ -0,0 +1,212 @@
...
+}
+
+# Insert with Nested Functions
+parse
+INSERT INTO "C##MYADMIN"."EMPLOYEE" ("ID", "NAME_TARGET", "SALARY") 
+VALUES (1005, UPPER('nancy drew'), NVL(NULL, 60000));
+----
+{
+  "ID": "1005",
+  "NAME_TARGET": "UPPER('nancy drew')",
```
ryanluu12345, line 63: For functions, does `replicator` convert those to the CRDB equivalent? I'm just thinking, what happens if there are UDFs here?

ZhouXing19:
The code quality test finally passed. RFAL, thanks!

internal/oracleparser/testdata/update_antrl.ddt
```diff
@@ -0,0 +1,163 @@
+parse
+UPDATE "C##MYADMIN"."EMPLOYEE" SET "NAME_TARGET" = 'Pink White', "SALARY" = 75000 WHERE "NAME_TARGET" = 'John Doe';
```
ryanluu12345, line 2: Similar comment for collection types for Oracle

internal/oracleparser/README.md
Jeremyyang920: Only when you make a change the the actual parser

internal/oracleparser/parser_test.go
Jeremyyang920: It needs to satisfy the antlr interface which is written in java. This is a method override essentially. Same thing in SCT

internal/oracleparser/parser_test.go
```diff
@@ -0,0 +1,158 @@
...
+	colNames := colList.AllColumn_name()
+	for i, col := range colNames {
+		colName := col.GetText()
+		valStr := ValList[i].GetText()
+		if strings.HasPrefix(colName, "\"") && strings.HasSuffix(colName, "\"") {
+			colName = strings.TrimPrefix(strings.TrimSuffix(colName, "\""), "\"")
+		}
+		if strings.HasPrefix(valStr, "'") && strings.HasSuffix(valStr, "'") {
+			valStr = strings.TrimPrefix(strings.TrimSuffix(valStr, "'"), "'")
+		}
```
Jeremyyang920, line 103: Oops missed this on my other pass, but same backtick comment here for the " escape.

Jeremyyang920 (approved):
Initial parser LGTM. We can always add more tests and cases as we keep developing.

internal/oracleparser/testdata/insert_antrl.ddt
ZhouXing19: actually log miner won't output subqueries in the logs, i'm just trying to be universal here. the logs that logminer gives will only give the actual values to be updated / inserted.

bobvawter (approved):
:lgtm:  I'd still like to see this placed into `internal/util/oracleparser` since it's not a stateful service API.

> Reviewed 1 of 18 files at r1, 1 of 13 files at r2.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/960#-:-O4Ql_vkAdPc710Ez_iR:b-9rc4k7)* status: 2 of 18 files reviewed, 8 unresolved discussions (waiting on @Jeremyyang920, @ryanluu12345, and @ZhouXing19)

___
*[`internal/oracleparser/parser_test.go` line 34 at r3](https://reviewable.io/reviews/cockroachdb/replicator/960#-O4QjTyEDtPU2InidHHb:-O4QjTyEDtPU2InidHHc:br2vj8o) ([raw file](https://github.com/cockroachdb/replicator/blob/141d0c4bb3654cf3ba4e70e18c4c9225a3d1bcf8/internal/oracleparser/parser_test.go#L34)):*
> var _ antlr.SomeInterface = (*mockListener)(nil)

is typical to declare conformation to an interface.
___
*[`internal/oracleparser/parser_test.go` line 58 at r3](https://reviewable.io/reviews/cockroachdb/replicator/960#gh-1719949251:-O4Qj0sp3kIX6MDPBKiN:b-jnwhxj) ([raw file](https://github.com/cockroachdb/replicator/blob/141d0c4bb3654cf3ba4e70e18c4c9225a3d1bcf8/internal/oracleparser/parser_test.go#L58)):*
> // EnterUpdate_statement implements antlr.SomeInterface

We use comments like that to explain away odd names, unused parameters, or other method signatures that would be surprising to the novice reader.
___
*[`internal/oracleparser/README.md` line 21 at r3](https://reviewable.io/reviews/cockroachdb/replicator/960#gh-1719947806:-O4Qk8SbCPWMBns9XnzV:b7rbnr1) ([raw file](https://github.com/cockroachdb/replicator/blob/141d0c4bb3654cf3ba4e70e18c4c9225a3d1bcf8/internal/oracleparser/README.md#L21)):*
<details><summary><i>Previously, Jeremyyang920 (Jeremy Yang) wrote…</i></summary><blockquote>

Only when you make a change the the actual parser
</blockquote></details>

So, probably almost never?
___
*[`internal/oracleparser/testdata/insert_antrl.ddt` line 63 at r3](https://reviewable.io/reviews/cockroachdb/replicator/960#gh-1719951391:-O4QkLsw9NgB7bHd3RbU:b-71tf3i) ([raw file](https://github.com/cockroachdb/replicator/blob/141d0c4bb3654cf3ba4e70e18c4c9225a3d1bcf8/internal/oracleparser/testdata/insert_antrl.ddt#L63)):*
<details><summary><i>Previously, ZhouXing19 (Jane Xing) wrote…</i></summary><blockquote>

actually log miner won't output subqueries in the logs, i'm just trying to be universal here. the logs that logminer gives will only give the actual values to be updated / inserted. 
</blockquote></details>

Replicator only processes logical records and doesn't generally deal with expressions as the unit of work.  If we were to find cases where logminer outputs expressions, the Oracle frontend would need to evaluate it or configure the apply stack to generate SQL that calls a UDF if it's something dynamic.
___
*[`internal/oracleparser/testdata/insert_antrl.ddt` line 212 at r3](https://reviewable.io/reviews/cockroachdb/replicator/960#gh-1719953388:-O4QlA5UAlRPxzSVMg-c:b-ouq5o5) ([raw file](https://github.com/cockroachdb/replicator/blob/141d0c4bb3654cf3ba4e70e18c4c9225a3d1bcf8/internal/oracleparser/testdata/insert_antrl.ddt#L212)):*

Really exhaustive testing here! Great work on the test cases. Another thought I had was to support collection types like the VARRAY or an ARRAY-LIKE type:
https://docs.oracle.com/en/database/oracle/oracle-database/21/adobj/collection-data-types.html#GUID-7AA98D8C-EA70-43AB-BFC8-724186D1E9B4
</blockquote></details>

Agree with the above.  Please also add a couple of high-magnitude `NUMBER` examples as well, to ensure that things don't accidentally get `float`-ed at some point in the stack.

Jeremyyang920 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/960)* status: 2 of 18 files reviewed, 8 unresolved discussions (waiting on @ryanluu12345 and @ZhouXing19)

___
*[`internal/oracleparser/README.md` line 21 at r3](https://reviewable.io/reviews/cockroachdb/replicator/960#gh-1719947806:-O4QnYML3RzjOo-e-I3k:bwq2im1) ([raw file](https://github.com/cockroachdb/replicator/blob/141d0c4bb3654cf3ba4e70e18c4c9225a3d1bcf8/internal/oracleparser/README.md#L21)):*

So, probably almost never?
</blockquote></details>

For the parsing in replicator most likely never. We've done modifications to the DDL statement grammar in the schema conversion tool, but DML statements are probably more stable.

ZhouXing19:
*[`internal/oracleparser/parser_test.go` line 34 at r3](https://reviewable.io/reviews/cockroachdb/replicator/960#-O4QjTyEDtPU2InidHHb:-O4QqVZ45qTaEHO8mueR:b-lnej8a) ([raw file](https://github.com/cockroachdb/replicator/blob/141d0c4bb3654cf3ba4e70e18c4c9225a3d1bcf8/internal/oracleparser/parser_test.go#L34)):*

> var _ antlr.SomeInterface = (*mockListener)(nil)

is typical to declare conformation to an interface.
</blockquote></details>

We might not want to do this because otherwise we will have to implement all interfaces (for all other stmt types as well) but the only key method we will call are just for insert and update here.

ZhouXing19 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/960)* status: 0 of 18 files reviewed, 8 unresolved discussions (waiting on @bobvawter, @Jeremyyang920, and @ryanluu12345)

___
*[`internal/oracleparser/parser_test.go` line 58 at r3](https://reviewable.io/reviews/cockroachdb/replicator/960#gh-1719949251:-O4Qqr_F1N7X_PLteSHP:b-896fix) ([raw file](https://github.com/cockroachdb/replicator/blob/141d0c4bb3654cf3ba4e70e18c4c9225a3d1bcf8/internal/oracleparser/parser_test.go#L58)):*

> // EnterUpdate_statement implements antlr.SomeInterface

We use comments like that to explain away odd names, unused parameters, or other method signatures that would be surprising to the novice reader.
</blockquote></details>

Done.
___
*[`internal/oracleparser/parser_test.go` line 103 at r3](https://reviewable.io/reviews/cockroachdb/replicator/960#gh-1719977009:-O4QqsJK-xPy5aBFDB01:b-896fix) ([raw file](https://github.com/cockroachdb/replicator/blob/141d0c4bb3654cf3ba4e70e18c4c9225a3d1bcf8/internal/oracleparser/parser_test.go#L103)):*
<details><summary><i>Previously, Jeremyyang920 (Jeremy Yang) wrote…</i></summary><blockquote>

Oops missed this on my other pass, but same backtick comment here for the " escape.
</blockquote></details>

Done.
___
*[`internal/oracleparser/testdata/insert_antrl.ddt` line 212 at r3](https://reviewable.io/reviews/cockroachdb/replicator/960#gh-1719953388:-O4QrnA_1e7TqRNiIIJO:b-896fix) ([raw file](https://github.com/cockroachdb/replicator/blob/141d0c4bb3654cf3ba4e70e18c4c9225a3d1bcf8/internal/oracleparser/testdata/insert_antrl.ddt#L212)):*

Agree with the above.  Please also add a couple of high-magnitude `NUMBER` examples as well, to ensure that things don't accidentally get `float`-ed at some point in the stack.
</blockquote></details>

Done.

ZhouXing19:
*[`internal/oracleparser/testdata/insert_antrl.ddt` line 63 at r3](https://reviewable.io/reviews/cockroachdb/replicator/960#gh-1719951391:-O4QsZyBEJtfh2aAhMYq:b-on6zoh) ([raw file](https://github.com/cockroachdb/replicator/blob/141d0c4bb3654cf3ba4e70e18c4c9225a3d1bcf8/internal/oracleparser/testdata/insert_antrl.ddt#L63)):*

Replicator only processes logical records and doesn't generally deal with expressions as the unit of work.  If we were to find cases where logminer outputs expressions, the Oracle frontend would need to evaluate it or configure the apply stack to generate SQL that calls a UDF if it's something dynamic.
</blockquote></details>

Added https://github.com/cockroachdb/replicator/issues/962 for tracking

PR #973, opened 2024-08-22
Replicator: Support fan-in and wide-deletion use cases

bobvawter:
This change allows Replicator to support use cases where incoming table names
and primary-key structures do not align with the target database schema. This
is of particular utility for migrations into CockroachDB from partitioned or
sharded tables in a replication source. Replicator no longer requires that
incoming table names exist in the target schema.

The existing dispatch() and deletesTo() callbacks are used to support fan-in.
The crux of the implementation change is to split the script-processing path to
apply configureSource() policies before mutations are received by the
underlying sequencer. In the fully-consistent case, dispatch functions are
applied before mutations are staged. The code in script.sourceAcceptor and
script.targetAcceptor is essentially unchanged from the previous
script.acceptor implementation.

In order to allow the deletesTo() callback access to "wide" delete
notifications, such that PostgreSQL REPLICA IDENTITY FULL, might generate, the
core Mutation type gains a Deletion field. Thus, deletions may have an
associated payload body from which PK columns can be extracted once the
deletion has been assigned to a target table. The existing behavior of
converting a key-only deletion to a document is preserved in cases where the
source table name matches a target table name. As a final fallback, the key
array will be presented to the deletesTo() function in a stub document, thus
allowing all deletion types to be dispatched.

The seqtest.Check helper has a scenario for generating mutations for tables
that do not exist in the target database. Similar partition cases are added to
other logical-replication frontends.

The staging package will now persist a mutation's deletion flag in a new table
column. An automatic schema migration is added.

The apply package will prefer to extract the deletion key from the body of a
deletion-type mutation, if possible. Otherwise, it will use the existing
Mutation.Key field.

The switcher sequencer no longer installs the script sequencer on its own in
order to improve testing usability.

The cdc/handler_test.go file is updated to reflect that the dispatch functions
can rewrite a mutation before it is staged.

The logical_test.ts script now strips any incoming _P_ suffix from incoming
table names in order to support partitioned-input tests.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/973)

### DIFF
--- a/internal/script/applier.go
+++ b/internal/script/applier.go
@@ -187,8 +187,7 @@ func opsIntoBatch(ctx context.Context, ops []*applyOp, batch *types.TableBatch) 
 		}
 		switch applyAction(strings.ToLower(op.Action)) {
 		case actionDelete:
-			// Absence of data means delete.
-			batch.Data[idx].Data = nil
+			batch.Data[idx].Deletion = true
 		case actionUpsert:
 			// OK.
 		default:
--- a/internal/sequencer/script/script.go
+++ b/internal/sequencer/script/script.go
@@ -73,7 +73,7 @@ func (w *wrapper) Start(
 	}
 
 	// Only inject if the source or any tables have a configuration.
-	_, inject := scr.Sources.Get(opts.Group.Name)
+	sourceBindings, inject := scr.Sources.Get(opts.Group.Name)
 	if !inject {
 		for _, tbl := range opts.Group.Tables {
 			_, inject = scr.Targets.Get(tbl)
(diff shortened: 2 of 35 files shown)


internal/script/applier.go
```diff
@@ -187,8 +187,7 @@ func opsIntoBatch(ctx context.Context, ops []*applyOp, batch *types.TableBatch)
 		}
 		switch applyAction(strings.ToLower(op.Action)) {
 		case actionDelete:
-			// Absence of data means delete.
-			batch.Data[idx].Data = nil
+			batch.Data[idx].Deletion = true
```
Jeremyyang920, line 190: Just for my understanding, what exactly does this change do in this context? Is it just making it so all deletions will pass through?

bobvawter: Previously, if a mutation's `Data` field were either `nil` or `[]byte("null")`, it would be treated as a deletion (the model was based on CDC changefeeds and then adapted to other sources).  This specific change updates to the new data model where deletions are explicitly tagged as such, regardless of whether the mutation has a data blob.

internal/sequencer/script/script.go
```diff
@@ -82,32 +82,58 @@ func (w *wrapper) Start(
...
-		opts.Delegate = types.UnorderedAcceptorFrom(&acceptor{
-			delegate:   opts.Delegate,
-			ensureTX:   ensureTX,
-			group:      opts.Group,
-			targetPool: w.targetPool,
-			userScript: scr,
-		})
+	if !inject {
+		return w.delegate.Start(ctx, opts)
+	}
```
Jeremyyang920, line 87: Does it make sense to put this return after the for loop in the same !inject block above since its the same if condition?

internal/types/types.go
```diff
@@ -130,18 +130,31 @@ const CustomUpsert = "upsert.custom"
...
 
 var nullBytes = []byte("null")
 
+// HasData returns true if the Data field is populated.
+func (m *Mutation) HasData() bool {
+	switch len(m.Data) {
+	case 0:
+		return false
+	case 4:
+		return !bytes.Equal(m.Data, nullBytes)
```
Jeremyyang920, line 149: Will this be a problem is the actual m.Data is the string null? We've seen some cases of this in Fetch where people for some reason use string null as a value instead of null value.

Jeremyyang920 (approved):
Contextually changes look good to me but will defer for Bram and Silvano with more context of all the changed parts.

internal/sequencer/script/script.go
bobvawter: The inject variable is reassigned within the loop if there is at least once target table with defined handlers.

internal/types/types.go
bobvawter: The data field is otherwise an object, it represents the contents of an entire row or a delta. A null row therefore represents the absence/deletion of the row.

BramGruneir (approved):
Every time we change the userscript, I think we really need to add the dryrun option we've talked about.  Pass in data via csv, return sql, so we can verify that the resulting sql is as expected.  Besides the usefulness in debugging while creating a user script, It would enable users to incorporate Replicator into their CI/CD and test new releases before deploying them.


:lgtm:

> Reviewed 33 of 33 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/973#-:-O5-1Toa-EIgU-bIOEzc:botr5mh)* status: all files reviewed, 4 unresolved discussions (waiting on @bobvawter, @Jeremyyang920, and @sravotto)

___
*[`internal/script/applier.go` line 190 at r1](https://reviewable.io/reviews/cockroachdb/replicator/973#gh-1729102804:-O4zx8_K4H7yU55s8PtL:b-w2jikl) ([raw file](https://github.com/cockroachdb/replicator/blob/8e0c71a435d2eb714e9b21dce0e82cdec576414e/internal/script/applier.go#L190)):*

Previously, if a mutation's `Data` field were either `nil` or `[]byte("null")`, it would be treated as a deletion (the model was based on CDC changefeeds and then adapted to other sources).  This specific change updates to the new data model where deletions are explicitly tagged as such, regardless of whether the mutation has a data blob. We want to preserve the existing data blob to make it available later to the userscript dispatch functions.
</blockquote></details>

Will this break previous userscripts that were using the nil or "null"?
___
*[`internal/script/testdata/main.ts` line 49 at r1](https://reviewable.io/reviews/cockroachdb/replicator/973#-O5-2YaR1ThY7LOxc9u4:-O5-2YaR1ThY7LOxc9u5:b-r9vfyl) ([raw file](https://github.com/cockroachdb/replicator/blob/8e0c71a435d2eb714e9b21dce0e82cdec576414e/internal/script/testdata/main.ts#L49)):*
> ```XML
>                 };
>             case "not_in_target_schema":
>                 // This demonstrates how an incoming from a mutation
> ```

nit: grammar

sravotto (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/973#-:-O5-_6de1i2LXCA7MEN8:bnfp4nl)* status: all files reviewed, 4 unresolved discussions (waiting on @bobvawter and @Jeremyyang920)

bobvawter (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/973)* status: all files reviewed, 3 unresolved discussions (waiting on @Jeremyyang920)

___
*[`internal/script/applier.go` line 190 at r1](https://reviewable.io/reviews/cockroachdb/replicator/973#gh-1729102804:-O5-b62M4BKaN99aowcA:bfcewk2) ([raw file](https://github.com/cockroachdb/replicator/blob/8e0c71a435d2eb714e9b21dce0e82cdec576414e/internal/script/applier.go#L190)):*

Will this break previous userscripts that were using the nil or "null"?
</blockquote></details>

No, because that `nil` or "null" value would never have been presented to the userscript to begin with.
___
*[`internal/script/testdata/main.ts` line 49 at r1](https://reviewable.io/reviews/cockroachdb/replicator/973#-O5-2YaR1ThY7LOxc9u4:-O5-u0SV-qJ1-xjmEqgb:b-896fix) ([raw file](https://github.com/cockroachdb/replicator/blob/8e0c71a435d2eb714e9b21dce0e82cdec576414e/internal/script/testdata/main.ts#L49)):*

nit: grammar
</blockquote></details>

Done.

PR #978, opened 2024-08-27
sequencer: Improve handling of multi-table deletes

bobvawter:
Previously, Replicator had trouble with correctly ordering cross-table
deletions due to orderedAdaptor not having sufficiently complex logic. This
change improves the workload checker to generate scenarios that involve
coordinated deletions between parent and child tables.

The bulk of this change is driven by the need to have BestEffort operate on
groups of tables that have foreign key relationships. Instead of starting a
core sequencer per target table, one is now started per connected group of
tables. This provides an opportunity for mutations on related tables to be
swept and applied atomically.

The core sequencer is updated with additional options to support BestEffort
mode. Rather than deferring entire batches of mutations to be re-attempted
later, a new partial-commit behavior is added. This uses savepoints in the
target transaction to attempt to apply individual mutations. The poisonSet
helper gains additional methods to operate on individual mutations. Ihe
BestEffort configuration will skip over mutations in the staging table, so an
IdleExit flag is added to restart the sweep process once all staged data has
been read.

The sequencer checker can be configured to send a percentage of the generated
mutations to the staging tables, instead of to the sequencer. This simulates
another Replicator process inserting data into staging. The generated mutation
batches are further fragmented to provide more opportunities to either stage or
accept the data.

Sequencers now obtain a lease for each table they are configured to operate on.
The leases package receives a small update to accept a slice of named leases to
be acquired atomically. This finer-grained leasing allows multiple instances of
Replicator to better coordinate their activities around non-intersecting groups
of related tables.

Computing the related-table groups requries building a digraph of parent to
child relationships. The schema inspection queries are simplified and the work
of computing dependecy order is moved into the existing SchemaData and new
SchemaComponent types. An unused method on the watcher type is removed.

The orderedAdapter type continues to aggregate mutations, but will now flush
the accumulator if deletions are encountered in the next batch to consume. The
deletions are applied in reverse dependency order before proceeding with
upserts.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/978)


bobvawter:
Closing in favor of simpler approach in #979 .

PR #979, opened 2024-08-28
besteffort: Improve handling of cascading deletions

bobvawter:
The BestEffort sequencer is simplified such that it starts multiple instances
of its delegate sequencer, one for each group of referential tables. This
solves an otherwise intractable ordering problem, where cascading deletions
across multiple tables with foreign key constraints must act as a
synchronization point for those tables.  Tables with no FK relationships retain
a direct-write path that will fall back to staging as necessary.

Sequencer leases are now acquired for each target table that a core sequencer
will operate on. This will allow multiple Replicator instances to unstage and
apply mutations from disjoint groups of tables, which can spread the BestEffort
load.

The workload checker is updated to generate transactions in which parent and
child rows are deleted within the same transaction. Writing test mutations to
staging is now expressed as a percentage of mutations to write as though
another Replicator instance were receiving mutations. The previous
besteffort.TestStepByStep() function is entirely redundant with the sequencer
checker and has been removed.

This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/979)


BramGruneir (approved):
:lgtm:

This is a significant improvement over what we had before and I think is going to be the final form of BestEffort.  Is it possible to  distribute the leases on table groups amongst replicator instances?  Then it would get a significant throughput boost.  

> Reviewed 25 of 25 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/979#-:-O6748wVDMsBy9ZH1ta6:b-319yx7)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Jeremyyang920 and @sravotto)

PR #980, opened 2024-08-28
Add a schema backup/restore mechanism

MattWhelan:
Store the schema to the staging memo, and restore it when the target database is down on startup.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/980)

### DIFF
--- a/internal/target/schemawatch/backup.go
+++ b/internal/target/schemawatch/backup.go
@@ -1,0 +1,104 @@
+// Copyright 2024 The Cockroach Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package schemawatch
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/cockroachdb/field-eng-powertools/notify"
+	"github.com/cockroachdb/field-eng-powertools/stopper"
+	"github.com/cockroachdb/field-eng-powertools/stopvar"
+	"github.com/cockroachdb/replicator/internal/types"
+	"github.com/cockroachdb/replicator/internal/util/ident"
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+)
+
+// Backup handles schema backups and restores. It allows staging
+// progress when the target database is down.
+type Backup interface {
+	backup(ctx *stopper.Context, schema ident.Schema, data *types.SchemaData) error
+	restore(ctx *stopper.Context, schema ident.Schema) (*types.SchemaData, error)
+	startUpdates(
+		ctx *stopper.Context,
+		schemaVar *notify.Var[*types.SchemaData],
+		schema ident.Schema,
+	)
+}
+
+// memoBackup implements Backup and is backed by the staging table memo
+// store
+type memoBackup struct {
+	memo        types.Memo
+	stagingPool *types.StagingPool
+}
+
+var _ Backup = (*memoBackup)(nil)
+
+// backup backs up a schema
+func (b *memoBackup) backup(
+	ctx *stopper.Context, schema ident.Schema, data *types.SchemaData,
+) error {
+	dataBytes, err := json.Marshal(data)
+	if err != nil {
+		return errors.Wrap(err, "could not marshal schema data")
+	}
+	err = b.memo.Put(ctx, b.stagingPool, b.memoKey(schema), dataBytes)
+	if err != nil {
+		return errors.Wrap(err, "couldn't save schema data")
+	}
+	return nil
+}
+
+// memoKey constructs a key for a schema memo
+func (b *memoBackup) memoKey(schema ident.Schema) string {
+	return fmt.Sprintf("schema-%s", schema.Canonical().Raw())
+}
+
+// restore fetches the latest schema backup
+func (b *memoBackup) restore(ctx *stopper.Context, schema ident.Schema) (*types.SchemaData, error) {
+	schemaBytes, err := b.memo.Get(ctx, b.stagingPool, b.memoKey(schema))
+	if err != nil {
+		return nil, err
+	}
+	data := &types.SchemaData{}
+	err = json.Unmarshal(schemaBytes, data)
+	if err != nil {
+		return nil, err
+	}
+	return data, nil
+}
+
+// startUpdates watches the supplied schemaVar for changes and updates
+// the backup. It records the initial value of schemaVar as the first "change".
+func (b *memoBackup) startUpdates(
+	ctx *stopper.Context, schemaVar *notify.Var[*types.SchemaData], schema ident.Schema,
+) {
+	// watch schemaVar and write changes as a staging memo
+	ctx.Go(func(ctx *stopper.Context) error {
+		_, err := stopvar.DoWhenChanged(ctx, nil, schemaVar, func(ctx *stopper.Context, _, d *types.SchemaData) error {
+			err := b.backup(ctx, schema, d)
+			if err != nil {
+				// Log the error, but don't exit the loop
+				log.WithError(err).WithField("target", schema).Warn("failed to backup schema data")
+			}
+			return nil
+		})
+		return err
+	})
+}
--- a/internal/target/schemawatch/factory.go
+++ b/internal/target/schemawatch/factory.go
@@ -28,9 +28,10 @@ import (
 
 // factory is a memoizing factory for watcher instances.
 type factory struct {
-	pool *types.TargetPool
-	stop *stopper.Context
-	mu   struct {
+	backup Backup
+	pool   *types.TargetPool
+	stop   *stopper.Context
+	mu     struct {
 		sync.RWMutex
 		data *ident.SchemaMap[*watcher]
 	}
@@ -72,7 +73,7 @@ func (f *factory) createUnlocked(db ident.Schema) (*watcher, error) {
 		return ret, nil
 	}
 
-	ret, err := newWatcher(f.stop, f.pool, db)
+	ret, err := newWatcher(f.stop, f.pool, db, f.backup)
 	if err != nil {
 		return nil, err
 	}
--- a/internal/target/schemawatch/watcher.go
+++ b/internal/target/schemawatch/watcher.go
@@ -23,6 +23,7 @@ import (
 	"database/sql"
 	"flag"
 	"fmt"
+	"reflect"
 	"time"
 
 	"github.com/cockroachdb/field-eng-powertools/notify"
@@ -53,18 +54,38 @@ var _ types.Watcher = (*watcher)(nil)
 // newWatcher constructs a new watcher to monitor the table schema in the
 // named database. The returned watcher will internally refresh
 // until the cancel callback is executed.
-func newWatcher(ctx *stopper.Context, tx *types.TargetPool, schema ident.Schema) (*watcher, error) {
+func newWatcher(
+	ctx *stopper.Context, tx *types.TargetPool, schema ident.Schema, b Backup,
+) (*watcher, error) {
 	w := &watcher{
 		delay:  *RefreshDelay,
 		schema: schema,
 	}
 
-	// Initial data load to sanity-check and make ready.
-	data, err := w.getTables(ctx, tx)
-	if err != nil {
-		return nil, err
+	var data *types.SchemaData
+	if err := tx.Ping(); err != nil {
+		log.WithError(err).Warn("failed to ping target database; trying to restore backup")
+		// On start up, when the target database is down, fall back to
+		// the staging memo about the table schema
+		data, err = b.restore(ctx, schema)
+		if err != nil {
+			return nil, err
+		}
+		if data == nil {
+			// Restore saw no errors, but we also didn't have a value
+			return nil, errors.Wrapf(err, "no backup schema data for %s", schema)
+		}
+	} else {
+		// Initial data load to sanity-check and make ready.
+		data, err = w.getTables(ctx, tx)
+		if err != nil {
+			return nil, err
+		}
 	}
+
 	w.data = notify.VarOf(data)
+
+	b.startUpdates(ctx, w.data, schema)
 
 	if w.delay > 0 {
 		ctx.Go(func(ctx *stopper.Context) error {
@@ -99,7 +120,10 @@ func (w *watcher) Refresh(ctx context.Context, tx *types.TargetPool) error {
 		return err
 	}
 
-	w.data.Set(next)
+	prev, _ := w.data.Get()
+	if !reflect.DeepEqual(prev, next) {
+		w.data.Set(next)
+	}
 	return nil
 }
 
--- a/internal/sinktest/all/wire_gen.go
+++ b/internal/sinktest/all/wire_gen.go
@@ -71,7 +71,12 @@ func NewFixture(t testing.TB) (*Fixture, error) {
 	if err != nil {
 		return nil, err
 	}
-	watchers, err := schemawatch.ProvideFactory(context, targetPool, diagnostics)
+	memoMemo, err := memo.ProvideMemo(context, stagingPool, stagingSchema)
+	if err != nil {
+		return nil, err
+	}
+	backup := schemawatch.ProvideBackup(memoMemo, stagingPool)
+	watchers, err := schemawatch.ProvideFactory(context, targetPool, diagnostics, backup)
 	if err != nil {
 		return nil, err
 	}
@@ -85,10 +90,6 @@ func NewFixture(t testing.TB) (*Fixture, error) {
 		return nil, err
 	}
 	checkpoints, err := checkpoint.ProvideCheckpoints(context, stagingPool, stagingSchema)
-	if err != nil {
-		return nil, err
-	}
-	memoMemo, err := memo.ProvideMemo(context, stagingPool, stagingSchema)
 	if err != nil {
 		return nil, err
 	}
--- a/internal/source/cdc/server/test_fixture.go
+++ b/internal/source/cdc/server/test_fixture.go
@@ -36,6 +36,7 @@ type testFixture struct {
 	Config        *Config
 	Diagnostics   *diag.Diagnostics
 	Listener      net.Listener
+	Memo          types.Memo
 	StagingPool   *types.StagingPool
 	Server        *stdserver.Server
 	StagingDB     ident.StagingSchema
--- a/internal/source/cdc/server/wire_gen.go
+++ b/internal/source/cdc/server/wire_gen.go
@@ -90,7 +90,8 @@ func NewServer(ctx *stopper.Context, config *Config) (*Server, error) {
 		return nil, err
 	}
 	dlqConfig := cdc.ProvideDLQConfig(cdcConfig)
-	watchers, err := schemawatch.ProvideFactory(ctx, targetPool, diagnostics)
+	backup := schemawatch.ProvideBackup(memoMemo, stagingPool)
+	watchers, err := schemawatch.ProvideFactory(ctx, targetPool, diagnostics, backup)
 	if err != nil {
 		return nil, err
 	}
@@ -175,11 +176,11 @@ func newTestFixture(context *stopper.Context, config *Config) (*testFixture, fun
 	if err != nil {
 		return nil, nil, err
 	}
-	cdcConfig := &config.CDC
 	memoMemo, err := memo.ProvideMemo(context, stagingPool, stagingSchema)
 	if err != nil {
 		return nil, nil, err
 	}
+	cdcConfig := &config.CDC
 	checker := version.ProvideChecker(stagingPool, memoMemo)
 	targetPool, err := sinkprod.ProvideTargetPool(context, checker, targetConfig, diagnostics)
 	if err != nil {
@@ -194,7 +195,8 @@ func newTestFixture(context *stopper.Context, config *Config) (*testFixture, fun
 		return nil, nil, err
 	}
 	dlqConfig := cdc.ProvideDLQConfig(cdcConfig)
-	watchers, err := schemawatch.ProvideFactory(context, targetPool, diagnostics)
+	backup := schemawatch.ProvideBackup(memoMemo, stagingPool)
+	watchers, err := schemawatch.ProvideFactory(context, targetPool, diagnostics, backup)
 	if err != nil {
 		return nil, nil, err
 	}
@@ -255,6 +257,7 @@ func newTestFixture(context *stopper.Context, config *Config) (*testFixture, fun
 		Config:        config,
 		Diagnostics:   diagnostics,
 		Listener:      listener,
+		Memo:          memoMemo,
 		StagingPool:   stagingPool,
 		Server:        server,
 		StagingDB:     stagingSchema,
@@ -278,6 +281,7 @@ type testFixture struct {
 	Config        *Config
 	Diagnostics   *diag.Diagnostics
 	Listener      net.Listener
+	Memo          types.Memo
 	StagingPool   *types.StagingPool
 	Server        *stdserver.Server
 	StagingDB     ident.StagingSchema
--- a/internal/source/cdc/test_fixture.go
+++ b/internal/source/cdc/test_fixture.go
@@ -51,7 +51,7 @@ func newTestFixture(*all.Fixture, *Config) (*testFixture, error) {
 		wire.FieldsOf(new(*base.Fixture),
 			"Context", "StagingDB", "StagingPool", "TargetCache", "TargetPool"),
 		wire.FieldsOf(new(*all.Fixture),
-			"Configs", "Fixture", "Stagers"),
+			"Configs", "Fixture", "Stagers", "Memo"),
 		diag.New,
 		leases.Set,
 		checkpoint.Set,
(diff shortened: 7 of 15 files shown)


internal/target/schemawatch/backup.go
```diff
@@ -0,0 +1,87 @@
...
+		oldValue, ch := schemaVar.Get()
+		for {
+			select {
+			case <-ctx.Stopping():
+				return nil
+			case <-ch:
+				var d *types.SchemaData
+				d, ch = schemaVar.Get()
+				// TODO: Is change detection a good idea here, or is that handled upstream?
+				if !reflect.DeepEqual(d, oldValue) {
```
MattWhelan, line 70: Question: is this worth doing? It doesn't seem like there's any upstream debouncing.

internal/target/schemawatch/factory.go
```diff
@@ -72,7 +74,11 @@ func (f *factory) createUnlocked(db ident.Schema) (*watcher, error) {
 		return ret, nil
 	}
 
-	ret, err := newWatcher(f.stop, f.pool, db)
+	b := &memoBackup{
+		memo:        f.memo,
+		stagingPool: f.stagingPool,
+	}
```
MattWhelan, line 80: Should I construct this this way, or should I inject it further upstream, with its own provider, etc?

internal/target/schemawatch/watcher.go
```diff
@@ -53,19 +53,31 @@ var _ types.Watcher = (*watcher)(nil)
...
 	}
 
 	// Initial data load to sanity-check and make ready.
 	data, err := w.getTables(ctx, tx)
+
+	// On start up, when the target database is down, fall back to the staging memo about the table schema
 	if err != nil {
-		return nil, err
+		// TODO: detect whether the error is because the target DB is down, or something else
+		if !isDatabaseDown(err) {
```
MattWhelan, line 68: This is the big one: how do we differentiate between fatal and non-fatal errors.

bobvawter (changes requested):
General stye note: Please wrap code and doc comments at 72 columns (the "Wrap to Columns" plugin for GoLand  is what I use).  Also, target 100 columns for code.

> Reviewed 8 of 14 files at r2.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/980#-:-O5TNck2BpJYaPQF1plu:btbzmks)* status: 8 of 14 files reviewed, 9 unresolved discussions (waiting on @MattWhelan)

___
*[`internal/target/schemawatch/backup.go` line 70 at r1](https://reviewable.io/reviews/cockroachdb/replicator/980#gh-1735145530:-O5TE5EqBhdPDIbQ8Eq7:b67xfem) ([raw file](https://github.com/cockroachdb/replicator/blob/b6ef084c94d6cad0403f58fe8ebfe238506dfbc2/internal/target/schemawatch/backup.go#L70)):*
<details><summary><i>Previously, MattWhelan (Matt Whelan) wrote…</i></summary><blockquote>

Question: is this worth doing? It doesn't seem like there's any upstream debouncing.
</blockquote></details>

That top-level var had never been exposed directly. There's debouncing in the `Watch()` method and a utility method to help with the comparisons.  I'd say to add debouncing to the watcher's `refresh()` method.  That would make this storage loop cleaner, you could use `DoWhenChanged` here.

https://pkg.go.dev/github.com/cockroachdb/field-eng-powertools@v0.1.1/stopvar#DoWhenChanged
___
*[`internal/target/schemawatch/backup.go` line 50 at r2](https://reviewable.io/reviews/cockroachdb/replicator/980#-O5TFPPo1IRLkGBNRbsn:-O5TFPPo1IRLkGBNRbso:b-kxn7y3) ([raw file](https://github.com/cockroachdb/replicator/blob/9f778b8ab8355a73ebf811b66e5ee513e1060e5f/internal/target/schemawatch/backup.go#L50)):*
> ```Go
> 	dataBytes, err := json.Marshal(data)
> 	if err != nil {
> 		return fmt.Errorf("marshalling schema data: %w", err)
> ```

Replicator uses `github.com/pkg/errors`, because it provides stack traces.  In this case `errors.Wrap(err, "could not marshal schema data")`

The general rule is to call either `Wrap()` or `WithStack()` when calling external APIs and assume that any Replicator APIs will already return decorated errors.
___
*[`internal/target/schemawatch/backup.go` line 61 at r2](https://reviewable.io/reviews/cockroachdb/replicator/980#-O5TGAwTBwcdAb3rezTn:-O5TGAwTBwcdAb3rezTo:bpxboz6) ([raw file](https://github.com/cockroachdb/replicator/blob/9f778b8ab8355a73ebf811b66e5ee513e1060e5f/internal/target/schemawatch/backup.go#L61)):*
> ```Go
> // memoKey constructs a
> func (b *memoBackup) memoKey(schema ident.Schema) string {
> 	return fmt.Sprintf("schema-%v", schema)
> ```

Prefer `fmt.Sprintf("schema-", schema.Canonical().Raw())` for computing externalized keys that don't need to be re-parsed.
___
*[`internal/target/schemawatch/backup_test.go` line 75 at r2](https://reviewable.io/reviews/cockroachdb/replicator/980#-O5TH_829uHNQGj8nELQ:-O5TH_829uHNQGj8nELR:bbiv4li) ([raw file](https://github.com/cockroachdb/replicator/blob/9f778b8ab8355a73ebf811b66e5ee513e1060e5f/internal/target/schemawatch/backup_test.go#L75)):*
> ```Go
> 
> 	// await the backup call
> 	time.Sleep(100 * time.Millisecond)
> ```

Sleep-based tests are going to be flaky on the CI system.

The general pattern in Replicator for stateless "something happened" notifications would be a `notify.Var`. 

Since this is only for testing, and the test is constructing the instance, what I'd suggest is that you add a notification field to `memoBackup` that is set by the test and which is notified when a new schema configuration has been persisted.  In your `startUpdates()` method:

```
// onStore *notify.Var[int] is a generation counter
if b.onStore != nil {
  b.onStore.Update(func(old int) (int, error) { return old + 1, nil })
}
```

Then, in your test code, you can monitor the value of the `onStore` variable.  Since it's a counter, you could also verify how many times the persistence code actually fires to validate debouncing.
___
*[`internal/target/schemawatch/provider.go` line 30 at r2](https://reviewable.io/reviews/cockroachdb/replicator/980#-O5TJryj-9nT3VIq45l6:-O5TJryj-9nT3VIq45l7:b-eu97i7) ([raw file](https://github.com/cockroachdb/replicator/blob/9f778b8ab8355a73ebf811b66e5ee513e1060e5f/internal/target/schemawatch/provider.go#L30)):*
> ```Go
> var Set = wire.NewSet(
> 	ProvideFactory,
> 	ProvideBackup,
> ```

Please keep runs of things in sorted order, it minimizes diff thrash.
___
*[`internal/target/schemawatch/provider.go` line 46 at r2](https://reviewable.io/reviews/cockroachdb/replicator/980#-O5TKCTxCLl1Ttvs5EiX:-O5TKCTxCLl1Ttvs5EiY:b93vcns) ([raw file](https://github.com/cockroachdb/replicator/blob/9f778b8ab8355a73ebf811b66e5ee513e1060e5f/internal/target/schemawatch/provider.go#L46)):*
> ```Go
> 
> // ProvideBackup provides the Backup service used by Watcher
> func ProvideBackup(m types.Memo, stagingPool *types.StagingPool) Backup {
> ```

It would actually be fine to drop the `Backup` interface entirely and just call the `memoBackup` type `Backup`.  This provider would return `*Backup`.

The `types` package is a bit of a historical artifact when I reworked the PoC code that Bram wrote.  I needed to forward-declare a bunch of types to help when I was refactoring.  It was also nice to have a single package to look at to identify all major functional elements of the system.  Having said that, I'm trying to move away from interfaces and just use pointers to implementations in cases where there's no actual requirement for an interface.  This package was one of the first things extracted from the PoC code and it's not quite as Wired as the rest of the codebase is.
___
*[`internal/target/schemawatch/watcher.go` line 68 at r1](https://reviewable.io/reviews/cockroachdb/replicator/980#gh-1735147155:-O5TLLaQ0tUSoA561iOi:bq9b7vf) ([raw file](https://github.com/cockroachdb/replicator/blob/b6ef084c94d6cad0403f58fe8ebfe238506dfbc2/internal/target/schemawatch/watcher.go#L68)):*
<details><summary><i>Previously, MattWhelan (Matt Whelan) wrote…</i></summary><blockquote>

This is the big one: how do we differentiate between fatal and non-fatal errors.
</blockquote></details>

I'd structure it as follows:

```go
data, getErr := w.getTables(ctx, tx)
if getErr != nil {
  data, restoreErr = b.restore(ctx, schema)
  if restoreErr != nil {
    return nil, errors.Join(
       errors.Wrapf(getErr, "could not load schema %s", schema),
       errors.Wrapf(restoreErr, "could not load backup of schema %s", schema)
    )
   if data == nil /* or however no-data case goes */ {
    return nil, errors.Wrapf(getErr, "no cached schema data available for %s", schema)
   }
}
```
___
*[`internal/target/schemawatch/watcher.go` line 82 at r2](https://reviewable.io/reviews/cockroachdb/replicator/980#-O5TNXD90INwZksQFQVW:-O5TNXD90INwZksQFQVX:b7lxz4f) ([raw file](https://github.com/cockroachdb/replicator/blob/9f778b8ab8355a73ebf811b66e5ee513e1060e5f/internal/target/schemawatch/watcher.go#L82)):*
> ```Go
> 
> 		// Update the backup, since we fetched the schema successfully
> 		err = b.backup(ctx, schema, data)
> ```

Would this be redundant with also calling `startUpdates()` ?

MattWhelan (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/980)* status: 8 of 14 files reviewed, 8 unresolved discussions (waiting on @bobvawter)

___
*[`internal/target/schemawatch/backup.go` line 70 at r1](https://reviewable.io/reviews/cockroachdb/replicator/980#gh-1735145530:-O5Tppp63Oa8N705bZp8:bsd0365) ([raw file](https://github.com/cockroachdb/replicator/blob/b6ef084c94d6cad0403f58fe8ebfe238506dfbc2/internal/target/schemawatch/backup.go#L70)):*

That top-level var had never been exposed directly. There's debouncing in the `Watch()` method and a utility method to help with the comparisons.  I'd say to add debouncing to the watcher's `refresh()` method.  That would make this storage loop cleaner, you could use `DoWhenChanged` here.

https://pkg.go.dev/github.com/cockroachdb/field-eng-powertools@v0.1.1/stopvar#DoWhenChanged
</blockquote></details>

Let me make sure I understand the suggestion and context: modify `watcher.Refresh` to get, compare and conditionally set the SchemaData value, using `reflect.DeepEqual` (since SchemaData doesn't appear to have any other equality operator). Then, use `DoWhenChanged` to implement `startUpdates`, and assume that every time it fires, it's a real change (because we've debounced upstream; DWC is only comparing pointer addresses, so its debouncing implementation isn't valid here).
___
*[`internal/target/schemawatch/backup.go` line 50 at r2](https://reviewable.io/reviews/cockroachdb/replicator/980#-O5TFPPo1IRLkGBNRbsn:-O5TlXWc5w2K_w8EINxx:b-896fix) ([raw file](https://github.com/cockroachdb/replicator/blob/9f778b8ab8355a73ebf811b66e5ee513e1060e5f/internal/target/schemawatch/backup.go#L50)):*

Replicator uses `github.com/pkg/errors`, because it provides stack traces.  In this case `errors.Wrap(err, "could not marshal schema data")`

The general rule is to call either `Wrap()` or `WithStack()` when calling external APIs and assume that any Replicator APIs will already return decorated errors.
</blockquote></details>

Done.
___
*[`internal/target/schemawatch/backup.go` line 61 at r2](https://reviewable.io/reviews/cockroachdb/replicator/980#-O5TGAwTBwcdAb3rezTn:-O5TllDk-xlM_XAsDQy9:b-896fix) ([raw file](https://github.com/cockroachdb/replicator/blob/9f778b8ab8355a73ebf811b66e5ee513e1060e5f/internal/target/schemawatch/backup.go#L61)):*

Prefer `fmt.Sprintf("schema-", schema.Canonical().Raw())` for computing externalized keys that don't need to be re-parsed.
</blockquote></details>

Done.
___
*[`internal/target/schemawatch/backup_test.go` line 75 at r2](https://reviewable.io/reviews/cockroachdb/replicator/980#-O5TH_829uHNQGj8nELQ:-O5Tv3El8VFY2dXKW5x_:b9u6jid) ([raw file](https://github.com/cockroachdb/replicator/blob/9f778b8ab8355a73ebf811b66e5ee513e1060e5f/internal/target/schemawatch/backup_test.go#L75)):*

Sleep-based tests are going to be flaky on the CI system.

The general pattern in Replicator for stateless "something happened" notifications would be a `notify.Var`. 

Since this is only for testing, and the test is constructing the instance, what I'd suggest is that you add a notification field to `memoBackup` that is set by the test and which is notified when a new schema configuration has been persisted.  In your `startUpdates()` method:

```
// onStore *notify.Var[int] is a generation counter
if b.onStore != nil {
  b.onStore.Update(func(old int) (int, error) { return old + 1, nil })
}
```

Then, in your test code, you can monitor the value of the `onStore` variable.  Since it's a counter, you could also verify how many times the persistence code actually fires to validate debouncing.
</blockquote></details>

Any objection to working that into the Memory implementation of Memo?
___
*[`internal/target/schemawatch/provider.go` line 30 at r2](https://reviewable.io/reviews/cockroachdb/replicator/980#-O5TJryj-9nT3VIq45l6:-O5TmJF4D9nMJMMaLepM:b-896fix) ([raw file](https://github.com/cockroachdb/replicator/blob/9f778b8ab8355a73ebf811b66e5ee513e1060e5f/internal/target/schemawatch/provider.go#L30)):*

Please keep runs of things in sorted order, it minimizes diff thrash.
</blockquote></details>

Done.
___
*[`internal/target/schemawatch/provider.go` line 46 at r2](https://reviewable.io/reviews/cockroachdb/replicator/980#-O5TKCTxCLl1Ttvs5EiX:-O5TnG479g2j2z2qXEUW:b-p4dcur) ([raw file](https://github.com/cockroachdb/replicator/blob/9f778b8ab8355a73ebf811b66e5ee513e1060e5f/internal/target/schemawatch/provider.go#L46)):*

It would actually be fine to drop the `Backup` interface entirely and just call the `memoBackup` type `Backup`.  This provider would return `*Backup`.

The `types` package is a bit of a historical artifact when I reworked the PoC code that Bram wrote.  I needed to forward-declare a bunch of types to help when I was refactoring.  It was also nice to have a single package to look at to identify all major functional elements of the system.  Having said that, I'm trying to move away from interfaces and just use pointers to implementations in cases where there's no actual requirement for an interface.  This package was one of the first things extracted from the PoC code and it's not quite as Wired as the rest of the codebase is.
</blockquote></details>

It allows for test mocks as well; I'll leave it unless having an interface actively bothers you.
___
*[`internal/target/schemawatch/watcher.go` line 68 at r1](https://reviewable.io/reviews/cockroachdb/replicator/980#gh-1735147155:-O5TuQ8VAh2_M_9DPq_F:b-292qf7) ([raw file](https://github.com/cockroachdb/replicator/blob/b6ef084c94d6cad0403f58fe8ebfe238506dfbc2/internal/target/schemawatch/watcher.go#L68)):*

I'd structure it as follows:

```go
data, getErr := w.getTables(ctx, tx)
if getErr != nil {
  data, restoreErr = b.restore(ctx, schema)
  if restoreErr != nil {
    return nil, errors.Join(
       errors.Wrapf(getErr, "could not load schema %s", schema),
       errors.Wrapf(restoreErr, "could not load backup of schema %s", schema)
    )
   if data == nil /* or however no-data case goes */ {
    return nil, errors.Wrapf(getErr, "no cached schema data available for %s", schema)
   }
}
```
</blockquote></details>

You dislike the explicit check for connectivity?
___
*[`internal/target/schemawatch/watcher.go` line 82 at r2](https://reviewable.io/reviews/cockroachdb/replicator/980#-O5TNXD90INwZksQFQVW:-O5TmgzR2Zoxp1nvmg6j:bsefmfz) ([raw file](https://github.com/cockroachdb/replicator/blob/9f778b8ab8355a73ebf811b66e5ee513e1060e5f/internal/target/schemawatch/watcher.go#L82)):*

Would this be redundant with also calling `startUpdates()` ?
</blockquote></details>

No. The watcher only responds to changes, and we start out with the initial value already set, so we need this to backup a schema that never changes.

bobvawter (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/980)* status: 8 of 15 files reviewed, 6 unresolved discussions (waiting on @MattWhelan)

___
*[`internal/target/schemawatch/backup.go` line 70 at r1](https://reviewable.io/reviews/cockroachdb/replicator/980#gh-1735145530:-O5URCxs9AylU5P3bob0:bsbcmpi) ([raw file](https://github.com/cockroachdb/replicator/blob/b6ef084c94d6cad0403f58fe8ebfe238506dfbc2/internal/target/schemawatch/backup.go#L70)):*
<details><summary><i>Previously, MattWhelan (Matt Whelan) wrote…</i></summary><blockquote>

Let me make sure I understand the suggestion and context: modify `watcher.Refresh` to get, compare and conditionally set the SchemaData value, using `reflect.DeepEqual` (since SchemaData doesn't appear to have any other equality operator). Then, use `DoWhenChanged` to implement `startUpdates`, and assume that every time it fires, it's a real change (because we've debounced upstream; DWC is only comparing pointer addresses, so its debouncing implementation isn't valid here).
</blockquote></details>

Your understanding is correct.

MattWhelan (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/980)* status: 8 of 15 files reviewed, 5 unresolved discussions (waiting on @bobvawter)

___
*[`internal/target/schemawatch/watcher.go` line 82 at r2](https://reviewable.io/reviews/cockroachdb/replicator/980#-O5TNXD90INwZksQFQVW:-O5Un7Gc-dgikClI4PtJ:b1kldvh) ([raw file](https://github.com/cockroachdb/replicator/blob/9f778b8ab8355a73ebf811b66e5ee513e1060e5f/internal/target/schemawatch/watcher.go#L82)):*
<details><summary><i>Previously, MattWhelan (Matt Whelan) wrote…</i></summary><blockquote>

No. The watcher only responds to changes, and we start out with the initial value already set, so we need this to backup a schema that never changes.
</blockquote></details>

With the changes to the implementation of startUpdates, it became redundant.

bobvawter (changes requested):
> Reviewed 1 of 5 files at r4.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/980)* status: 9 of 15 files reviewed, 2 unresolved discussions (waiting on @MattWhelan)

___
*[`internal/target/schemawatch/backup.go` line 68 at r4](https://reviewable.io/reviews/cockroachdb/replicator/980#-O5YPlNPE_0wls_xU6F6:-O5YPlNPE_0wls_xU6F7:bkmobmn) ([raw file](https://github.com/cockroachdb/replicator/blob/f5f9db05221be00878c4b63b3cc3aa8ff52de170/internal/target/schemawatch/backup.go#L68)):*
> ```Go
> }
> 
> // memoKey constructs a
> ```

Unfinished comment.
___
*[`internal/target/schemawatch/watcher.go` line 68 at r1](https://reviewable.io/reviews/cockroachdb/replicator/980#gh-1735147155:-O5YQp23-C4jMZNYzdM9:bqj3uxr) ([raw file](https://github.com/cockroachdb/replicator/blob/b6ef084c94d6cad0403f58fe8ebfe238506dfbc2/internal/target/schemawatch/watcher.go#L68)):*
<details><summary><i>Previously, MattWhelan (Matt Whelan) wrote…</i></summary><blockquote>

You dislike the explicit check for connectivity?
</blockquote></details>

In hindsight, I'm thinking that the check for connectivity is redundant with attempting to load the schema. A failed ping is equally as informative as a failed SQL query. As currently written it also looks like the error from ping would be discarded?  As I read it, the flow control is `failed ping -> restore -> getTables() -> return err`

MattWhelan (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/980)* status: 9 of 15 files reviewed, 2 unresolved discussions (waiting on @bobvawter)

___
*[`internal/target/schemawatch/backup.go` line 68 at r4](https://reviewable.io/reviews/cockroachdb/replicator/980#-O5YPlNPE_0wls_xU6F6:-O5Z8gISBnfOeRbjlYkE:b-896fix) ([raw file](https://github.com/cockroachdb/replicator/blob/f5f9db05221be00878c4b63b3cc3aa8ff52de170/internal/target/schemawatch/backup.go#L68)):*

Unfinished comment.
</blockquote></details>

Done.
___
*[`internal/target/schemawatch/watcher.go` line 68 at r1](https://reviewable.io/reviews/cockroachdb/replicator/980#gh-1735147155:-O5Z8klZDdlJPd2A-6Mh:b-gv7df6) ([raw file](https://github.com/cockroachdb/replicator/blob/b6ef084c94d6cad0403f58fe8ebfe238506dfbc2/internal/target/schemawatch/watcher.go#L68)):*

In hindsight, I'm thinking that the check for connectivity is redundant with attempting to load the schema. A failed ping is equally as informative as a failed SQL query. As currently written it also looks like the error from ping would be discarded?  As I read it, the flow control is `failed ping -> restore -> getTables() -> return err`
</blockquote></details>

My assumption with the probe is that there exist errors that are still fatal, that we don't want to use a restore to paper over. If that's not the case, then we can totally just restore on all errors. 

The ping error does get discarded, as does any connection acquire error. There's no call to `getTables` if we do the restore, as it's in the else branch. If we can't connect, we know `getTables` can't succeed.

It's `let data = if pingError() { restore() } else { getTables() };`

MattWhelan (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/980)* status: 8 of 15 files reviewed, 2 unresolved discussions (waiting on @bobvawter)

___
*[`internal/target/schemawatch/watcher.go` line 68 at r1](https://reviewable.io/reviews/cockroachdb/replicator/980#gh-1735147155:-O5ZRoX66AXASAE-2ku-:b4dp1kr) ([raw file](https://github.com/cockroachdb/replicator/blob/b6ef084c94d6cad0403f58fe8ebfe238506dfbc2/internal/target/schemawatch/watcher.go#L68)):*
<details><summary><i>Previously, MattWhelan (Matt Whelan) wrote…</i></summary><blockquote>

My assumption with the probe is that there exist errors that are still fatal, that we don't want to use a restore to paper over. If that's not the case, then we can totally just restore on all errors. 

The ping error does get discarded, as does any connection acquire error. There's no call to `getTables` if we do the restore, as it's in the else branch. If we can't connect, we know `getTables` can't succeed.

It's `let data = if pingError() { restore() } else { getTables() };`
</blockquote></details>

So let's say the target database is up, but the target table doesn't exist. It's been removed/renamed, or the target is actually the wrong database host. Is it better to restore the backup and continue, or fail here? The probe strategy will fail if the problem isn't the connection. The other could very well succeed, with whatever consequences that has.

bobvawter (approved):
:lgtm:

> Reviewed 2 of 14 files at r2, 4 of 5 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/980#-:-O5ZjQ6I2h6kVUoK8RRi:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @MattWhelan)

___
*[`internal/target/schemawatch/watcher.go` line 68 at r1](https://reviewable.io/reviews/cockroachdb/replicator/980#gh-1735147155:-O5Zh1fs3_6Zuv8eKZL5:b-ygyyaj) ([raw file](https://github.com/cockroachdb/replicator/blob/b6ef084c94d6cad0403f58fe8ebfe238506dfbc2/internal/target/schemawatch/watcher.go#L68)):*
<details><summary><i>Previously, MattWhelan (Matt Whelan) wrote…</i></summary><blockquote>

So let's say the target database is up, but the target table doesn't exist. It's been removed/renamed, or the target is actually the wrong database host. Is it better to restore the backup and continue, or fail here? The probe strategy will fail if the problem isn't the connection. The other could very well succeed, with whatever consequences that has.
</blockquote></details>

I lost track of the level of indentation; the discussion UI in Reviewable is occupying most of my vertical screen real-estate.  Now that I'm reading what it is that you actually wrote, it makes sense as written.  It would be nice, however, to `log.WithErr().Warn()` the error from acquiring a connection or ping.  Also, the Ping method exists directly on `*sql.DB` so the explicit connection acquisition isn't strictly necessary.

There's always a risk in relying on cached data; the joke about the two most difficult software problems being naming things, cache invalidation, and off-by-one errors applies here.  I'm reasonably confident that the `getTables()` queries will do the right thing if the requested `schema` doesn't actually exist in the target database. They'll return an empty, no-error response.  If they don't it's a bug.

MattWhelan (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/980)* status: 14 of 15 files reviewed, all discussions resolved (waiting on @bobvawter)

___
*[`internal/target/schemawatch/watcher.go` line 68 at r1](https://reviewable.io/reviews/cockroachdb/replicator/980#gh-1735147155:-O5_6-F7EbHYRBuSBPVc:bddwma4) ([raw file](https://github.com/cockroachdb/replicator/blob/b6ef084c94d6cad0403f58fe8ebfe238506dfbc2/internal/target/schemawatch/watcher.go#L68)):*

I lost track of the level of indentation; the discussion UI in Reviewable is occupying most of my vertical screen real-estate.  Now that I'm reading what it is that you actually wrote, it makes sense as written.  It would be nice, however, to `log.WithErr().Warn()` the error from acquiring a connection or ping.  Also, the Ping method exists directly on `*sql.DB` so the explicit connection acquisition isn't strictly necessary.

There's always a risk in relying on cached data; the joke about the two most difficult software problems being naming things, cache invalidation, and off-by-one errors applies here.  I'm reasonably confident that the `getTables()` queries will do the right thing if the requested `schema` doesn't actually exist in the target database. They'll return an empty, no-error response.  If they don't it's a bug.
</blockquote></details>

Ok, cool. I added the logging and used the DB ping.

PR #983, opened 2024-08-29
checkpoint: Add option to limit the resolving range

bobvawter:
Currently, the checkpointing code always returns the latest checkpoint when computing the resolving range. This maximizes the number of mutations that can be concurrently applied, but it also means that if an apply transaction is delayed, say by encountering a locked row in the target database, the amount of temporal skew in the target data is also unbounded.

This change allows the operator to limit the number of checkpoints that will be considered when computing the upper bound of the resolving range. Lowering this value may have an adverse impact on total throughput.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/983)

### DIFF
--- a/internal/conveyor/config.go
+++ b/internal/conveyor/config.go
@@ -42,6 +42,13 @@ type Config struct {
 	// Write directly to staging tables. May limit compatibility with
 	// schemas that contain foreign keys.
 	Immediate bool
+
+	// If non-zero, limit the number of checkpoint rows that will be
+	// used to compute the resolving range. This will limit the maximum
+	// amount of observable skew in the target due to blocked mutations
+	// (e.g. running into a lock), but will cause replication to stall
+	// if behind by this many checkpoints.
+	LimitLookahead int
 }
 
 // Bind adds configuration flags to the set.
@@ -54,6 +61,9 @@ func (c *Config) Bind(f *pflag.FlagSet) {
 	f.BoolVar(&c.Immediate, "immediate", false,
 		"bypass staging tables and write directly to target; "+
 			"recommended only for KV-style workloads with no FKs")
+	f.IntVar(&c.LimitLookahead, "limitLookahead", 0,
+		"limit number of checkpoints to be considered when computing the resolving range; "+
+			"may cause replication to stall completely if older mutations cannot be applied")
 }
 
 // Preflight ensures the Config is in a known-good state.
--- a/internal/conveyor/conveyor.go
+++ b/internal/conveyor/conveyor.go
@@ -96,7 +96,11 @@ func (c *Conveyors) Get(schema ident.Schema) (*Conveyor, error) {
 		watcher: w,
 	}
 
-	ret.checkpoint, err = c.checkpoints.Start(c.stopper, tableGroup, &ret.resolvingRange)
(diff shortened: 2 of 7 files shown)


BramGruneir (dismissed):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/983)* status: all files reviewed, 2 unresolved discussions (waiting on @bobvawter, @Jeremyyang920, and @sravotto)

___
*[`internal/staging/checkpoint/group.go` line 245 at r1](https://reviewable.io/reviews/cockroachdb/replicator/983#-O5YFscY5TSP-VbaRG33:-O5YFscY5TSP-VbaRG34:b-6uj9lz) ([raw file](https://github.com/cockroachdb/replicator/blob/4da0427017b4f54927ede833e9f7a37fca2b49bc/internal/staging/checkpoint/group.go#L245)):*
> ```Go
> //     each partition within the group.
> //   - visible_data: Restricts available_data by the minimum-maximum
> //     value from p_max_times.
> ```

Add a comment about the limit please.
___
*[`internal/staging/checkpoint/group.go` line 273 at r1](https://reviewable.io/reviews/cockroachdb/replicator/983#-O5YFk0J8iHVl4l_Wtmp:-O5YFk0J8iHVl4l_Wtmq:b-cgpbxy) ([raw file](https://github.com/cockroachdb/replicator/blob/4da0427017b4f54927ede833e9f7a37fca2b49bc/internal/staging/checkpoint/group.go#L273)):*
> ```Go
>   FROM available_data
>  WHERE source_hlc <= (SELECT min(hlc) FROM partition_max_times)
>  LIMIT $3
> ```

Don't we only want a limit when when was explicitly added?

bobvawter (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/983)* status: 4 of 7 files reviewed, 2 unresolved discussions (waiting on @BramGruneir, @Jeremyyang920, and @sravotto)

___
*[`internal/staging/checkpoint/group.go` line 245 at r1](https://reviewable.io/reviews/cockroachdb/replicator/983#-O5YFscY5TSP-VbaRG33:-O5YNjqV39JzUr_qzPw4:b-896fix) ([raw file](https://github.com/cockroachdb/replicator/blob/4da0427017b4f54927ede833e9f7a37fca2b49bc/internal/staging/checkpoint/group.go#L245)):*

Add a comment about the limit please.
</blockquote></details>

Done.
___
*[`internal/staging/checkpoint/group.go` line 273 at r1](https://reviewable.io/reviews/cockroachdb/replicator/983#-O5YFk0J8iHVl4l_Wtmp:-O5YNk7vBzZEFhT0-2mh:b1w08hz) ([raw file](https://github.com/cockroachdb/replicator/blob/4da0427017b4f54927ede833e9f7a37fca2b49bc/internal/staging/checkpoint/group.go#L273)):*

Don't we only want a limit when when was explicitly added?
</blockquote></details>

The presence of the `LIMIT` clause is now conditioned on having a lookahead set.  I'm baking the limit in the SQL, since that's constant for the life of the `Group`.

BramGruneir:
Lgtm

PR #985, opened 2024-08-31
checkpoint: Rework limited checkpoint scan query

bobvawter:
The previous iteration of this query did not correctly handle operating on an existing checkpoint table. The applied limit would prevent the query from scanning ahead to actually find new work to do. This change reworks the query to progressively narrow the range of timestamps for each checkpoint partition before applying the limit. The test is also updated to ensure that the restart case is correctly handled.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/985)

### DIFF
--- a/internal/staging/checkpoint/checkpoint.go
+++ b/internal/staging/checkpoint/checkpoint.go
@@ -87,7 +87,9 @@ func (r *Checkpoints) newGroup(
 
 	var limit string
 	if lookahead > 0 {
-		limit = fmt.Sprintf("ORDER BY source_hlc LIMIT %d", lookahead)
+		// +1 to have a window that includes both the last applied
+		// checkpoint and the next unapplied checkpoint.
+		limit = fmt.Sprintf("WHERE r <= %d", lookahead+1)
 	}
 	// This query may indeed require a full table scan.
 	ret.sql.refresh = fmt.Sprintf(refreshTemplate, r.metaTable, limit)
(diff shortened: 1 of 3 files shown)


BramGruneir (changes requested):
I don't like this change at all.  It's adding a ton of complexity and is probably significantly slower when there are a lot of mutations.  If the goal is to stop applying while there is one stuck mutation, I think we could do this work in Go instead of in SQL.

> Reviewed 1 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/985#-:-O5nDgsRDIOWXuCplAqI:b-dpeba0)* status: 1 of 3 files reviewed, all discussions resolved

bobvawter:
Stopping on a locked target row is already handled by parallelism=1. This was the other part of the ask, to limit the maximum amount of temporal skew, given the transaction reordering that may have happened before the locked row is encountered.

Doesn't the optimizer elide unused projections within a result set? If there's no reference to the row numbering operation, this should devolve to a grouped min-max scan over the tail of the checkpoints table. I'll run some explains on Tuesday to compare the combination of plans.

bobvawter:
Comparing the old versus new queries, the new ones wind up iterating over the buffered rows a smaller number of times.  The overall query performance looks to be about the same with 10,000 checkpoints.

[new_limit.txt](https://github.com/user-attachments/files/16852694/new_limit.txt)
[new.txt](https://github.com/user-attachments/files/16852695/new.txt)
[old.txt](https://github.com/user-attachments/files/16852696/old.txt)

BramGruneir (approved):
Thanks for doing the analysis.  

LGTM

> Reviewed 1 of 3 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/985#-:-O5tWqbP43YWwrCuimCD:bsb52ud)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @bobvawter)

PR #1000, opened 2024-09-06
Database version backups

MattWhelan:
We use the database version string to detect the product and version of the database (and thus the SQL dialect, among other things). That happens at connection time, so we need to restore the version from the memo if the target DB is down. 

I made a few tweaks to the workload tools to test this with Oracle.


MattWhelan:
This is a pretty small change. As I'm heading out on PTO for two weeks, feel free to may any changes you see fit to make in my absence. 

There isn't an automated integration test. I had to manually run through the process of setting up an initial changefeed so that the schema/version would be cached, and then using `docker network disconnect`, pull the plug on the database, to test the disconnected case. Driving all that within a CI/CD test seems... daunting.

PR #1008, opened 2024-09-16
hlc: Introduce a monotonic Clock type

bobvawter:
As the number of logical frontends increases, the variety of external clock representations will continue to grow. This change introduces a monotonic Clock type and a way to associate an arbitrary external value with a Time.

Several frontends are updated to use a Clock instead of trying to adapt their upstream progress data into the Time type. The pglogical and mylogical frontends are prepared for an upcoming change to make them use an in-memory buffer around the core sequencer. Rather than generating MultiBatches, it is expected that frontends need only to assemble whole transactions in a TemporalBatch before handing the transaction off for processing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1008)

### DIFF
--- a/internal/cmd/workload/runner.go
+++ b/internal/cmd/workload/runner.go
@@ -36,7 +36,7 @@ type runner struct {
 	cfg          *clientConfig
 	gen          *workload.GeneratorBase
 	lastResolved hlc.Time
-	lastTime     hlc.Time // Ensures distinct values from hlcNow.
+	monotonic    hlc.Clock
 	limiter      *rate.Limiter
 	sender       *sender
 }
@@ -49,11 +49,10 @@ func newRunner(
 		return nil, err
 	}
 	return &runner{
-		cfg:      cfg,
-		gen:      gen,
-		lastTime: hlc.Time{},
-		limiter:  rate.NewLimiter(rate.Limit(cfg.rate), 1),
-		sender:   sender,
+		cfg:     cfg,
+		gen:     gen,
+		limiter: rate.NewLimiter(rate.Limit(cfg.rate), 1),
+		sender:  sender,
 	}, nil
 }
 
@@ -102,14 +101,14 @@ func (r *runner) Run(ctx *stopper.Context) error {
 		select {
 		case <-ctx.Stopping():
 			// Send a final checkpoint to allow the test to complete.
-			resolved := r.hlcNow()
+			resolved := r.monotonic.Now()
 			r.lastResolved = resolved
 			final := r.sendResolved(ctx, resolved, sendResolvedAfter)
 			return lockset.Wait(ctx, []lockset.Outcome{final})
 
 		case <-resolvedTicker.C:
 			// Enough time has elapsed that we want to send a checkpoint.
-			resolved := r.hlcNow()
+			resolved := r.monotonic.Now()
 			r.lastResolved = resolved
 			waitFor := sendResolvedAfter
 			sendResolvedAfter = nil
@@ -128,7 +127,7 @@ func (r *runner) sendBatch(ctx *stopper.Context) (int, lockset.Outcome) {
 	}
 	batch := &types.MultiBatch{}
 	for range r.cfg.batchSize {
-		r.gen.GenerateInto(batch, r.hlcNow())
+		r.gen.GenerateInto(batch, r.monotonic.Now())
 
 		// Send a smaller payload if we've hit the rate limit.
 		if !r.limiter.Allow() {
@@ -153,17 +152,4 @@ func (r *runner) sendResolved(
 		return notify.VarOf(lockset.StatusFor(err))
 	}
 	return r.sender.Send(ctx, payload)
-}
-
-// hlcNow returns a new HLC time that is always greater than the one
-// returned previously. It will increment the logical counter if the
-// wall time has not advanced.
-func (r *runner) hlcNow() hlc.Time {
-	nextNanos := time.Now().UnixNano()
-	if nextNanos > r.lastTime.Nanos() {
-		r.lastTime = hlc.New(nextNanos, 0)
-		return r.lastTime
-	}
-	r.lastTime = r.lastTime.Next()
-	return r.lastTime
 }
--- a/internal/sequencer/sequencer.go
+++ b/internal/sequencer/sequencer.go
@@ -87,9 +87,10 @@ func NewStat(group *types.TableGroup, progress *ident.TableMap[hlc.Range]) Stat 
 
 // CommonProgress returns the minimum progress across all tables within
 // the [Stat.Group]. If no progress has been made for one or more tables
-// in the group, [hlc.RangeEmpty] will be returned.
+// in the group, or if the group is empty, [hlc.RangeEmpty] will be
+// returned.
 func CommonProgress(s Stat) hlc.Range {
-	if s == nil {
+	if s == nil || len(s.Group().Tables) == 0 {
 		return hlc.RangeEmpty()
 	}
 	group := s.Group()
--- a/internal/sequencer/sequencer_test.go
+++ b/internal/sequencer/sequencer_test.go
@@ -52,4 +52,8 @@ func TestCommonProgress(t *testing.T) {
 	// All tables have progress.
 	progress.Put(group.Tables[1], hlc.RangeIncluding(hlc.Zero(), hlc.New(2, 1)))
 	r.Equal(hlc.RangeIncluding(hlc.Zero(), hlc.New(1, 1)), CommonProgress(stat))
+
+	// Empty group with no tables.
+	r.Equal(hlc.RangeEmpty(),
+		CommonProgress(NewStat(&types.TableGroup{}, &ident.TableMap[hlc.Range]{})))
 }
--- a/internal/source/kafka/integration_test.go
+++ b/internal/source/kafka/integration_test.go
@@ -309,21 +309,19 @@ func testWorkload(t *testing.T, fc *fixtureConfig) {
 		Stat() *notify.Var[sequencer.Stat]
 	}).Stat()
 
-	lastTime := hlc.New(time.Now().UnixNano(), 0)
-	r.NoError(producer.writeResolved(lastTime))
+	var clock hlc.Clock
+	r.NoError(producer.writeResolved(clock.Now()))
 	for iter := 1; iter <= maxIterations; iter++ {
 		batch := &types.MultiBatch{}
 		size := rand.IntN(maxBatchSize) + 1
 		for i := 0; i < size; i++ {
-			lastTime = hlcNow(lastTime)
-			workload.GenerateInto(batch, lastTime)
+			workload.GenerateInto(batch, clock.Now())
 		}
 		r.NoError(producer.writeBatch(batch))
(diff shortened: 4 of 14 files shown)


ZhouXing19 (commented):
This is super helpful, thanks so much! 
I just have some clarification questions. 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1008#-:-O7-U6ZX0_vocJL9asd_:b-sdui5r)* status: 0 of 14 files reviewed, 3 unresolved discussions (waiting on @bobvawter, @BramGruneir, @Jeremyyang920, and @sravotto)

___
*[`internal/source/pglogical/conn.go` line 128 at r2](https://reviewable.io/reviews/cockroachdb/replicator/1008#-O7-TCvO57obb7MTQILi:-O7-TCvO57obb7MTQILj:b1qcw2t) ([raw file](https://github.com/cockroachdb/replicator/blob/d34f12c6549b30b13643ad028c8a74d5c34bc49f/internal/source/pglogical/conn.go#L128)):*
> ```Go
> 		ignoreLSN, _ := c.walOffset.Get()
> 
> 		if msg.CommitLSN <= ignoreLSN {
> ```

Maybe I missed something, but why we no longer need this check?
___
*[`internal/util/hlc/clock.go` line 109 at r2](https://reviewable.io/reviews/cockroachdb/replicator/1008#-O7-SB7c4a56o2WCXLjH:-O7-SB7d8_-nIPYap5PE:b-i33dvo) ([raw file](https://github.com/cockroachdb/replicator/blob/51dd3044acc4a5bb9e510058e75ecc028d5bf496/internal/util/hlc/clock.go#L109)):*
> ```Go
> 	} else {
> 		// The clock and/or logical values have gone backwards.
> 		nextNanos = last.Nanos() + 1
> ```

Not opposing and just out of curiosity, why do we always +1 to the nano time here? 

The scenario I am thinking is that for oracle frontend, the only "timestamp" equivalent is SCN, which, IIUC, should be supplied as ext in the construction of a `Time` variable. In this case, I'm wondering if we still need the nano / logical concept, except to possibly gauge the receive and/or latency. 

Or to put my question in another way, I am still a bit confused about the purpose of each part of the `Time` struct. Is ext only be used to report the progress, but not to sort temporal batches, iiuc? 
___
*[`internal/util/hlc/clock_test.go` line 53 at r2](https://reviewable.io/reviews/cockroachdb/replicator/1008#-O7-QpSL1QQ1fbHr4JCJ:-O7-QpSL1QQ1fbHr4JCK:b-e5kg9) ([raw file](https://github.com/cockroachdb/replicator/blob/51dd3044acc4a5bb9e510058e75ecc028d5bf496/internal/util/hlc/clock_test.go#L53)):*
> ```Go
> 		r.Equal(logical, next.Logical())
> 
> 		// Nanos must always
> ```

nit: comment seems to be cut off

BramGruneir (changes requested):
> Reviewed 12 of 12 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1008)* status: all files reviewed (commit messages unreviewed), 5 unresolved discussions (waiting on @bobvawter, @Jeremyyang920, and @sravotto)

___
*[`internal/util/hlc/clock.go` line 63 at r1](https://reviewable.io/reviews/cockroachdb/replicator/1008#-O7-V8OC0vjMOT5sA18J:-O7-V8OC0vjMOT5sA18K:b-4lhakw) ([raw file](https://github.com/cockroachdb/replicator/blob/6f43f1c71b792fb1dd90a41d2bfc92bf756793b2/internal/util/hlc/clock.go#L63)):*
> ```Go
> // Last returns the most recently emitted Time.
> func (c *Clock) Last() Time {
> 	c.mu.Lock()
> ```

Is it worth doing a RLock here ?
___
*[`internal/util/hlc/clock.go` line 106 at r1](https://reviewable.io/reviews/cockroachdb/replicator/1008#-O7-XPo822O7Ar-cxWOG:-O7-XPo822O7Ar-cxWOH:bh4jm68) ([raw file](https://github.com/cockroachdb/replicator/blob/6f43f1c71b792fb1dd90a41d2bfc92bf756793b2/internal/util/hlc/clock.go#L106)):*
> ```Go
> 	} else if proposedNanos == last.Nanos() && logical > last.Logical() {
> 		// Within clock's resolution, but logical advances.
> 		nextLogical = logical
> ```

Won't this set nextNanos to 0 in some cases?  Should you not just have one if case?

if (proposedNanos > last.Nonos()) || (proposedNanos == last.Nanos() && logical > last.Logical()) {
	nextNanos = proposedNanos
	nextLogical = logical
} else {
  	nextNanos = last.Nanos() + 1
}

And I think you're missing a test case.

ZhouXing19:
Inspired by the walkthrough session, do we need to update the staging table to include ext? 

i.e. this table
```go
const tableSchema = `
CREATE TABLE IF NOT EXISTS %[1]s (
       nanos INT NOT NULL,
     logical INT NOT NULL,
         key STRING NOT NULL,
         mut BYTES NOT NULL,
      before BYTES NULL,
     applied BOOL NOT NULL DEFAULT false,
  applied_at TIMESTAMPTZ NULL,
    deletion BOOL NOT NULL DEFAULT false,
  %[2]s
  PRIMARY KEY (nanos, logical, key),
    INDEX %[3]s (key) STORING (applied), -- Improve performance of StageIfExists
   FAMILY cold (mut, before, deletion),
   FAMILY hot (applied, applied_at)
)`
```

bobvawter (commented):
No, I don't see a need for single-stream frontends to write to the staging tables, since there's no need to assemble whole transactions from disparate receivers (unlike CDC changefeeds, where no incoming message can be assumed to contain an entire transaction).

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1008#-:-O713CZ40q7eEFs_8Ae8:b-s5g56o)* status: 12 of 14 files reviewed, 5 unresolved discussions (waiting on @BramGruneir, @Jeremyyang920, @sravotto, and @ZhouXing19)

___
*[`internal/source/pglogical/conn.go` line 128 at r2](https://reviewable.io/reviews/cockroachdb/replicator/1008#-O7-TCvO57obb7MTQILi:-O7-Waol5r6KT0bxNnQ-:b3njmr4) ([raw file](https://github.com/cockroachdb/replicator/blob/d34f12c6549b30b13643ad028c8a74d5c34bc49f/internal/source/pglogical/conn.go#L128)):*
<details><summary><i>Previously, ZhouXing19 (Jane Xing) wrote…</i></summary><blockquote>

Maybe I missed something, but why we no longer need this check?
</blockquote></details>

This check was written before the sequencer stack A) existed, and B) learned to deal with non-idempotent redelivery of data. It's valid for a CockroachDB changefeed to emit, for a single rows, a sequence of updates `(1, 2, 3, 4, 2)`. The stack can suppress out-of-sequence updates, even if the stream is restarted.
___
*[`internal/util/hlc/clock.go` line 63 at r1](https://reviewable.io/reviews/cockroachdb/replicator/1008#-O7-V8OC0vjMOT5sA18J:-O70KUO42EJjvvIN6ozE:b-ulssde) ([raw file](https://github.com/cockroachdb/replicator/blob/6f43f1c71b792fb1dd90a41d2bfc92bf756793b2/internal/util/hlc/clock.go#L63)):*

Is it worth doing a RLock here ?
</blockquote></details>

I thought about an RLock, but this is the only reader and only really for testing.
___
*[`internal/util/hlc/clock.go` line 106 at r1](https://reviewable.io/reviews/cockroachdb/replicator/1008#-O7-XPo822O7Ar-cxWOG:-O70Kh7J7O6dQrnaxz7f:b-k2rwtz) ([raw file](https://github.com/cockroachdb/replicator/blob/6f43f1c71b792fb1dd90a41d2bfc92bf756793b2/internal/util/hlc/clock.go#L106)):*

Won't this set nextNanos to 0 in some cases?  Should you not just have one if case?

if (proposedNanos > last.Nonos()) || (proposedNanos == last.Nanos() && logical > last.Logical()) {
	nextNanos = proposedNanos
	nextLogical = logical
} else {
  	nextNanos = last.Nanos() + 1
}

And I think you're missing a test case.
</blockquote></details>

Good catch. Fixed.
___
*[`internal/util/hlc/clock.go` line 109 at r2](https://reviewable.io/reviews/cockroachdb/replicator/1008#-O7-SB7c4a56o2WCXLjH:-O7-Y1MZ-UO4LafxA4M4:b-g4l2p0) ([raw file](https://github.com/cockroachdb/replicator/blob/51dd3044acc4a5bb9e510058e75ecc028d5bf496/internal/util/hlc/clock.go#L109)):*
<details><summary><i>Previously, ZhouXing19 (Jane Xing) wrote…</i></summary><blockquote>

Not opposing and just out of curiosity, why do we always +1 to the nano time here? 

The scenario I am thinking is that for oracle frontend, the only "timestamp" equivalent is SCN, which, IIUC, should be supplied as ext in the construction of a `Time` variable. In this case, I'm wondering if we still need the nano / logical concept, except to possibly gauge the receive and/or latency. 

Or to put my question in another way, I am still a bit confused about the purpose of each part of the `Time` struct. Is ext only be used to report the progress, but not to sort temporal batches, iiuc?
</blockquote></details>

The `SCN` would be passed to `Clock.External()`, the same way the `GTIDSet` is for MySQL.  For these single-stream frontends, the Time acts as a marker to summarize the progress of (concurrently) applied transactions.  It's convenient for it to contain an approximate wall time, since that can drive lag metrics.
___
*[`internal/util/hlc/clock_test.go` line 53 at r2](https://reviewable.io/reviews/cockroachdb/replicator/1008#-O7-QpSL1QQ1fbHr4JCJ:-O70Ma9C2iXTh1X0kr3y:bt2vr1b) ([raw file](https://github.com/cockroachdb/replicator/blob/51dd3044acc4a5bb9e510058e75ecc028d5bf496/internal/util/hlc/clock_test.go#L53)):*
<details><summary><i>Previously, ZhouXing19 (Jane Xing) wrote…</i></summary><blockquote>

nit: comment seems to be cut off
</blockquote></details>

Fixed.

BramGruneir (approved):
:lgtm:

> Reviewed 1 of 2 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1008#-:-O7538wY99vRyyPFmR8B:bnfp4nl)* status: all files reviewed, 4 unresolved discussions (waiting on @bobvawter, @Jeremyyang920, @sravotto, and @ZhouXing19)

___
*[`internal/util/hlc/clock.go` line 101 at r4](https://reviewable.io/reviews/cockroachdb/replicator/1008#-O752DpT2FC_ynQiIZEz:-O752DpT2FC_ynQiIZF-:b-2qnetw) ([raw file](https://github.com/cockroachdb/replicator/blob/783b3476940c5fec0f4ed5bc282acdd360104e53/internal/util/hlc/clock.go#L101)):*
> ```Go
> 	var nextLogical int
> 	if proposedNanos > last.Nanos() ||
> 		proposedNanos == last.Nanos() && logical > last.Logical() {
> ```

oh.,.. please add parentheses
Just for clarity

I get the order of operations... but still, just makes it easier to read.

PR #1012, opened 2024-09-16
sinktest: Miscellaneous test rig improvements

bobvawter:
This is a preperatory commit to use the workload checker to write into source databases when building end-to-end integration tests.

It is now possible create a "swapped" base fixture, with the source and target databases reversed. This base fixture can be further modified before constructing an all.Fixture from it.

The workload generator uses 64-bit integer keys as a hedge against any future key truncation.

Creating the test-table schema is broken up, allowing the table names and the schema string expansion to happen in separate steps.

The generated payloads now include bodies for delete statements (used by upcoming frontends) and a bug in the NewWebhookPayload() is fixed to elide the data block for deletions.


ZhouXing19 (approved):
Thanks for making this! LGTM module the webhook related part, as I am not sure about the context for it. 
Would like to have others to finalize the review as well.

PR #1019, opened 2024-09-20
sequencer: Extract staging behavior from core

bobvawter:
This change allows the source of staging data provided to the Core sequencer to
be configured. The StagingCursor type is renamed to BatchCursor and a new
BatchReader interface defines a way to receive an asynchronous stream of cursors.

All existing code is updated to inject a Staging sequencer into the stack. This
sequencer provides an acceptor which writes incoming data to the staging tables
and injects the existing staging-reader implementation into Core.

The Core sequencer's acceptor is now a no-op. The workload checker is updated
to support providing data via BatchReader. This allows Core to be tested
without use of any staging tables.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1019)

### DIFF
--- a/internal/sequencer/besteffort/besteffort_test.go
+++ b/internal/sequencer/besteffort/besteffort_test.go
@@ -47,7 +47,9 @@ func TestBestEffort(t *testing.T) {
 					// We only want BestEffort to do work when told; the test
 					// rig uses a counter to generate timestamps.
 					seqFixture.BestEffort.SetTimeSource(hlc.Zero)
-					next, err := seqFixture.BestEffort.Wrap(fixture.Context, seqFixture.Core)
+					next, err := seqFixture.Staging.Wrap(fixture.Context, seqFixture.Core)
+					require.NoError(t, err)
+					next, err = seqFixture.BestEffort.Wrap(fixture.Context, next)
 					require.NoError(t, err)
 					return next
 				},
--- a/internal/sequencer/core/acceptor.go
+++ b/internal/sequencer/core/acceptor.go
@@ -20,66 +20,32 @@ import (
 	"context"
 
 	"github.com/cockroachdb/replicator/internal/types"
-	"github.com/cockroachdb/replicator/internal/util/hlc"
-	"github.com/cockroachdb/replicator/internal/util/ident"
-	"github.com/cockroachdb/replicator/internal/util/retry"
+	"github.com/pkg/errors"
 )
 
-// An acceptor writes incoming data to staging.
-type acceptor struct {
-	*Core
-}
+// acceptor rejects all incoming data since core only operates from a
+// BatchReader source.
+type acceptor struct{}
 
 var _ types.MultiAcceptor = (*acceptor)(nil)
 
-// AcceptMultiBatch implements [types.MultiAcceptor] and processes the
-// batch in time order.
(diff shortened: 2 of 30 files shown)


BramGruneir (approved):
:lgtm:

> Reviewed 31 of 31 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1019#-:-O7VtkTb5ocuTuJusOsW:bnfp4nl)* status: all files reviewed, 4 unresolved discussions (waiting on @bobvawter, @Jeremyyang920, @sravotto, and @ZhouXing19)

___
*[`internal/sequencer/core/core.go` line 41 at r1](https://reviewable.io/reviews/cockroachdb/replicator/1019#-O7G6EiR1haa0D6JTq5W:-O7G6EiR1haa0D6JTq5X:b-y861ob) ([raw file](https://github.com/cockroachdb/replicator/blob/925d6f341dd05f354129b022b0ac17a6dbc89691/internal/sequencer/core/core.go#L41)):*
> ```Go
> 
> // The Core sequencer applies transactional data read from a
> // [types.BatchReader] in a possibly-concurrent manner.
> ```

Can you expland on this comment about possible-concurrency?  
___
*[`internal/sequencer/core/core.go` line 114 at r1](https://reviewable.io/reviews/cockroachdb/replicator/1019#-O7G6OFm3ZOa8TiMiuh1:-O7G6OFm3ZOa8TiMiuh2:bgy7tjc) ([raw file](https://github.com/cockroachdb/replicator/blob/925d6f341dd05f354129b022b0ac17a6dbc89691/internal/sequencer/core/core.go#L114)):*
> ```Go
> 						_, _, _ = progress.Update(func(stat sequencer.Stat) (sequencer.Stat, error) {
> 							stat = stat.Copy()
> 							if len(group.Tables) == 0 {
> ```

Is there a test for this?
___
*[`internal/sequencer/staging/acceptor.go` line 58 at r1](https://reviewable.io/reviews/cockroachdb/replicator/1019#-O7GE6FB7Z8JMvzOfMvK:-O7GE6FB7Z8JMvzOfMvL:b-8p2eb8) ([raw file](https://github.com/cockroachdb/replicator/blob/925d6f341dd05f354129b022b0ac17a6dbc89691/internal/sequencer/staging/acceptor.go#L58)):*
> ```Go
> 	ctx context.Context, batch *types.TableBatch, opts *types.AcceptOptions,
> ) error {
> 	stager, err := a.stagers.Get(ctx, batch.Table)
> ```

Isn't this the same as the inner loop of Accept MultiBatch (with a change in varialbe for stage?
Why not have them both call the same function?  
___
*[`internal/sequencer/staging/staging.go` line 68 at r1](https://reviewable.io/reviews/cockroachdb/replicator/1019#-O7Vsi8m-BCWixpGR-Il:-O7Vsi8m-BCWixpGR-Im:b4v9tri) ([raw file](https://github.com/cockroachdb/replicator/blob/925d6f341dd05f354129b022b0ac17a6dbc89691/internal/sequencer/staging/staging.go#L68)):*
> ```Go
> 	opts.Delegate = s.markers.MultiAcceptor(opts.Delegate)
> 
> 	if err != nil {
> ```

Why is this after setting the delegate?

PR #1027, opened 2024-09-25
replicator: Use iterators instead of Range callbacks.

bobvawter:
This change replaces the Range callback functions with the new iterator type compatible with the range keyword in go 1.23.

The various map-like types receive iterator methods All(), Keys(), and Values(). The names of these methods are in line with the new iterator methods in the stdlib maps package. The Entries() and Range() methods are removed.

The apply templates were the only actual use of Entries(), which has been hoisted into a template helper function to support deadlined columns.

The net effect is to make flow-control more idiomatic. There were some cases where one would write "return nil" within a Range() callback when "continue" would be more natural or the use of "return sentinelError" when "break" would be the obvious choice. Many cases of always-nil error handling have been removed as well.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1027)

### DIFF
--- a/internal/script/bag_wrapper.go
+++ b/internal/script/bag_wrapper.go
@@ -60,11 +60,9 @@ func (m *bagWrapper) Has(key string) bool {
 // Keys implements goja.DynamicObject.
 func (m *bagWrapper) Keys() []string {
 	ret := make([]string, 0, m.data.Len())
-	// Ignoring error since callback returns nil.
-	_ = m.data.Range(func(k ident.Ident, v any) error {
+	for k := range m.data.Keys() {
 		ret = append(ret, k.Raw())
-		return nil
-	})
+	}
 	return ret
 }
 
--- a/internal/types/batches.go
+++ b/internal/types/batches.go
@@ -23,33 +23,31 @@ package types
 
 import (
 	"bytes"
+	"iter"
+	"slices"
 	"sort"
-	"strings"
 
 	"github.com/cockroachdb/replicator/internal/util/hlc"
 	"github.com/cockroachdb/replicator/internal/util/ident"
 	"github.com/pkg/errors"
 )
 
-// Accumulator contains the Accumulate method.
-type Accumulator interface {
-	// Accumulate the mutation and ensure that batch invariants are
-	// maintained.
-	Accumulate(table ident.Table, mut Mutation) error
-}
-
-// AccumulatorFunc adapts a function to the [Accumulator] interface.
-type AccumulatorFunc func(ident.Table, Mutation) error
-
-// Accumulate implements [Accumulator].
-func (f AccumulatorFunc) Accumulate(table ident.Table, mut Mutation) error {
-	return f(table, mut)
+// Apply the function to the given iterator.
+func Apply[K any, V any](it iter.Seq2[K, V], fn func(K, V) error) error {
+	for k, v := range it {
+		if err := fn(k, v); err != nil {
+			return err
+		}
+	}
+	return nil
 }
 
 // The Batch interface is implemented by the various Batch types in this
 // package.
 type Batch[B any] interface {
-	Accumulator
+	// Accumulate the mutation and ensure that batch invariants are
+	// maintained.
+	Accumulate(table ident.Table, mut Mutation) error
 
 	// Count returns the number of mutations contained in the Batch.
 	Count() int
@@ -57,34 +55,31 @@ type Batch[B any] interface {
 	// Copy returns a deep copy of the Batch.
 	Copy() B
 
-	// CopyInto copies the contents of the batch into accumulator.
-	CopyInto(acc Accumulator) error
-
 	// Empty returns a copy of the Batch, but with no enclosed
 	// mutations. This is useful when wanting to transform or filter a
 	// batch.
 	Empty() B
+
+	// Mutations returns an iterator over all mutations in the Batch. No
+	// specific iteration order is defined.
+	Mutations() iter.Seq2[ident.Table, Mutation]
 }
 
 // Flatten copies all mutations in a batch into a slice.
 func Flatten[B Batch[B]](batch B) []Mutation {
 	var ret []Mutation
-	// Ignoring error since callback returns nil.
-	_ = batch.CopyInto(AccumulatorFunc(func(_ ident.Table, mut Mutation) error {
+	for _, mut := range batch.Mutations() {
 		ret = append(ret, mut)
-		return nil
-	}))
+	}
 	return ret
 }
 
 // FlattenByTable copies the mutations into an [ident.TableMap].
 func FlattenByTable[B Batch[B]](batch B) *ident.TableMap[[]Mutation] {
 	ret := &ident.TableMap[[]Mutation]{}
-	// Ignoring error since callback returns nil.
-	_ = batch.CopyInto(AccumulatorFunc(func(table ident.Table, mut Mutation) error {
+	for table, mut := range batch.Mutations() {
 		ret.Put(table, append(ret.GetZero(table), mut))
-		return nil
-	}))
+	}
 	return ret
 }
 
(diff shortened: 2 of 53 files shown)


internal/script/bag_wrapper.go
```diff
@@ -60,11 +60,9 @@ func (m *bagWrapper) Has(key string) bool {
 // Keys implements goja.DynamicObject.
 func (m *bagWrapper) Keys() []string {
 	ret := make([]string, 0, m.data.Len())
-	// Ignoring error since callback returns nil.
-	_ = m.data.Range(func(k ident.Ident, v any) error {
+	for k := range m.data.Keys() {
```
ryanluu12345, line 63: Just for context, is there any reason the old range function callbacks returned error but we don't have to do any error handling now that we have switched to iterators? 

In the iterator code itself is there anything that could result in an error that we are handling within the iterator function?

internal/types/batches.go
```diff
@@ -23,68 +23,63 @@ package types
...
-}
-
-// AccumulatorFunc adapts a function to the [Accumulator] interface.
-type AccumulatorFunc func(ident.Table, Mutation) error
-
-// Accumulate implements [Accumulator].
-func (f AccumulatorFunc) Accumulate(table ident.Table, mut Mutation) error {
-	return f(table, mut)
+// Apply the function to the given iterator.
+func Apply[K any, V any](it iter.Seq2[K, V], fn func(K, V) error) error {
```
ryanluu12345, line 36: Is there ever a scenario where we want to apply the function to the iterator value but also get the result of the application? A scenario I can think about is if we want to transform the data, apply it to the target, and get the applied data for like debugging or logging purposes (really contrived, but wondering if it's a common use case)

ryanluu12345, line 36: Is it worthwhile to create an alias type for `iter.Seq2[K,V]` and put the apply as a method of the iterator type? Not sure if that's supported for iterators, but may simplify since users can call an iterator's apply function and just pass in their applied function.

internal/types/batches.go
```diff
@@ -187,6 +171,20 @@ func (b *MultiBatch) Less(i, j int) bool {
 	return hlc.Compare(b.Data[i].Time, b.Data[j].Time) < 0
 }
 
+// Mutations returns an iterator over all mutations in the Batch.
+// Mutations will be presented in time, table, and key order.
+func (b *MultiBatch) Mutations() iter.Seq2[ident.Table, Mutation] {
```
ryanluu12345, line 176: Nice! The meat of the iteration logic

internal/types/batches.go
```diff
@@ -252,6 +236,22 @@ func (b *TableBatch) Empty() *TableBatch {
 	}
 }
 
+// Mutations returns an iterator over all mutations in the Batch.
```
ryanluu12345, line 239: How come for the table batch and temporal batch modes, we sort by the keys, but in the multi batch case above we don't sort before yielding?

internal/util/merge/bag.go
```diff
@@ -102,15 +103,32 @@ func NewBagFrom(b *Bag) *Bag {
 	return NewBag(b.BagSpec)
 }
 
+// All returns an iterator over all mapped and unmapped properties in
+// the Bag.
+func (b *Bag) All() iter.Seq2[ident.Ident, any] {
```
ryanluu12345, line 108: Just to clarify, is the intended logic here that all the mapped datas get returned to the caller first, then all the unmapped data?

bobvawter (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1027)* status: 0 of 53 files reviewed, 6 unresolved discussions (waiting on @BramGruneir, @Jeremyyang920, @ryanluu12345, and @sravotto)

___
*[`internal/script/bag_wrapper.go` line 63 at r3](https://reviewable.io/reviews/cockroachdb/replicator/1027#gh-1781730053:-O83oRsC2GMrIuumHEvm:b-xbc4da) ([raw file](https://github.com/cockroachdb/replicator/blob/2b2ab6c28eab6e4cc1a4ef00664efb9cd9cffc7c/internal/script/bag_wrapper.go#L63)):*

Just for context, is there any reason the old range function callbacks returned error but we don't have to do any error handling now that we have switched to iterators? 

In the iterator code itself is there anything that could result in an error that we are handling within the iterator function?
</blockquote></details>

In a few cases the Range callback did return an error that would need to exit the outermost function.  In most cases, though, the callback always returned `nil`. 
___
*[`internal/types/batches.go` line 36 at r3](https://reviewable.io/reviews/cockroachdb/replicator/1027#gh-1781736928:-O83pWZ91KR94yz3sEYO:b478j0e) ([raw file](https://github.com/cockroachdb/replicator/blob/2b2ab6c28eab6e4cc1a4ef00664efb9cd9cffc7c/internal/types/batches.go#L36)):*

Is there ever a scenario where we want to apply the function to the iterator value but also get the result of the application? A scenario I can think about is if we want to transform the data, apply it to the target, and get the applied data for like debugging or logging purposes (really contrived, but wondering if it's a common use case)
</blockquote></details>

That sounds like either `Reduce` or `Fold`.
___
*[`internal/types/batches.go` line 36 at r3](https://reviewable.io/reviews/cockroachdb/replicator/1027#gh-1781738220:-O83pvmuDyfK7wD_yNfd:bk183dx) ([raw file](https://github.com/cockroachdb/replicator/blob/2b2ab6c28eab6e4cc1a4ef00664efb9cd9cffc7c/internal/types/batches.go#L36)):*

Is it worthwhile to create an alias type for `iter.Seq2[K,V]` and put the apply as a method of the iterator type? Not sure if that's supported for iterators, but may simplify since users can call an iterator's apply function and just pass in their applied function.
</blockquote></details>

With a type alias the callsite would look something like:

```
type Applier[K any, V any] iter.Seq2[K, V]
func (a Applier[K,V]) Apply(fn func() ... ) {}

// To use:
types.Applier(collection.All()).Apply(func () { .... })
```

I think the need for explicit type conversion makes it a wash.  If go had Kotlin-style extension methods, it would be a good choice.
___
*[`internal/types/batches.go` line 176 at r3](https://reviewable.io/reviews/cockroachdb/replicator/1027#gh-1781740449:-O83qioKEjEeNHasEert:b-896fix) ([raw file](https://github.com/cockroachdb/replicator/blob/2b2ab6c28eab6e4cc1a4ef00664efb9cd9cffc7c/internal/types/batches.go#L176)):*

Nice! The meat of the iteration logic
</blockquote></details>

Done.
___
*[`internal/types/batches.go` line 239 at r3](https://reviewable.io/reviews/cockroachdb/replicator/1027#gh-1781741132:-O83ql6C1jsR6kd_B6qW:b-d60x4j) ([raw file](https://github.com/cockroachdb/replicator/blob/2b2ab6c28eab6e4cc1a4ef00664efb9cd9cffc7c/internal/types/batches.go#L239)):*

How come for the table batch and temporal batch modes, we sort by the keys, but in the multi batch case above we don't sort before yielding?
</blockquote></details>

MultiBatches are internally sorted by time. That invariant is maintained by the `Accumulate` method.
___
*[`internal/util/merge/bag.go` line 108 at r3](https://reviewable.io/reviews/cockroachdb/replicator/1027#gh-1781745702:-O83qvF7AjTx1xo8pjQv:bb77ufg) ([raw file](https://github.com/cockroachdb/replicator/blob/2b2ab6c28eab6e4cc1a4ef00664efb9cd9cffc7c/internal/util/merge/bag.go#L108)):*

Just to clarify, is the intended logic here that all the mapped datas get returned to the caller first, then all the unmapped data?
</blockquote></details>

Yes.

BramGruneir (approved):
This makes the code much more readable without the inner functions everywhere!
:lgtm:

> Reviewed 2 of 2 files at r1, 45 of 45 files at r2, 13 of 13 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1027#-:-O8DVxsXER5ao_KSIm53:b-26n9dp)* status: all files reviewed, 6 unresolved discussions (waiting on @Jeremyyang920, @ryanluu12345, and @sravotto)

Issue #689, opened 2024-02-12
Add `sinktest/base.GetRowCountWithPredicate()` helper


bobvawter:
We have several cases where test code wants a variation on the existing `GetRowCount()`, but needs to add a substitution parameter or two. We should have some way of moving per-target complexity out of the individual tests.

ryanluu12345:
What are example tests that can benefit from this enhancement? Where exactly should I look for this?

bobvawter:
`git grep -i 'select count'`

https://github.com/cockroachdb/replicator/blob/master/internal/sequencer/script/script_test.go#L200-L211

PR #1032, opened 2024-09-30
sequencer: updated the script test so that it uses the new GetRowWithPredicate method

ryanluu12345:
Previously, product specific switch cases were used to determine how to build a query for the target database in the script test. We want to commonize this to a helper that can take care of the product specific logic and execute the query to find the row counts based on predicate.

Integrated this into the script test after implementing.

Resolves: #689
Release Note: None

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1032)

### DIFF
--- a/internal/sinktest/base/table_info.go
+++ b/internal/sinktest/base/table_info.go
@@ -81,3 +81,42 @@ func GetRowCount[P types.AnyPool](ctx context.Context, db P, name ident.Table) (
 	})
 	return count, err
 }
+
+// GetRowCountWithPredicate returns the number of rows in the table that match
+// the predicate. This method supports N predicates for filtering, passed
+// through args. The first argument is the predicate string (either a raw SQL
+// string or a parameterized one), and the remaining are the arguments for the
+// parameterized query. The predicate can be defined in the following ways: 1.
+// By using a formatted string for the predicate: fmt.Sprintf("foo = %d",
+// someNumber) 2. By using the SQL placeholder syntax: "foo = $1", value
+func GetRowCountWithPredicate[P types.AnyPool](
+	ctx context.Context, db P, name ident.Table, predicate string, args ...any,
+) (int, error) {
+	// Construct the query by interpolating the table name and predicate
+	query := fmt.Sprintf("SELECT count(*) FROM %s WHERE %s", name, predicate)
+
+	// Handle the pool specifics and execute the query
+	var count int
+	err := retry.Retry(ctx, db, func(ctx context.Context) error {
+		switch t := any(db).(type) {
+		case *types.SourcePool:
+			if len(args) > 0 {
+				return t.QueryRowContext(ctx, query, args...).Scan(&count)
+			}
+			return t.QueryRowContext(ctx, query).Scan(&count)
+		case *types.StagingPool:
+			if len(args) > 0 {
+				return t.QueryRow(ctx, query, args...).Scan(&count)
+			}
+			return t.QueryRow(ctx, query).Scan(&count)
+		case *types.TargetPool:
+			if len(args) > 0 {
+				return t.QueryRowContext(ctx, query, args...).Scan(&count)
+			}
+			return t.QueryRowContext(ctx, query).Scan(&count)
+		default:
+			return errors.Errorf("unimplemented %T", t)
+		}
+	})
+	return count, err
+}
--- a/internal/target/apply/apply_test.go
+++ b/internal/target/apply/apply_test.go
@@ -88,26 +88,9 @@ func TestApply(t *testing.T) {
 		return
 	}
 
-	// A helper to count the number of rows where has_default = lookFor.
-	// https://github.com/cockroachdb/replicator/issues/689
 	countHasDefault := func(lookFor string) (ct int, err error) {
-		var q string
-		switch fixture.TargetPool.Product {
-		case types.ProductCockroachDB, types.ProductPostgreSQL:
-			q = "SELECT count(*) FROM %s WHERE has_default=$1"
-		case types.ProductMariaDB, types.ProductMySQL:
-			q = "SELECT count(*) FROM %s WHERE has_default=?"
-		case types.ProductOracle:
-			q = "SELECT count(*) FROM %s WHERE has_default=:p1"
-		default:
-			panic("unimplemented")
-		}
-
-		err = fixture.TargetPool.QueryRowContext(ctx,
-			fmt.Sprintf(q, tbl.Name()),
-			lookFor,
-		).Scan(&ct)
-		return
+		count, err := base.GetRowCountWithPredicate(ctx, fixture.TargetPool, tbl.Name(), fmt.Sprintf("has_default='%s'", lookFor))
+		return count, err
 	}
 
 	// Use this jumbled name when accessing the API.
--- a/internal/script/script_test.go
+++ b/internal/script/script_test.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/cockroachdb/replicator/internal/sinktest"
 	"github.com/cockroachdb/replicator/internal/sinktest/all"
+	"github.com/cockroachdb/replicator/internal/sinktest/base"
 	"github.com/cockroachdb/replicator/internal/target/apply"
 	"github.com/cockroachdb/replicator/internal/types"
 	"github.com/cockroachdb/replicator/internal/util/applycfg"
@@ -403,10 +404,8 @@ CREATE TABLE %s.skewed_merge_times(
 		// Fix case for e.g. Oracle.
 		table, _ = fixture.Watcher.Get().OriginalName(table)
 		// Verify execution.
-		var count int
-		r.NoError(fixture.TargetPool.QueryRow(
-			fmt.Sprintf("SELECT count(*) FROM %s WHERE is_deleted = 1", table),
-		).Scan(&count))
+		count, err := base.GetRowCountWithPredicate(ctx, fixture.TargetPool, table, "is_deleted = 1")
+		r.NoError(err)
 		r.Equal(2, count)
 	})
 
--- a/internal/sequencer/script/script_test.go
+++ b/internal/sequencer/script/script_test.go
@@ -32,6 +32,7 @@ import (
 	"github.com/cockroachdb/replicator/internal/sequencer/switcher"
 	"github.com/cockroachdb/replicator/internal/sinktest"
 	"github.com/cockroachdb/replicator/internal/sinktest/all"
+	"github.com/cockroachdb/replicator/internal/sinktest/base"
 	"github.com/cockroachdb/replicator/internal/types"
 	"github.com/cockroachdb/replicator/internal/util/hlc"
 	"github.com/cockroachdb/replicator/internal/util/ident"
@@ -134,11 +135,11 @@ api.configureTable("t_2", {
 	// Fake timestamps in use.
 	seqFixture.BestEffort.SetTimeSource(hlc.Zero)
 
-	base, err := seqFixture.SequencerFor(ctx, baseMode)
+	baseSeq, err := seqFixture.SequencerFor(ctx, baseMode)
 	r.NoError(err)
 
 	bounds := &notify.Var[hlc.Range]{}
-	wrapped, err := seqFixture.Script.Wrap(ctx, base)
+	wrapped, err := seqFixture.Script.Wrap(ctx, baseSeq)
 	r.NoError(err)
 	acc, stats, err := wrapped.Start(ctx, &sequencer.StartOptions{
 		Bounds:   bounds,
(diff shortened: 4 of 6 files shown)


internal/sinktest/base/table_info.go
```diff
@@ -81,3 +81,41 @@ func GetRowCount[P types.AnyPool](ctx context.Context, db P, name ident.Table) (
 	})
 	return count, err
 }
+
+func GetRowCountWithPredicate[P types.AnyPool](ctx context.Context, db P, name ident.Table, product types.Product, predicate string) (int, error) {
+	// Handles the query strings for various product types.
+	var q string
+	switch product {
+	case types.ProductCockroachDB, types.ProductPostgreSQL:
+		// TODO: open question: should the "v" be another parameter that
```
ryanluu12345, line 90: @bobvawter note that this is a rough draft of something that works, but I think it depends on how much flexibility we want to give to the user here. I think at a minimum it seems like we want to make sure the column we filter by in the `WHERE` clause is configurable.

ryanluu12345: The most flexible way I can think about doing this is have people specify their own full predicates. This also makes it so they can define multiple conditions more easily. Consider a use case where the caller does something like this:
```
predicate := `age > 10 AND name = 'trevor'`
GetRowCountWithPredicate(ctx, db, name, product, predicate)
```

The resulting query that will be run based on these assumptions is:
```
SELECT COUNT(*) FROM name WHERE age > 10 AND name = 'trevor';
```

The downside here is that we aren't using the param substitutions we get for free from QueryRow or QueryRowContext, but it is quite flexible.

ryanluu12345: Did a rough prototype, and for the `script_test` example, it looks like this:
```
for idx, tgt := range tgts {
		var search string
		switch idx {
		case 0:
			search = "cowbell"
		case 1:
			search = "llebwoc"
		}

		//count, err := table_info.GetRowCountWithPredicate(ctx, pool, tgt, fixture.TargetPool.Product, search)
		predicate := fmt.Sprintf("v = '%s'", search)
		count, err := table_info.GetRowCountWithPredicateFlex(ctx, pool, tgt, fixture.TargetPool.Product, predicate)
		r.NoError(err)
		r.Equalf(numEmits, count, "in table %s", tgt)
	}
```

ryanluu12345:
Left some comments here and also notes about my implementation thought process. I think once we land on something we're mutually happy with, I'll go ahead and make these updates to the other test locations:

I searched the repo for this: `SELECT count(*) FROM %s`

Found places that need some fixing:
- `integration_test.go` for pglogical and mylogical
- `internal/target/apply/apply_test.go`
- `dlq_test.go`

internal/target/apply/apply_test.go
```diff
@@ -91,23 +92,9 @@ func TestApply(t *testing.T) {
...
-		default:
-			panic("unimplemented")
-		}
-
-		err = fixture.TargetPool.QueryRowContext(ctx,
-			fmt.Sprintf(q, tbl.Name()),
-			lookFor,
-		).Scan(&ct)
-		return
+		predicate := fmt.Sprintf(`has_default='%s'`, lookFor)
```
ryanluu12345, line 95: Here is another example of how to use this. Honestly, can simplify this to one line and just do `GetRowCountWithPredicateFlex(...,fmt.Sprintf(`has_default='%s'`, lookFor))`

bobvawter (commented):
```
func GetRowCountWithPredicate[P types.AnyPool](ctx context.Context, db P, name ident.Table, where string, args... any) (int, error) {}


// The caller can do whatever is situationally appropriate.
GetRowCountWithPredicate(ctx, pool, table, fmt.Sprintf("foo = %d", someNumber))
GetRowCountWithPredicate(ctx, pool, table,"foo = $1", someNumber)
```

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1032#-:-O8CpzKD9p3CqQn8g3z1:bms3g68)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @BramGruneir and @Jeremyyang920)

ryanluu12345 (commented):
We support both these use cases now. I intentionally refactored one of these tests to do the SQL substitution. And for the rest, I did the fmt.Sprintf way.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1032#-:-O8Dpf7J72V7ZwyjcRxj:b-o6ry2j)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @BramGruneir and @Jeremyyang920)

bobvawter (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1032)* status: 0 of 6 files reviewed, 5 unresolved discussions (waiting on @BramGruneir, @Jeremyyang920, and @ryanluu12345)

___
*[`internal/script/script_test.go` line 28 at r4](https://reviewable.io/reviews/cockroachdb/replicator/1032#-O8HgKSYD3ZzEeFvGHZP:-O8HgKSYD3ZzEeFvGHZQ:bzc8ig4) ([raw file](https://github.com/cockroachdb/replicator/blob/3a8c0b96290531e75ece5ee72e9c18fa05dfe72c/internal/script/script_test.go#L28)):*
> ```Go
> 	"github.com/cockroachdb/replicator/internal/sinktest"
> 	"github.com/cockroachdb/replicator/internal/sinktest/all"
> 	table_info "github.com/cockroachdb/replicator/internal/sinktest/base"
> ```

Unnecessary alias, here and elsewhere.
___
*[`internal/sinktest/base/table_info.go` line 100 at r4](https://reviewable.io/reviews/cockroachdb/replicator/1032#-O8Hgel52ovBmwckerq2:-O8Hgel52ovBmwckerq3:b-dxqgrd) ([raw file](https://github.com/cockroachdb/replicator/blob/3a8c0b96290531e75ece5ee72e9c18fa05dfe72c/internal/sinktest/base/table_info.go#L100)):*
> ```Go
> 
> 	// The first argument is the predicate
> 	predicate, ok := args[0].(string)
> ```

There's no advantage to passing the predicate in the varargs.
___
*[`internal/sinktest/base/table_info.go` line 116 at r4](https://reviewable.io/reviews/cockroachdb/replicator/1032#-O8HgnMFAIr3XvmGIWoY:-O8HgnMFAIr3XvmGIWoZ:b-odplmn) ([raw file](https://github.com/cockroachdb/replicator/blob/3a8c0b96290531e75ece5ee72e9c18fa05dfe72c/internal/sinktest/base/table_info.go#L116)):*
> ```Go
> 		q = "SELECT count(*) FROM %s WHERE %s"
> 	case types.ProductOracle:
> 		q = "SELECT count(*) FROM %s WHERE %s"
> ```

These queries are identical.
___
*[`internal/target/apply/apply_test.go` line 95 at r3](https://reviewable.io/reviews/cockroachdb/replicator/1032#gh-1781865464:-O8Hh6si5x8ma_wPLQiV:b-jc43kd) ([raw file](https://github.com/cockroachdb/replicator/blob/9dcfc70710579303b68914b5e6acc5d8c32f8a08/internal/target/apply/apply_test.go#L95)):*

Here is another example of how to use this. Honestly, can simplify this to one line and just do `GetRowCountWithPredicateFlex(...,fmt.Sprintf(`has_default='%s'`, lookFor))`
</blockquote></details>

Good. That's the goal.

ryanluu12345 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1032)* status: 0 of 6 files reviewed, 5 unresolved discussions (waiting on @bobvawter, @BramGruneir, and @Jeremyyang920)

___
*[`internal/script/script_test.go` line 28 at r4](https://reviewable.io/reviews/cockroachdb/replicator/1032#-O8HgKSYD3ZzEeFvGHZP:-O8INg7G1Fqk8d-tgTej:bxo1b19) ([raw file](https://github.com/cockroachdb/replicator/blob/3a8c0b96290531e75ece5ee72e9c18fa05dfe72c/internal/script/script_test.go#L28)):*

Unnecessary alias, here and elsewhere.
</blockquote></details>

I think in some areas, there is already a `base` already used, which is why I decided to use an alias
___
*[`internal/sinktest/base/table_info.go` line 100 at r4](https://reviewable.io/reviews/cockroachdb/replicator/1032#-O8Hgel52ovBmwckerq2:-O8INl7q0UT1Mq9d_ZOx:b-iun300) ([raw file](https://github.com/cockroachdb/replicator/blob/3a8c0b96290531e75ece5ee72e9c18fa05dfe72c/internal/sinktest/base/table_info.go#L100)):*

There's no advantage to passing the predicate in the varargs.
</blockquote></details>

Sorry, could you please clarify what you mean? Where do you think the predicate should be? Or do you think I should just pass the varargs from the outer function straight into the inner QueryRow?
___
*[`internal/sinktest/base/table_info.go` line 116 at r4](https://reviewable.io/reviews/cockroachdb/replicator/1032#-O8HgnMFAIr3XvmGIWoY:-O8INxev2LqAb7IQJwzQ:boumstz) ([raw file](https://github.com/cockroachdb/replicator/blob/3a8c0b96290531e75ece5ee72e9c18fa05dfe72c/internal/sinktest/base/table_info.go#L116)):*

These queries are identical.
</blockquote></details>

Good point. I should probably just consolidate to one query and remove the product type parameter. But then this forces the caller to know how to do the substitution for SQL per dialect, unless I add some more complicated logic to do the v = $ or v = :v, etc.
___
*[`internal/target/apply/apply_test.go` line 95 at r3](https://reviewable.io/reviews/cockroachdb/replicator/1032#gh-1781865464:-O8IOKblDbCaUP-YyHLv:bimati3) ([raw file](https://github.com/cockroachdb/replicator/blob/9dcfc70710579303b68914b5e6acc5d8c32f8a08/internal/target/apply/apply_test.go#L95)):*

Good. That's the goal.
</blockquote></details>

Yep, simplified this.

ryanluu12345 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1032)* status: 0 of 6 files reviewed, 3 unresolved discussions (waiting on @bobvawter, @BramGruneir, and @Jeremyyang920)

___
*[`internal/script/script_test.go` line 28 at r4](https://reviewable.io/reviews/cockroachdb/replicator/1032#-O8HgKSYD3ZzEeFvGHZP:-O8Imeei8drV1sWT5FtF:b-7mifun) ([raw file](https://github.com/cockroachdb/replicator/blob/3a8c0b96290531e75ece5ee72e9c18fa05dfe72c/internal/script/script_test.go#L28)):*

I think in some areas, there is already a `base` already used, which is why I decided to use an alias
</blockquote></details>

Fixed this. I just changed one of the variable inits to `baseSeq` and then fixed the aliasing everywhere.
___
*[`internal/sinktest/base/table_info.go` line 100 at r4](https://reviewable.io/reviews/cockroachdb/replicator/1032#-O8Hgel52ovBmwckerq2:-O8ImjpCDGjzSrMSWfdS:b-1pbwuz) ([raw file](https://github.com/cockroachdb/replicator/blob/3a8c0b96290531e75ece5ee72e9c18fa05dfe72c/internal/sinktest/base/table_info.go#L100)):*

Sorry, could you please clarify what you mean? Where do you think the predicate should be? Or do you think I should just pass the varargs from the outer function straight into the inner QueryRow?
</blockquote></details>

Done! Spoke offline.
___
*[`internal/sinktest/base/table_info.go` line 116 at r4](https://reviewable.io/reviews/cockroachdb/replicator/1032#-O8HgnMFAIr3XvmGIWoY:-O8Iml1M-J_BUBXYZ5Co:b79glk1) ([raw file](https://github.com/cockroachdb/replicator/blob/3a8c0b96290531e75ece5ee72e9c18fa05dfe72c/internal/sinktest/base/table_info.go#L116)):*

Good point. I should probably just consolidate to one query and remove the product type parameter. But then this forces the caller to know how to do the substitution for SQL per dialect, unless I add some more complicated logic to do the v = $ or v = :v, etc.
</blockquote></details>

Yep, removed.

bobvawter (approved):
:lgtm: w/ nits

> Reviewed 6 of 6 files at r5.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1032#-:-O8MhoF58v97BcOMHX0u:b-rjhpvx)* status: all files reviewed (commit messages unreviewed), 2 unresolved discussions (waiting on @BramGruneir, @Jeremyyang920, and @ryanluu12345)

___
*[`-- commits` line 6 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1032#-O8MhOrw2grjCrjrQTBV):*
This is a good commit message.  Please reflow the long lines.  If using `vim` then `:gq <down-arrow>`
___
*[`internal/sinktest/base/table_info.go` line 87 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1032#-O8MguU62PCXyt9jtRU2:-O8MguU71INRqCL0K-HY:bmpjsi3) ([raw file](https://github.com/cockroachdb/replicator/blob/d55ebd5f3bd001f6eacdc0a347a224349ae643bc/internal/sinktest/base/table_info.go#L87)):*
> ```Go
> // GetRowCountWithPredicate returns the number of rows in the table that match the predicate.
> // This method supports N predicates for filtering, passed through args. The first argument is the predicate string
> // (either a raw SQL string or a parameterized one), and the remaining are the arguments for the parameterized query.
> ```

Reflow to 72 chars.

ryanluu12345 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1032)* status: all files reviewed (commit messages unreviewed), 2 unresolved discussions (waiting on @bobvawter, @BramGruneir, and @Jeremyyang920)

___
*[`-- commits` line 6 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1032#-O8MhOrw2grjCrjrQTBV):*

This is a good commit message.  Please reflow the long lines.  If using `vim` then `:gq <down-arrow>`
</blockquote></details>

Fixed!
___
*[`internal/sinktest/base/table_info.go` line 87 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1032#-O8MguU62PCXyt9jtRU2:-O8NPkPKCD5-gZitqT3Q:b-896fja) ([raw file](https://github.com/cockroachdb/replicator/blob/d55ebd5f3bd001f6eacdc0a347a224349ae643bc/internal/sinktest/base/table_info.go#L87)):*

Reflow to 72 chars.
</blockquote></details>

Done!

ryanluu12345 (commented):
Addressed, gonna merge

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1032#-:-O8NQM2h5AVbsQh48KYQ:b-98mnn5)* status: 5 of 6 files reviewed, 2 unresolved discussions (waiting on @bobvawter, @BramGruneir, and @Jeremyyang920)

Issue #995, opened 2024-09-04
MarkApplied can send over-large payloads


bobvawter:
It's possible to exceed `sql.conn.max_read_buffer_message_size` if a bulk operation has been processed which contained millions of rows. The `MarkApplied()` method should segment its work, opening a transaction if one was not already provided. It would be tempting to mark all mutations applied at some timestamp, but this could mask situations in which a row was written to staging out-of-sequence and would not be detected by the consistency check query.

ryanluu12345:
Hi @bobvawter , I understand the gist that we can exceed the limit here and because of that we need to segment work so that each segment can be under the limit.

Is the ask here mainly to break up the work into a set of transactions? Or the same transactions but just different INSERT queries within that transaction?

Also, are there ways to set this max buffer size low so that we can trigger this behavior for testing without doing millions of rows?

bobvawter:
Same transaction. 

https://pkg.go.dev/github.com/cockroachdb/replicator@v1.0.3/internal/util/batches#Window

PR #1033, opened 2024-10-02
stage: updated the MarkApplied so that table updates happen in batches

ryanluu12345:
Integrate the batches.Window functionality from `util.Batches` to split up the work for MarkApplied in order to address concerns with sending payloads that exceed the sql.conn.max_read_buffer_message_size.

The new logic handles the processing of the mutations inside of batches.Window and sets the default window size to 100,000.

Resolves: #995
Release Note: None

**Testing**
- Hardcoded in a batch size of 1
- Inserted in 4 rows
- Verified that they get applied in individual batches
```
DEBUG  [Oct  1 17:45:13] MarkApplied: "_replicator"."public"."molt_public_tbl1"@{NO_FULL_SCAN} marked 1 mutations 
DEBUG  [Oct  1 17:45:15]                                               httpRequest="&{0x14000611c20 45 200 3 4.961584ms   false false}"
DEBUG  [Oct  1 17:45:15] upserted rows                                 conflicts=0 duration=1.053291ms proposed=1 target="\"molt\".\"public\".\"tbl1\"" upserted=1
DEBUG  [Oct  1 17:45:15] MarkApplied: "_replicator"."public"."molt_public_tbl1"@{NO_FULL_SCAN} marked 1 mutations 
DEBUG  [Oct  1 17:45:16] staged mutations                              count=1 duration=4.098209ms target="\"_replicator\".\"public\".\"molt_public_tbl1\"@{NO_FULL_SCAN}"
DEBUG  [Oct  1 17:45:16]                                               httpRequest="&{0x140006fc6c0 125 200 3 4.272833ms   false false}"
DEBUG  [Oct  1 17:45:18]                                               httpRequest="&{0x140006fcb40 45 200 3 3.637875ms   false false}"
DEBUG  [Oct  1 17:45:18] upserted rows                                 conflicts=0 duration="807.875µs" proposed=1 target="\"molt\".\"public\".\"tbl1\"" upserted=1
DEBUG  [Oct  1 17:45:18] MarkApplied: "_replicator"."public"."molt_public_tbl1"@{NO_FULL_SCAN} marked 1 mutations 
DEBUG  [Oct  1 17:45:21]                                               httpRequest="&{0x140006fd7a0 45 200 3 4.040834ms   false false}"
DEBUG  [Oct  1 17:45:21] upserted rows                                 conflicts=0 duration=1.278084ms proposed=1 target="\"molt\".\"public\".\"tbl1\"" upserted=1
DEBUG  [Oct  1 17:45:21] MarkApplied: "_replicator"."public"."molt_public_tbl1"@{NO_FULL_SCAN} marked 1 mutations 
DEBUG  [Oct  1 17:45:24]                                               httpRequest="&{0x140008198c0 45 200 3 4.716833ms   false false}"
DEBUG  [Oct  1 17:45:27]                                               httpRequest="&{0x1400061a120 45 200 3 5.514375ms   false false}"
```
- Verified that the table ends up with mark applied
```
root@127.0.0.1:26257/_replicator> SELECT * FROM molt_public_tbl1;                                                                    
         nanos        | logical |     key     |                         mut                          | before | applied |          applied_at           | deletion |          source_time
----------------------+---------+-------------+------------------------------------------------------+--------+---------+-------------------------------+----------+--------------------------------
  1727829869833460550 |       0 | [1204568]   | \x7b226964223a313230343536382c2274223a22796f227d     | NULL   |    t    | 2024-10-02 00:44:36.032538+00 |    f     | 2024-10-02 00:44:29.83346+00
  1727829908772441943 |       0 | [1204560]   | \x7b226964223a313230343536302c2274223a22796f227d     | NULL   |    t    | 2024-10-02 00:45:13.415785+00 |    f     | 2024-10-02 00:45:08.772442+00
  1727829910926786528 |       0 | [12045601]  | \x7b226964223a31323034353630312c2274223a22796f227d   | NULL   |    t    | 2024-10-02 00:45:15.030812+00 |    f     | 2024-10-02 00:45:10.926787+00
  1727829913129158195 |       0 | [12045604]  | \x7b226964223a31323034353630342c2274223a22796f227d   | NULL   |    t    | 2024-10-02 00:45:18.032043+00 |    f     | 2024-10-02 00:45:13.129158+00
  1727829915248298738 |       0 | [120456046] | \x7b226964223a3132303435363034362c2274223a22796f227d | NULL   |    t    | 2024-10-02 00:45:21.031444+00 |    f     | 2024-10-02 00:45:15.248299+00
(5 rows)
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1033)

### DIFF
--- a/internal/staging/stage/stage.go
+++ b/internal/staging/stage/stage.go
@@ -43,6 +43,10 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
+// defaultMarkApppliedBatchSize is the default number of mutations to mark
+// applied in the staging table at once.
+const defaultMarkAppliedBatchSize = 100_000
+
 // This mutation value will be stored when a mutation is marked as
 // having been applied without having been previously staged.
 var stubSentinel = json.RawMessage(`{"__stub__":true}`)
@@ -59,9 +63,10 @@ func stagingTable(stagingDB ident.Schema, target ident.Table) ident.Table {
 // Mutation instances.
 type stage struct {
 	// The staging table that holds the mutations.
-	stage      *ident.Hinted[ident.Table]
-	stagingDB  *types.StagingPool
-	retireFrom notify.Var[hlc.Time] // Makes subsequent calls to Retire() a bit faster.
+	stage                *ident.Hinted[ident.Table]
+	stagingDB            *types.StagingPool
+	retireFrom           notify.Var[hlc.Time] // Makes subsequent calls to Retire() a bit faster.
+	markAppliedBatchSize int
 
 	consistencyError prometheus.Gauge
 	filterApplied    prometheus.Observer
@@ -161,22 +166,23 @@ ALTER TABLE %s ADD COLUMN IF NOT EXISTS deletion BOOL NULL
 
 	labels := metrics.TableValues(target)
 	s := &stage{
-		stage:            db.HintNoFTS(table),
-		stagingDB:        db,
-		consistencyError: stageConsistencyErrors.WithLabelValues(labels...),
-		filterApplied:    stageFilterAppliedDuration.WithLabelValues(labels...),
-		filterCount:      stageFilterCount.WithLabelValues(labels...),
-		markDuration:     stageMarkDuration.WithLabelValues(labels...),
-		retireDuration:   stageRetireDurations.WithLabelValues(labels...),
-		retireError:      stageRetireErrors.WithLabelValues(labels...),
-		selectCount:      stageSelectCount.WithLabelValues(labels...),
-		selectDuration:   stageSelectDurations.WithLabelValues(labels...),
-		selectError:      stageSelectErrors.WithLabelValues(labels...),
-		staleCount:       stageStaleMutations.WithLabelValues(labels...),
-		stageCount:       stageCount.WithLabelValues(labels...),
-		stageDupes:       stageDuplicateCount.WithLabelValues(labels...),
-		stageDuration:    stageDuration.WithLabelValues(labels...),
-		stageError:       stageErrors.WithLabelValues(labels...),
+		stage:                db.HintNoFTS(table),
+		stagingDB:            db,
+		markAppliedBatchSize: defaultMarkAppliedBatchSize,
+		consistencyError:     stageConsistencyErrors.WithLabelValues(labels...),
+		filterApplied:        stageFilterAppliedDuration.WithLabelValues(labels...),
+		filterCount:          stageFilterCount.WithLabelValues(labels...),
+		markDuration:         stageMarkDuration.WithLabelValues(labels...),
+		retireDuration:       stageRetireDurations.WithLabelValues(labels...),
+		retireError:          stageRetireErrors.WithLabelValues(labels...),
+		selectCount:          stageSelectCount.WithLabelValues(labels...),
+		selectDuration:       stageSelectDurations.WithLabelValues(labels...),
+		selectError:          stageSelectErrors.WithLabelValues(labels...),
+		staleCount:           stageStaleMutations.WithLabelValues(labels...),
+		stageCount:           stageCount.WithLabelValues(labels...),
+		stageDupes:           stageDuplicateCount.WithLabelValues(labels...),
+		stageDuration:        stageDuration.WithLabelValues(labels...),
+		stageError:           stageErrors.WithLabelValues(labels...),
 	}
 
 	// Prevent these hot-path queries from being planned with a full
@@ -551,6 +557,12 @@ func (s *stage) stageOneBatch(
 	return nil
 }
 
+// SetMarkAppliedBatchSize is used for testing and helps set the mark applied
+// batch size so we can verify the behavior of different batch sizes.
+func (s *stage) SetMarkAppliedBatchSize(size int) {
+	s.markAppliedBatchSize = size
+}
+
 const markAppliedTemplate = `
 WITH t (key, nanos, logical) AS (SELECT unnest($1::STRING[]), unnest($2::INT8[]), unnest($3::INT8[]))
 INSERT INTO %s (key, nanos, logical, applied, applied_at, mut)
@@ -575,8 +587,15 @@ func (s *stage) MarkApplied(
 	return retry.Retry(ctx, s.stagingDB, func(ctx context.Context) error {
 		start := time.Now()
 
+		// 1) number of mutations larger than the `markAppliedBatchSize`: we
+		// want to introduce a transaction here to keep the updates for a
+		// batch's apply times atomic. However, we only want to do this if there
+		// is not already a transaction.
+		//
+		// 2) extraSanityChecks: we need a transaction in order to check if the
+		// data is consistent.
 		var tx pgx.Tx
-		if extraSanityChecks {
+		if extraSanityChecks || len(muts) > s.markAppliedBatchSize {
 			if _, isTx := db.(pgx.Tx); !isTx {
 				var err error
 				tx, err = s.stagingDB.Begin(ctx)
@@ -588,10 +607,32 @@ func (s *stage) MarkApplied(
 			}
 		}
 
-		tag, err := db.Exec(ctx, s.sql.markApplied, keys, nanos, logical)
-		if err != nil {
-			return errors.Wrap(err, s.sql.markApplied)
+		// Applies the mutations in batches to avoid exceeding the
+		// `sql.conn.max_read_buffer_message_size` This also reduces
+		// the memory being used during this step in the case there
+		// are millions or more rows.
+		if err := batches.Window(s.markAppliedBatchSize, len(muts), func(begin, end int) error {
+			lenWindow := end - begin
+			keys := make([]json.RawMessage, lenWindow)
+			nanos := make([]int64, lenWindow)
+			logical := make([]int, lenWindow)
+			for idx, mut := range muts[begin:end] {
+				keys[idx] = mut.Key
+				nanos[idx] = mut.Time.Nanos()
+				logical[idx] = mut.Time.Logical()
+			}
+
+			tag, err := db.Exec(ctx, s.sql.markApplied, keys, nanos, logical)
+			if err != nil {
+				return errors.Wrap(err, s.sql.markApplied)
+			}
+
+			log.Tracef("MarkApplied: %s marked %d mutations", s.stage, tag.RowsAffected())
+			return nil
+		}); err != nil {
+			return err
 		}
+
 		if extraSanityChecks {
 			count, err := s.CheckConsistency(ctx, db, muts, false /* current-time read */)
 			if err != nil {
@@ -602,7 +643,6 @@ func (s *stage) MarkApplied(
 			}
 		}
 		s.markDuration.Observe(time.Since(start).Seconds())
-		log.Tracef("MarkApplied: %s marked %d mutations", s.stage, tag.RowsAffected())
 		if tx != nil {
 			return errors.WithStack(tx.Commit(ctx))
 		}
--- a/internal/staging/stage/stage_test.go
+++ b/internal/staging/stage/stage_test.go
@@ -83,6 +83,37 @@ func TestFilterStubs(t *testing.T) {
 
 // TestPutAndDrain will insert and mark a batch of Mutations.
 func TestPutAndDrain(t *testing.T) {
+	tc := []struct {
+		name           string
+		totalMutations int
+		batchSize      int
+	}{
+		{
+			name:           "batch is larger than total mutations",
+			totalMutations: 500,
+			batchSize:      1_000,
+		},
+		{
+			name:           "full batches only",
+			totalMutations: 1_000,
+			batchSize:      500,
+		},
+		{
+			name:           "full batches and partial batch",
+			totalMutations: 1_200,
+			batchSize:      500,
+		},
+	}
+
+	for _, tc := range tc {
+		t.Run(tc.name, func(t *testing.T) {
+			testPutAndDrain(t, tc.totalMutations, tc.batchSize)
+		})
+	}
+
+}
+
+func testPutAndDrain(t *testing.T, total int, testBatchSize int) {
 	a := assert.New(t)
 	r := require.New(t)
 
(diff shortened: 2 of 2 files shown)


internal/staging/stage/stage.go
```diff
@@ -551,6 +551,7 @@ func (s *stage) stageOneBatch(
 	return nil
 }
 
+const markAppliedBatchSize = 100_000
```
ryanluu12345, line 554: Open questions here:
1. What is a good sane default? I seem to understand that millions to couple of millions is when we can start hitting the limits of the SQL cluster setting. But not sure if this is too low.
2. Should this be configurable? I defaulted not since this is a property of the staging logic which shouldn't be managed by the user.

internal/staging/stage/stage.go
```diff
@@ -588,10 +589,29 @@ func (s *stage) MarkApplied(
...
+			keys := make([]json.RawMessage, len(muts))
+			nanos := make([]int64, len(muts))
+			logical := make([]int, len(muts))
+			for idx, mut := range muts[begin:end] {
+				keys[idx] = mut.Key
+				nanos[idx] = mut.Time.Nanos()
+				logical[idx] = mut.Time.Logical()
+			}
+
+			tag, err := db.Exec(ctx, s.sql.markApplied, keys, nanos, logical)
```
ryanluu12345, line 604: From my read of the code `db` can be a `tx` if we want `extraSanityChecks == true`. This is why I don't do any rollback or TX committing here.

However, in the ticket, @bobvawter  did mention that you wanted the caller of this to pass in a TX and create it if it doesn't exist. To clarify, did you want to change this logic so that it's unconditionally a transaction instead of potentially separate db.Exec calls? 

I understand that it is good to keep this a transaction now that we update in batches because we want to update these atomically, just like we previously were technically. If that's the case, I can just make it unconditionally a transaction and make the `extraSanityChecks` merely check consistency, if you don't see issues with that.

bobvawter (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1033)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @Jeremyyang920 and @ryanluu12345)

___
*[`internal/staging/stage/stage.go` line 554 at r1](https://reviewable.io/reviews/cockroachdb/replicator/1033#gh-1783700407:-O8CIkr13JUQOFrPOqd-:b-9pr8b8) ([raw file](https://github.com/cockroachdb/replicator/blob/2bde3b81f89c1bdbad8f067153c785bf8cc098bb/internal/staging/stage/stage.go#L554)):*

Open questions here:
1. What is a good sane default? I seem to understand that millions to couple of millions is when we can start hitting the limits of the SQL cluster setting. But not sure if this is too low.
2. Should this be configurable? I defaulted not since this is a property of the staging logic which shouldn't be managed by the user.
</blockquote></details>

That seems like a reasonable default.  It's far larger than any interactive OLTP workload would tend to generate. This change is to address bulk transactions.
___
*[`internal/staging/stage/stage.go` line 592 at r1](https://reviewable.io/reviews/cockroachdb/replicator/1033#-O8CKETX6QbPrGnBv7RK:-O8CKETX6QbPrGnBv7RL:b-d0ikkw) ([raw file](https://github.com/cockroachdb/replicator/blob/2bde3b81f89c1bdbad8f067153c785bf8cc098bb/internal/staging/stage/stage.go#L592)):*
> ```Go
> 		}
> 
> 		// Applies the mutations in batches to avoid exceeding the `sql.conn.max_read_buffer_message_size`
> ```

Reflow comments to 72 chars. The GoLand "Wrap to Column" plugin is my preferred tool.
___
*[`internal/staging/stage/stage.go` line 595 at r1](https://reviewable.io/reviews/cockroachdb/replicator/1033#-O8CKWHe3ec3bk7pUwBS:-O8CKWHe3ec3bk7pUwBT:b764zky) ([raw file](https://github.com/cockroachdb/replicator/blob/2bde3b81f89c1bdbad8f067153c785bf8cc098bb/internal/staging/stage/stage.go#L595)):*
> ```Go
> 		// This also reduces the memory being used during this step in the case there are millions or more rows.
> 		if err := batches.Window(markAppliedBatchSize, len(muts), func(begin, end int) error {
> 			keys := make([]json.RawMessage, len(muts))
> ```

These slices are going to be full of empty values if windowing happens. The length should be `end-begin`.
___
*[`internal/staging/stage/stage.go` line 604 at r1](https://reviewable.io/reviews/cockroachdb/replicator/1033#gh-1783704839:-O8CH9h82wIUoCrVwjlN:bg88u6c) ([raw file](https://github.com/cockroachdb/replicator/blob/2bde3b81f89c1bdbad8f067153c785bf8cc098bb/internal/staging/stage/stage.go#L604)):*

From my read of the code `db` can be a `tx` if we want `extraSanityChecks == true`. This is why I don't do any rollback or TX committing here.

However, in the ticket, @bobvawter  did mention that you wanted the caller of this to pass in a TX and create it if it doesn't exist. To clarify, did you want to change this logic so that it's unconditionally a transaction instead of potentially separate db.Exec calls? 

I understand that it is good to keep this a transaction now that we update in batches because we want to update these atomically, just like we previously were technically. If that's the case, I can just make it unconditionally a transaction and make the `extraSanityChecks` merely check consistency, if you don't see issues with that.
</blockquote></details>

There's also
https://github.com/cockroachdb/replicator/blob/master/internal/sequencer/decorators/marker.go
and a handful of other places where a staging transaction may be created when multiple tables are in play.  The thing to do here is to ensure there's a transaction if the number of marks exceeds the window size.  We want to be able to use implicit transactions, unless there's a need to do something else.

ryanluu12345 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1033)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @bobvawter and @Jeremyyang920)

___
*[`internal/staging/stage/stage.go` line 554 at r1](https://reviewable.io/reviews/cockroachdb/replicator/1033#gh-1783700407:-O8EY0mSExISkv6sn-vM:b-o246i8) ([raw file](https://github.com/cockroachdb/replicator/blob/2bde3b81f89c1bdbad8f067153c785bf8cc098bb/internal/staging/stage/stage.go#L554)):*

That seems like a reasonable default.  It's far larger than any interactive OLTP workload would tend to generate. This change is to address bulk transactions.
</blockquote></details>

Ok great. Then I'll go ahead and leave it at this default. If configuration is needed later, we can drill it in, but unnecessary for now.
___
*[`internal/staging/stage/stage.go` line 592 at r1](https://reviewable.io/reviews/cockroachdb/replicator/1033#-O8CKETX6QbPrGnBv7RK:-O8EX4pGBeFQIHYcOOs1:b-1tod1c) ([raw file](https://github.com/cockroachdb/replicator/blob/2bde3b81f89c1bdbad8f067153c785bf8cc098bb/internal/staging/stage/stage.go#L592)):*

Reflow comments to 72 chars. The GoLand "Wrap to Column" plugin is my preferred tool.
</blockquote></details>

Noted, just installed `Rewrap` and will use that for my comment formatting. Note that I'll use `crlfmt` for my coding needs.
___
*[`internal/staging/stage/stage.go` line 595 at r1](https://reviewable.io/reviews/cockroachdb/replicator/1033#-O8CKWHe3ec3bk7pUwBS:-O8EXjgR1XqAuDqiVqUl:b-gj8tya) ([raw file](https://github.com/cockroachdb/replicator/blob/2bde3b81f89c1bdbad8f067153c785bf8cc098bb/internal/staging/stage/stage.go#L595)):*

These slices are going to be full of empty values if windowing happens. The length should be `end-begin`.
</blockquote></details>

Great catch!

ryanluu12345 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1033)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @bobvawter and @Jeremyyang920)

___
*[`internal/staging/stage/stage.go` line 604 at r1](https://reviewable.io/reviews/cockroachdb/replicator/1033#gh-1783704839:-O8EfMxQ4YkZAkr8X3Xl:b-rn101s) ([raw file](https://github.com/cockroachdb/replicator/blob/2bde3b81f89c1bdbad8f067153c785bf8cc098bb/internal/staging/stage/stage.go#L604)):*

There's also
https://github.com/cockroachdb/replicator/blob/master/internal/sequencer/decorators/marker.go
and a handful of other places where a staging transaction may be created when multiple tables are in play.  The thing to do here is to ensure there's a transaction if the number of marks exceeds the window size.  We want to be able to use implicit transactions, unless there's a need to do something else.
</blockquote></details>

Ok ended up fixing this! Please let me know your thoughts. I bundled this logic with the block above that creates a transaction for the extraSanityChecks. The logic here is that we need a new transaction in the two cases where we want to do consistency checks or we want to do multiple batches of applies but do so atomically.

bobvawter (changes requested):
The implementation looks good, but it needs a test before merging.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1033#-:-O8Hf8wV2-wi7_hEUhFp:b5uswgd)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @Jeremyyang920 and @ryanluu12345)

___
*[`internal/staging/stage/stage.go` line 604 at r1](https://reviewable.io/reviews/cockroachdb/replicator/1033#gh-1783704839:-O8Heju_7GMHPR31fYnP:bxxx4gf) ([raw file](https://github.com/cockroachdb/replicator/blob/2bde3b81f89c1bdbad8f067153c785bf8cc098bb/internal/staging/stage/stage.go#L604)):*

Ok ended up fixing this! Please let me know your thoughts. I bundled this logic with the block above that creates a transaction for the extraSanityChecks. The logic here is that we need a new transaction in the two cases where we want to do consistency checks or we want to do multiple batches of applies but do so atomically.
</blockquote></details>

That makes sense.
___
*[`internal/staging/stage/stage.go` line 619 at r3](https://reviewable.io/reviews/cockroachdb/replicator/1033#-O8HeqtH2WDAjxZ2Q61o:-O8HeqtH2WDAjxZ2Q61p:bkmdifu) ([raw file](https://github.com/cockroachdb/replicator/blob/3eead5d154918e629e6f0905d87be34642484242/internal/staging/stage/stage.go#L619)):*
> ```Go
> 			}
> 
> 			log.Debugf("MarkApplied: %s marked %d mutations", s.stage, tag.RowsAffected())
> ```

Drop this back to trace; it's going to be very spammy in production.

ryanluu12345 (commented):
Finished the tests. PR should be ready!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1033#-:-O8JOWfH5otQlrD2Adl_:b-vlhi5r)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @bobvawter and @Jeremyyang920)

bobvawter (approved):
:lgtm: w/ nits

> Reviewed 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1033#-:-O8MgJ_K4cVtYDjh0K8W:b-rjhpvx)* status: all files reviewed, 3 unresolved discussions (waiting on @Jeremyyang920 and @ryanluu12345)

___
*[`internal/staging/stage/stage.go` line 560 at r4](https://reviewable.io/reviews/cockroachdb/replicator/1033#-O8MfdsH1NsDB-4pcBQq:-O8MfdsH1NsDB-4pcBQr:b-aa25ec) ([raw file](https://github.com/cockroachdb/replicator/blob/b75fc012482f06331d36a92a0a8299305d2f54ae/internal/staging/stage/stage.go#L560)):*
> ```Go
> }
> 
> // These implementations are used for testing and help to set the mark applied
> ```

SetMarkAppliedBatchSize is used for testing ....
___
*[`internal/staging/stage/stage.go` line 566 at r4](https://reviewable.io/reviews/cockroachdb/replicator/1033#-O8Mf208CoJOidbOGNNp:-O8Mf2096cHFhG-D5OuT:bm2242s) ([raw file](https://github.com/cockroachdb/replicator/blob/b75fc012482f06331d36a92a0a8299305d2f54ae/internal/staging/stage/stage.go#L566)):*
> ```Go
> }
> 
> func (s *stage) GetMarkAppliedBatchSize() int {
> ```

This method is unused.
___
*[`internal/staging/stage/stage.go` line 583 at r4](https://reviewable.io/reviews/cockroachdb/replicator/1033#-O8MeyiU52Hw8-SJ79qg:-O8MeyiU52Hw8-SJ79qh:b-imhtee) ([raw file](https://github.com/cockroachdb/replicator/blob/b75fc012482f06331d36a92a0a8299305d2f54ae/internal/staging/stage/stage.go#L583)):*
> ```Go
> 	ctx context.Context, db types.StagingQuerier, muts []types.Mutation,
> ) error {
> 	fmt.Println(s.markAppliedBatchSize)
> ```

Dangling log statement.

ryanluu12345 (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1033#-:-O8NRLWe8Yejr6CRnDLA:bnfp4nl)* status: all files reviewed, 3 unresolved discussions (waiting on @bobvawter and @Jeremyyang920)

___
*[`internal/staging/stage/stage.go` line 560 at r4](https://reviewable.io/reviews/cockroachdb/replicator/1033#-O8MfdsH1NsDB-4pcBQq:-O8NQsZnA7y_kTs-yhbz:bb7f2mf) ([raw file](https://github.com/cockroachdb/replicator/blob/b75fc012482f06331d36a92a0a8299305d2f54ae/internal/staging/stage/stage.go#L560)):*

SetMarkAppliedBatchSize is used for testing ....
</blockquote></details>

done
___
*[`internal/staging/stage/stage.go` line 566 at r4](https://reviewable.io/reviews/cockroachdb/replicator/1033#-O8Mf208CoJOidbOGNNp:-O8NQt3_1tUZlfm5NBp8:bb7f2mf) ([raw file](https://github.com/cockroachdb/replicator/blob/b75fc012482f06331d36a92a0a8299305d2f54ae/internal/staging/stage/stage.go#L566)):*

This method is unused.
</blockquote></details>

done
___
*[`internal/staging/stage/stage.go` line 583 at r4](https://reviewable.io/reviews/cockroachdb/replicator/1033#-O8MeyiU52Hw8-SJ79qg:-O8NQvwE9xx9541YulGD:bb7f2mf) ([raw file](https://github.com/cockroachdb/replicator/blob/b75fc012482f06331d36a92a0a8299305d2f54ae/internal/staging/stage/stage.go#L583)):*

Dangling log statement.
</blockquote></details>

done

Issue #720, opened 2024-02-23
Improve error message when sending a changefeed query to non-query endpoint


bobvawter:
If a changefeed query payload is sent to a non-query path (i.e. it's missing the table path element), the following error is returned: `json: unknown field "__crdb__"`. We should add a special-case here to suggest the correct course of action if we see the `__crdb__` field.

ryanluu12345:
Hi @bobvawter , just to clarify here:
1. When you say that it is missing the table path element, what exactly do you mean? In the example below, do you mean it's missing the `FOR TABLE` or the `defaultdb/public` piece?
```
CREATE CHANGEFEED FOR TABLE public.tbl1 INTO 'webhook-https://localhost:30004/defaultdb/public?insecure_tls_skip_verify=true'
```

2. And for the correct course of action, my understanding is to alert them to include the table path (i.e. the `defaultdb/public` and a generic example. Is this a correct assumption?

bobvawter:
A changefeed query must be sent to a path like `/database/schema/table`, since a query doesn't implicitly have a table name.

X-ref: https://github.com/cockroachdb/replicator/wiki/C2C#change-data-capture-queries

ryanluu12345:
Reproduction notes (for reference)
1. I was able to verify that when running with this configuration, I'm able to ensure that messages can be sourced from Cockroach changefeeds and applied to non-homogeneous targets:
```
# Replicator run
go run . start --bindAddr 0.0.0.0:30004 --targetConn 'mysql://root:@localhost:3306/molt?sslmode=disable' --stagingConn 'postgres://root@localhost:26257/molt?sslmode=disable' --stagingSchema _replicator --disableAuthentication --tlsSelfSigned -v

# Changefeed config on CRDB source
root@127.0.0.1:26257/molt> CREATE CHANGEFEED                                                                                 
                        -> INTO 'webhook-https://host.docker.internal:30004/molt/tbl1?insecure_tls_skip_verify=true'         
                        -> WITH envelope='wrapped', format='json', diff, updated, resolved='1s',                             
                        -> min_checkpoint_frequency='1s',                                                                    
                        ->      webhook_sink_config='{"Flush":{"Bytes":1048576,"Frequency":"1s"}}'                           
                        -> AS SELECT                                                                                         
                        ->      *                                                                                            
                        -> FROM                                                                                              
                        ->      molt.public.tbl1                                                                             
                        -> ;   
```

[...]

PR #1035, opened 2024-10-04
ident: made the empty table logging more clear when the CDC query webhook URI doesn't contain the table

ryanluu12345:
Resolves: #720
Release Note: None

**Testing**
```
go test -v -run ^TestIntegration$ github.com/cockroachdb/replicator/internal/source/cdc/server | tee test.log
...
{
  "error": "table name is empty, please ensure the table name is included in the path",
  "severity": "ERROR",
  "timestamp": {
    "seconds": 1728000632,
    "nanos": 377887000
  },
  "uri": "/tgt-74858-58/public"
}
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1035)

### DIFF
--- a/internal/source/cdc/handler_test.go
+++ b/internal/source/cdc/handler_test.go
@@ -827,3 +827,30 @@ func TestSyntheticWebhooks(t *testing.T) {
 	r.NoError(gen.WaitForCatchUp(ctx, targetInfo.Stat()))
 	gen.CheckConsistent(fixture.Context, t)
 }
+
+func TestSyntheticWebhooksTopicEmpty(t *testing.T) {
+	r := require.New(t)
+	fixture, _ := createFixture(t, &fixtureConfig{})
+	ctx := fixture.Context
+
+	gen, _, err := fixture.NewWorkload(ctx, &all.WorkloadConfig{})
+	r.NoError(err)
+
+	batch := &types.MultiBatch{}
+	gen.GenerateInto(batch, hlc.New(1, 2))
+
+	requestPath := "/" + ident.Join(fixture.TargetSchema, ident.Raw, '/')
+
+	// Send data.
+	payload, err := NewWebhookPayload(batch)
+	payload.Payload[0].Topic = ""
+	r.NoError(err)
+	payloadBytes, err := json.Marshal(payload)
+	r.NoError(err)
+	req := httptest.NewRequest("POST", requestPath,
+		bytes.NewReader(payloadBytes)).WithContext(ctx)
+	resp := httptest.NewRecorder()
+	fixture.Handler.ServeHTTP(resp, req)
+	r.Contains(resp.Body.String(), "table name is empty")
+	r.Equal(http.StatusBadRequest, resp.Code)
+}
--- a/internal/source/cdc/webhook.go
+++ b/internal/source/cdc/webhook.go
@@ -114,8 +114,9 @@ func (h *Handler) webhook(ctx context.Context, req *request) error {
 			return err
 		}
 
-		// In the webhook case, if the payload topic is that means the table was
-		// not passed in properly.
+		// If the topic is empty, we have received a CDC query payload and the
+		// user has not specified a three-segment request path to set a table
+		// name.
 		if payload.Payload[i].Topic == "" {
 			return errors.New("table name is empty, please ensure the table name is included in the path")
 		}
--- a/internal/source/cdc/webhook.go
+++ b/internal/source/cdc/webhook.go
@@ -114,6 +114,12 @@ func (h *Handler) webhook(ctx context.Context, req *request) error {
 			return err
 		}
 
+		// In the webhook case, if the payload topic is that means the table was
+		// not passed in properly.
+		if payload.Payload[i].Topic == "" {
+			return errors.New("table name is empty, please ensure the table name is included in the path")
+		}
+
 		table, qual, err := ident.ParseTableRelative(payload.Payload[i].Topic, req.target.Schema())
 		if err != nil {
 			return err


internal/util/ident/parse.go
```diff
@@ -155,7 +155,7 @@ func ParseTableRelative(s string, relativeTo Schema) (Table, Qualification, erro
 	}
 
 	if len(parts) == 0 {
-		return Table{}, 0, errors.New("empty table name")
+		return Table{}, 0, errors.New("table name is empty, please ensure the table name is included in the path")
```
ryanluu12345, line 158: @bobvawter please let me know how I can make this more clear and actionable as needed.

bobvawter (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1035)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @ryanluu12345)

___
*[`internal/util/ident/parse.go` line 158 at r1](https://reviewable.io/reviews/cockroachdb/replicator/1035#gh-1786959412:-O8MdKhT2HIvESjU8KgZ:b-7611ni) ([raw file](https://github.com/cockroachdb/replicator/blob/d7f5e002773694153b11ece95ecd2e173cce4eb1/internal/util/ident/parse.go#L158)):*

@bobvawter please let me know how I can make this more clear and actionable as needed.
</blockquote></details>

Parse doesn't know where it's being called from; this error message could very well wind up being misleading in some other troubleshooting context that has nothing to do with a missing path element.  For example, this function is reachable from the script package when the results of a `dispatch()` call are being routed.

The code which sends the empty `s` value in, having failed to extract something useful from the path, should be the thing to signal a causal concern.  It could do that either with an `if s == ""` check or an `errors.Is(err, ident.ErrEmptyName)` approach.

ryanluu12345:
Good notes here @bobvawter , I went ahead and checked for the empty string before erroring out and I also do this in the cdc internal server webhook handler.

bobvawter:
Seems reasonable. Check out `handler_test.go` for ways to test this without needing to deal with a complete HTTP server.  `TestSyntheticWebhooks()` is probably your best template to start from.

ryanluu12345:
@bobvawter , noted. What are your thoughts on adding in a `testKnobs` parameter here:
```
// NewWebhookPayload constructs a payload for the batch.
func NewWebhookPayload(batch *types.MultiBatch) (*WebhookPayload, error) {
	ret := &WebhookPayload{}
	// Ignoring error since callback returns nil.
	if err := batch.CopyInto(types.AccumulatorFunc(func(table ident.Table, mut types.Mutation) error {
		line := WebhookPayloadLine{
			After:   mut.Data,
			Before:  mut.Before,
			Key:     mut.Key,
			Topic:   table.Raw(),
			Updated: mut.Time.String(),
		}
		// Webhook encodes a deletion as an absence of an after block.
		if mut.IsDelete() {
			line.After = nil
		}
		ret.Payload = append(ret.Payload, line)
		ret.Range = ret.Range.Extend(mut.Time)
		return nil
	})); err != nil {
		return nil, err
	}
	ret.Length = len(ret.Payload)
	return ret, nil
}
```
That way we can mock the `Topic` so that it's an empty string. Then we can simulate this behavior that we are seeing.

bobvawter (approved):
:lgtm: w/ nits. 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1035#-:-O8cPxTc-U0i_rNUiBZ3:br70gx1)* status: 0 of 4 files reviewed, 3 unresolved discussions (waiting on @ryanluu12345)

___
*[`internal/source/cdc/handler_test.go` line 849 at r3](https://reviewable.io/reviews/cockroachdb/replicator/1035#-O8cHS87BgkQJy-E8lqK:-O8cHS87BgkQJy-E8lqL:b-pjb1zx) ([raw file](https://github.com/cockroachdb/replicator/blob/11febaad2e323f25944e569475ea46907875971b/internal/source/cdc/handler_test.go#L849)):*
> ```Go
> 
> 	batch := &types.MultiBatch{}
> 	for i := range batches {
> ```

You don't really need to create any more than a single enclosed batch.

```
batch := &types.MultiBatch{}
gen.GenerateInto()
batch.Data[0].Topic = ""
```
___
*[`internal/source/cdc/handler_test.go` line 856 at r3](https://reviewable.io/reviews/cockroachdb/replicator/1035#-O8cNz-P1g7kNJYk2Q8p:-O8cNz-P1g7kNJYk2Q8q:bd6rmt9) ([raw file](https://github.com/cockroachdb/replicator/blob/11febaad2e323f25944e569475ea46907875971b/internal/source/cdc/handler_test.go#L856)):*
> ```Go
> 
> 	// Send data.
> 	{
> ```

The nested blocks in the original method are only there because there are back-to-back operations that I wanted to make sure couldn't bleed into one anther.  In this case, just hoist the code out since the nested scope doesn't really buy you anything.
___
*[`internal/source/cdc/webhook.go` line 121 at r3](https://reviewable.io/reviews/cockroachdb/replicator/1035#-O8cOFPYDz3Joi9B9Ltc:-O8cOFPYDz3Joi9B9Ltd:b5p83b) ([raw file](https://github.com/cockroachdb/replicator/blob/11febaad2e323f25944e569475ea46907875971b/internal/source/cdc/webhook.go#L121)):*
> If the topic is empty, we have received a CDC query payload and the user has not specified a three-segment request path to set a table name.

ryanluu12345 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1035)* status: 0 of 4 files reviewed, 3 unresolved discussions (waiting on @bobvawter)

___
*[`internal/source/cdc/handler_test.go` line 849 at r3](https://reviewable.io/reviews/cockroachdb/replicator/1035#-O8cHS87BgkQJy-E8lqK:-O8cxJbu2prWBLk6P_p_:b-896fix) ([raw file](https://github.com/cockroachdb/replicator/blob/11febaad2e323f25944e569475ea46907875971b/internal/source/cdc/handler_test.go#L849)):*

You don't really need to create any more than a single enclosed batch.

```
batch := &types.MultiBatch{}
gen.GenerateInto()
batch.Data[0].Topic = ""
```
</blockquote></details>

Done.
___
*[`internal/source/cdc/handler_test.go` line 856 at r3](https://reviewable.io/reviews/cockroachdb/replicator/1035#-O8cNz-P1g7kNJYk2Q8p:-O8cxJ7e9xfr40KmHm-B:b-896fix) ([raw file](https://github.com/cockroachdb/replicator/blob/11febaad2e323f25944e569475ea46907875971b/internal/source/cdc/handler_test.go#L856)):*

The nested blocks in the original method are only there because there are back-to-back operations that I wanted to make sure couldn't bleed into one anther.  In this case, just hoist the code out since the nested scope doesn't really buy you anything.
</blockquote></details>

Done.
___
*[`internal/source/cdc/webhook.go` line 121 at r3](https://reviewable.io/reviews/cockroachdb/replicator/1035#-O8cOFPYDz3Joi9B9Ltc:-O8cxFrF387A7ckPadKZ:b-896fix) ([raw file](https://github.com/cockroachdb/replicator/blob/11febaad2e323f25944e569475ea46907875971b/internal/source/cdc/webhook.go#L121)):*

> If the topic is empty, we have received a CDC query payload and the user has not specified a three-segment request path to set a table name.
</blockquote></details>

Done.

Issue #906, opened 2024-06-07
Update cdc/server/integration_test.go with workload.Checker


bobvawter:
This is a pretty old test and could stand to be refreshed with the newer `workload.Checker`.

ryanluu12345:
Is this the intended scope of this change to rewrite this test so that it follows the same `Workload` convention in `internal/source/kafka/integration_test.go` ? Basically, rewriting it so that it does all aspects of the test with the workload struct helpers? Or is it just to use the `workload.Checker` for verifying the results are reasonable?

bobvawter:
If not rewrite, then to add another test method that inserts into the source tables using data derived from the generator and then validate the target tables with the checker.

ryanluu12345:
Ok sounds like a plan. After digging a bit deeper into the code today, I think I have a general understanding of the workflow here:
1. Instantiate a new `all.Fixture` which gives us access to the `NewWorkload` helper that creates a workload on our behalf
2. Get the configuration for the type of replicator mode we are trying to run (in this case CDC)
3. Create a new "producer" which is essentially what writes the data to the source (also thinking in the case of CDC, we can start the changefeed job here for our workload table)
4. Start the replicator service
5. Inside of a loop, generate batches into a multi batch and then later on write the batch we generate using the "producer" from above
6. Use the workload checker to wait for catch up and then check all counters + target data matching source data

Please let me know if I'm misunderstanding any of these steps.

I do have a few open questions:
1. In the case of the CDC case, the "producer" isn't a message queue like kafka or object store like s3, but rather the source database. Is it a correct assumption that we need to create a new producer that has access to the sourceFixture and can create the changefeed that ends up sending data to the target?
2. I see that in the other cases of kafka and objstore, there is a `Start` method that gets called. Looking into `source/cdc/server` I don't see an equivalent, but I do see the `newTestFixture` that seems to do the same thing: wire everything up and return a server. Would this be the one I use in place of the `Start` That I see in the other examples. Code from the kafka example for reference:
```
	serverCfg, err := getConfig(destFixture, fc, []string{target.Raw()}, target)
	r.NoError(err)
	timeoutCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
	defer cancel()
	connCtx := stopper.WithContext(timeoutCtx)
	kafka, err := Start(connCtx, serverCfg)
	r.NoError(err)

```
[...]

PR #1042, opened 2024-10-11
cdc/server/integration_test.go: added in a new test for the CDC server using the workload.Checker

ryanluu12345:
Integrated the logic from the existing CDC server integration test and the latest developments for the pglogical integration tests to create a test based on the workload.Checker that validates the CDC server. This implementation also aims to refactor the Changefeed creation string logic so that it's reusable for other types of tests or the workload command.
    
Resolves: #906
Release Note: None

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1042)


ryanluu12345 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1042)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @bobvawter and @Jeremyyang920)

___
*[`internal/source/cdc/server/integration_test.go` line 455 at r1](https://reviewable.io/reviews/cockroachdb/replicator/1042#-O8td0NP8BbpPDNDCdfj:-O8td0NP8BbpPDNDCdfk:b-6fz4d8) ([raw file](https://github.com/cockroachdb/replicator/blob/202c58124e615931f286650e326d9f0602c75020/internal/source/cdc/server/integration_test.go#L455)):*
> ```Go
> // changefeed. This makes it so this is more portable for future work that needs
> // this: for example, integrating the e2e workload checker with this logic.
> func CreateChangefeedStatement(
> ```

The plan here is that when we tackle https://github.com/cockroachdb/replicator/issues/926, we refactor this to a common location that both can access so that we have a helper to create a changefeed that is compatible with the `workload` helpers.

ryanluu12345:
Discussed offline but @bobvawter and I were considering using Go templates or the `cockroachdb-parser/tree` in order to create the changefeed statements. This is not scoped as part of this PR, but can be adapted for when we port this over the workload demo.

Additionally, all helpers will subsequently be moved over to `sinktest` which is where the helpers all live right now.

ryanluu12345:
@bobvawter , I'm currently digging into these failures and I have identified some issues that I'm patching locally:
1. Webhook option not found because of unsupported version -> brought in the min version checks from `testIntegration`
2. Data seems to be inconsistent for V23.2 tests; I thought this may be a case of all the data not being written yet, but it looks like the correct number of items got written: 8 items (7 + 1 expected; 6 + 2 got). Not sure if a parent can get accidentally written to a child.....
```
workload.go:130: 
        	Error Trace:	/home/runner/work/replicator/replicator/internal/sinktest/all/workload.go:130
        	            				/home/runner/work/replicator/replicator/internal/source/cdc/server/integration_test.go:740
        	            				/home/runner/work/replicator/replicator/internal/source/cdc/server/integration_test.go:618
        	Error:      	Should be empty, but was [expected 7 parents, got 6 expected 1 children, got 2 parent 8003786112974678833 not found parent 2022766499421304972 not found parent 255718460[8026](https://github.com/cockroachdb/replicator/actions/runs/11285840194/job/31389309792?pr=1042#step:7:8027)950244 not found]
        	Test:       	TestWorkload
    integration_test.go:740: 
        	Error Trace:	/home/runner/work/replicator/replicator/internal/source/cdc/server/integration_test.go:740
        	            				/home/runner/work/replicator/replicator/internal/source/cdc/server/integration_test.go:618
        	Error:      	Should be true
        	Test:       	TestWorkload
```
3. Unknown schema for the source, even though I can clearly see that it is defined:
- Integration test: CRDB 24.1 and permuations of each CRDB version
- Integration test: CRDB 24.1 and MySQL / MariaDB / Oracle

What I'm thinking here is that for some reason it's the wrong schema name and this should actually be the target.... I'm not exactly sure what to make of this and when I should be using the "target" schema for this and "source" schema.

```
{
  "message": "create changefeed statement is CREATE CHANGEFEED FOR TABLE \"src-15840-196\".\"public\".\"tbl-33\", \"src-15840-196\".\"public\".\"tbl-34\" INTO 'webhook-https://127.0.0.1:33853/tgt-15840-198/public?insecure_tls_skip_verify=true'  WITH updated,     resolved='1s',     webhook_auth_header='***'",
  "severity": "DEBUG",
  "timestamp": {
    "seconds": 1728618370,
    "nanos": 330893135
  }
}
    integration_test.go:718: 
        	Error Trace:	/home/runner/work/replicator/replicator/internal/source/cdc/server/integration_test.go:718
        	            				/home/runner/work/replicator/replicator/internal/source/cdc/server/integration_test.go:618
        	Error:      	Received unexpected error:
        	            	unknown schema: "src-15840-196"."public"
```

Wondering if you have thoughts here?

ryanluu12345:
So I'm doing some testing locally just to better understand what may be wrong here. I've identified that I actually need to use the source accumulators and watchers so that the data ends up in the source (CRDB) instead of the target. However, I've done testing here and verified it matches what I see in this reference ([link](https://github.com/cockroachdb/replicator/blob/bob_core_open/internal/source/pglogical/integration_test.go#L129-L172)) and am now hitting a different error:
```
Error:          Received unexpected error:
                                update sent to unknown tables: src-91611-1.public.tbl-1
                                github.com/cockroachdb/replicator/internal/types.(*orderedAdapter).flush
```

This is with this setup:
```
acc := types.OrderedAcceptorFrom(sourceFixture.ApplyAcceptor, sourceFixture.Watchers)

	for i := range maxIterations {
		batch := &types.MultiBatch{}
		sourceGeneratorWorkload.GenerateInto(batch, hlc.New(int64(i), i+1))

		r.NoError(err)
		r.NoError(acc.AcceptMultiBatch(ctx, batch, &types.AcceptOptions{TargetQuerier: tx}))
		r.NoError(tx.Commit())
	}
```
I've also done a sanity check and verified the tables are created on the source correctly:
```
source conn:  postgresql://root@localhost:26257/defaultdb
CREATE TABLE "src-91216-1"."public"."tbl-1" (parent BIGINT PRIMARY KEY, val BIGINT DEFAULT 0 NOT NULL)
CREATE TABLE "src-91216-1"."public"."tbl-2" (
child BIGINT PRIMARY KEY,
parent BIGINT NOT NULL,
val BIGINT DEFAULT 0 NOT NULL,
CONSTRAINT parent_fk FOREIGN KEY(parent) REFERENCES "src-91216-1"."public"."tbl-1"(parent)
)
```

If I understand correctly, it should accept the batch and be able to write it to the source. The thing that is a bit suspect is the source conn string, which points to defaultdb. I'm wondering if it's an issue because that source connection should be pointing towards the src schema instead. I'm not touching the source or target pool connection strings, so I wonder why it's not using the correct schema on the source.

internal/source/cdc/server/integration_test.go
```diff
@@ -395,3 +401,360 @@ func supportsQueries(version string) (bool, error) {
...
+		sourceVersion)
+	r.NoError(err)
+	log.Debugf("create changefeed statement is %s", createStmt)
+	_, err = sourcePool.ExecContext(ctx, createStmt)
+	r.NoError(err)
+
+	// Make this the target fixture for the accumulator. This is
+	// required for the data to write properly later on when
+	// we accumulate the batch.
+	acc := types.OrderedAcceptorFrom(targetFixture.ApplyAcceptor, targetFixture.Watchers)
```
bobvawter, line 720: This is backwards; you want to build the accumulator using the source fixture.

ryanluu12345: Discussed offline. This resolved the issue!

ryanluu12345: There is a remaining test failure, but I can fix it. I'll just skip tests for versions of cockroach that don't support `ERROR: unimplemented: recursive query "refs" uses UNION which is not implemented (only UNION ALL is supported) (SQLSTATE 0A000)`

bobvawter: What context are you seeing that `UNION ALL` error message in?

ryanluu12345: I'm seeing it in the context of creating the new fixture from base:
```
sourceFixture, err := all.NewFixtureFromBase(targetFixture.Swapped())
r.NoError(err)
```
This is used in `schemawatch/dependencies.go` and the query is stored in `depOrderTemplatePg`. So we use this to get the dependency refs it appears. 

Union is used right [here](https://github.com/cockroachdb/replicator/blob/master/internal/target/schemawatch/dependencies.go#L203).

ryanluu12345: Mentioned offline, but going to try to update the usage of `UNION` with `UNION ALL`. My understanding of the implication is that this may lead to that union having duplicated rows upon completion of joining those two sets of data, vs. UNION which should deduplicate.

ryanluu12345: Ok, tried this and it caused test failures in many of the other test cases. I've reverted it back. But the run from the `UNION` test is here: https://github.com/cockroachdb/replicator/actions/runs/11335422865/job/31523549708?pr=1042

ryanluu12345: Discussed offline here but we are adding a gate for UNION being supported for recursive CTEs in order to run the workload test. We already have coverage for older versions using `testIntegration`.

ryanluu12345:
@bobvawter , addressed all major concerns mentioned above and also added the version gate (22.1) for union in recursive CTEs. This is ready for a final review. I'll clean up the commits and squash them after addressing any comments you have here.

bobvawter (commented):
Also, reflow to 100 columns.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1042#-:-O9ofJ658azKWQrgXZR6:bmugz1i)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @Jeremyyang920 and @ryanluu12345)

ryanluu12345 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1042)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @bobvawter and @Jeremyyang920)

___
*[`internal/source/cdc/server/integration_test.go` line 765 at r4](https://reviewable.io/reviews/cockroachdb/replicator/1042#-O9oeqT-6WWSFSXjBVKM:-O9qFTSH3F_px54dpYbQ:b-896fix) ([raw file](https://github.com/cockroachdb/replicator/blob/98ae9ca355cc13b81fcf36adb27003f429121ef8/internal/source/cdc/server/integration_test.go#L765)):*

This test is brittle.  Remove `7` in favor of checking for the number of expected values per the workload checker.
</blockquote></details>

Done.
___
*[`internal/target/schemawatch/watcher.go` line 221 at r4](https://reviewable.io/reviews/cockroachdb/replicator/1042#-O9oeWpR4qPA6kPy_LjG:-O9qEoNO-eoeeSinkGxJ:b-896fix) ([raw file](https://github.com/cockroachdb/replicator/blob/98ae9ca355cc13b81fcf36adb27003f429121ef8/internal/target/schemawatch/watcher.go#L221)):*

Remove printf.
</blockquote></details>

Done.

ryanluu12345 (commented):
Done!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1042#-:-O9qH0Pv6BpBwZghcm0b:b-896fja)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @bobvawter and @Jeremyyang920)

ryanluu12345:
Hi @bobvawter , just wanted to bump this PR since I've addressed all outstanding comments here. I'll fix the issues on the other PRs you've noted.

bobvawter (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1042)* status: 0 of 4 files reviewed, 15 unresolved discussions (waiting on @Jeremyyang920 and @ryanluu12345)

___
*[`internal/source/cdc/server/integration_test.go` line 389 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3JAmZ6Yin0y5yLKIw:-OA3JAmZ6Yin0y5yLKIx:b-80xfat) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L389)):*
> ```Go
> 
> // Union recursive CTEs are only supported in v22.1 and later.
> func supportsUnionRecursiveCTEs(version string) (bool, error) {
> ```

Sort.
___
*[`internal/source/cdc/server/integration_test.go` line 412 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3JJoM2CvM-j8JakqM:-OA3JJoM2CvM-j8JakqN:b-qppabq) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L412)):*
> ```Go
> func getConfig(
> 	cfg *testConfig, fixture *all.Fixture, targetPool *types.TargetPool,
> ) (*Config, error) {
> ```

Why `error` ?
___
*[`internal/source/cdc/server/integration_test.go` line 458 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3KD_l8Z8cf6PH1zHt:-OA3KD_l8Z8cf6PH1zHu:br4oyrw) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L458)):*
> ```Go
> // the input parameters as specific as possible so that the caller doesn't need
> // to construct whole structs with unrelated information to create the
> // changefeed. This makes it so this is more portable for future work that needs
> ```

Huh?  Any use of a `create changefeed` template function is going to require more-or-less the same data to be passed in and this is a function signature that's only going to get larger.  This change wants to be in a utility package and the `CreateChangefeedStatement` is just that type's `String()` method.
___
*[`internal/source/cdc/server/integration_test.go` line 646 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3LPYk4vC2yCtThjAi:-OA3LPYk4vC2yCtThjAj:bse2itg) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L646)):*
> ```Go
> 
> func TestWorkload(t *testing.T) {
> 	testWorkload(t)
> ```

Silly delegation.
___
*[`internal/source/cdc/server/integration_test.go` line 660 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3NNDQ7OLTx7EJTo1z:-OA3NNDQ7OLTx7EJTo2-:b-3ihtiv) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L660)):*
> ```Go
> 
> 	// Union for recursive CTEs is required in order to use the workload checker
> 	// since there is a UNION under the hood.
> ```

That's not quite correct.  The checker itself is fine, the complication is the schema-inspection code that's part of setting up the fixture.
___
*[`internal/source/cdc/server/integration_test.go` line 668 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3No_c9qd46Ixv8dzq:-OA3No_d3lnuUu3xjGLY:b-hh856f) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L668)):*
> ```Go
> 	}
> 
> 	cfg := &testConfig{webhook: true}
> ```

Should this cycle through other flavors?
___
*[`internal/source/cdc/server/integration_test.go` line 677 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3NzamA73hiZMgDeqU:-OA3NzamA73hiZMgDeqV:ba1tdnn) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L677)):*
> ```Go
> 	supportsQueries, err := supportsQueries(sourceVersion)
> 	r.NoError(err)
> 	if cfg.queries && !supportsQueries {
> ```

This is always false.
___
*[`internal/source/cdc/server/integration_test.go` line 691 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3O9IQBQvfXaNIgfoh:-OA3O9IQBQvfXaNIgfoi:b-xqc69v) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L691)):*
> ```Go
> 	child := ident.NewTable(sourceSchema, targetChecker.Child.Name().Table())
> 	sourceGeneratorWorkload := workload.NewGeneratorBase(parent, child)
> 	r.NoError(err)
> ```

This check is out of place.
___
*[`internal/source/cdc/server/integration_test.go` line 698 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3OShLDLd4rHjd4KIG:-OA3OShM8dzanwZcuJDq:bi61c8r) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L698)):*
> ```Go
> 	sourcePool := targetFixture.SourcePool
> 	parent = sourceGeneratorWorkload.Parent
> 	child = sourceGeneratorWorkload.Child
> ```

Didn't you just set this a couple of lines above?
___
*[`internal/source/cdc/server/integration_test.go` line 723 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3P-Od2uP9Ezd5V0bz:-OA3P-Od2uP9Ezd5V0c-:b-qvsmh3) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L723)):*
> ```Go
> 	// Create the test server fixture.
> 	// Preflight sets default values that are not set in the testConfig.
> 	r.NoError(serverCfg.Preflight())
> ```

Why doesn't it?
___
*[`internal/source/cdc/server/integration_test.go` line 725 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3MI7Q4diWYYxnHa39:-OA3MI7Q4diWYYxnHa3A:bp78tr6) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L725)):*
> ```Go
> 	r.NoError(serverCfg.Preflight())
> 
> 	timeoutCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
> ```

The `Context` in a test fixture already has a timeout behavior.
___
*[`internal/source/cdc/server/integration_test.go` line 749 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3PKnT4mcWPQ35mnjk:-OA3PKnT4mcWPQ35mnjl:b9uu85f) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L749)):*
> ```Go
> 		diff:    cfg.diff,
> 		queries: cfg.queries,
> 		webhook: cfg.webhook,
> ```

Are these fields in `cfg` set?
___
*[`internal/source/cdc/server/integration_test.go` line 763 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3N16X-ekJLnLDIr-V:-OA3N16X-ekJLnLDIr-W:b-cjodqw) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L763)):*
> ```Go
> 	// Make this the target fixture for the accumulator. This is
> 	// required for the data to write properly later on when
> 	// we accumulate the batch.
> ```

I don't understand what this comment is trying to say.
___
*[`internal/source/cdc/server/integration_test.go` line 792 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3PsM22Xug-LIdLH5n:-OA3PsM22Xug-LIdLH5o:b-t11qej) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L792)):*
> ```Go
> 			break
> 		}
> 		log.Debug("waiting for target rows to be written")
> ```

This loop should probably also check for the expected number of child rows, since the generator's last action could be to create a new child.
___
*[`internal/source/cdc/server/integration_test.go` line 799 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3QAb32KvephCg1Cmr:-OA3QAb32KvephCg1Cms:b-nyy020) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L799)):*
> ```Go
> 
> 	// We need to wait for the connection to shut down, otherwise the
> 	// database cleanup callbacks (to drop the publication, etc.) from
> ```

Publication?

ryanluu12345 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1042)* status: 0 of 4 files reviewed, 15 unresolved discussions (waiting on @bobvawter and @Jeremyyang920)

___
*[`internal/source/cdc/server/integration_test.go` line 412 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3JJoM2CvM-j8JakqM:-OA3tm7R3GUFEtUYQYOB:b7bzyof) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L412)):*

Why `error` ?
</blockquote></details>

Good call out, removing.
___
*[`internal/source/cdc/server/integration_test.go` line 660 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3NNDQ7OLTx7EJTo1z:-OA3u3eY9tqj5TMeQwWX:b-896fix) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L660)):*

That's not quite correct.  The checker itself is fine, the complication is the schema-inspection code that's part of setting up the fixture.
</blockquote></details>

Done.
___
*[`internal/source/cdc/server/integration_test.go` line 691 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3O9IQBQvfXaNIgfoh:-OA3uXwPCb6pm7AJSlYI:b-19papw) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L691)):*

This check is out of place.
</blockquote></details>

Good catch, removing.
___
*[`internal/source/cdc/server/integration_test.go` line 698 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3OShLDLd4rHjd4KIG:-OA3uZ33BaaHo5LUDGzS:bf0q64i) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L698)):*

Didn't you just set this a couple of lines above?
</blockquote></details>

Updated this so we don't create this twice. I'm going to just pass in the result of ident.NewTable to the `NewGeneratorBase` call and then use the `sourceGeneratorWorkload` as the source of truth to grab the parent and child table SQLs.
___
*[`internal/source/cdc/server/integration_test.go` line 792 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3PsM22Xug-LIdLH5n:-OA3w1d3EbCobqQ_O461:bk18h8f) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L792)):*

This loop should probably also check for the expected number of child rows, since the generator's last action could be to create a new child.
</blockquote></details>

Checking for both the child and parent rows now!
___
*[`internal/source/cdc/server/integration_test.go` line 799 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3QAb32KvephCg1Cmr:-OA3vEtsDnT_yV8hgYdh:b-7ofgf4) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L799)):*

Publication?
</blockquote></details>

Removed and updated comment to be more clear.

ryanluu12345 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1042)* status: 0 of 4 files reviewed, 15 unresolved discussions (waiting on @bobvawter and @Jeremyyang920)

___
*[`internal/source/cdc/server/integration_test.go` line 763 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3N16X-ekJLnLDIr-V:-OA3wGre5Ia_jXjjbKRl:b-kawcy) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L763)):*

I don't understand what this comment is trying to say.
</blockquote></details>

Yeah comment was a note to self when I didn't really understand how it works. I don't think a comment is really needed here. We just want to create a multiacceptor so that we can accept the batches and write them over to the source database.

ryanluu12345 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1042)* status: 0 of 4 files reviewed, 15 unresolved discussions (waiting on @bobvawter and @Jeremyyang920)

___
*[`internal/source/cdc/server/integration_test.go` line 749 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3PKnT4mcWPQ35mnjk:-OA3waV40QDpc8OJ5XIY:bpx4ofd) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L749)):*

Are these fields in `cfg` set?
</blockquote></details>

They aren't currently in this test, but I'd imagine for other tests, they're useful to set so that they can run with diff mode, in cdc queries mode, or webhook

ryanluu12345 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1042)* status: 0 of 4 files reviewed, 15 unresolved discussions (waiting on @bobvawter and @Jeremyyang920)

___
*[`internal/source/cdc/server/integration_test.go` line 677 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3NzamA73hiZMgDeqU:-OA3xHcJ6AVDtqel5IPy:b-ddt5vf) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L677)):*

This is always false.
</blockquote></details>

Removing the `cfg.queries` since you're right, it's always false

ryanluu12345 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1042)* status: 0 of 4 files reviewed, 15 unresolved discussions (waiting on @bobvawter and @Jeremyyang920)

___
*[`internal/source/cdc/server/integration_test.go` line 646 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3LPYk4vC2yCtThjAi:-OA3xmKR6oDjmkmxcLix:b-h155di) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L646)):*

Silly delegation.
</blockquote></details>

Resolving.
___
*[`internal/source/cdc/server/integration_test.go` line 668 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3No_c9qd46Ixv8dzq:-OA3xR3A3hz_LGXEcd6M:b-puv00r) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L668)):*

Should this cycle through other flavors?
</blockquote></details>

Decided to test the simplest case, webhook, since the setup for the changefeed is simpler (doesn't require multiple changefeed queries for multiple tables). Also, we have coverage for the other flavors already inside of `TestIntegration` above. 
___
*[`internal/source/cdc/server/integration_test.go` line 725 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3MI7Q4diWYYxnHa39:-OA3yV-ODLX-SQ7WL_Hk:blhvknd) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L725)):*

The `Context` in a test fixture already has a timeout behavior.
</blockquote></details>

Good call, going to just use `targetFixture.Context`

ryanluu12345 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1042)* status: 0 of 4 files reviewed, 15 unresolved discussions (waiting on @bobvawter and @Jeremyyang920)

___
*[`internal/source/cdc/server/integration_test.go` line 389 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3JAmZ6Yin0y5yLKIw:-OA3z4h32r0P10tSgkZ4:b-896fix) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L389)):*

Sort.
</blockquote></details>

Done.

ryanluu12345 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1042)* status: 0 of 4 files reviewed, 15 unresolved discussions (waiting on @bobvawter and @Jeremyyang920)

___
*[`internal/source/cdc/server/integration_test.go` line 458 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3KD_l8Z8cf6PH1zHt:-OA3zMziAdpN1oIudfZa:b-n5y930) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L458)):*

Huh?  Any use of a `create changefeed` template function is going to require more-or-less the same data to be passed in and this is a function signature that's only going to get larger.  This change wants to be in a utility package and the `CreateChangefeedStatement` is just that type's `String()` method.
</blockquote></details>

Ok got it, so if I'm understanding this comment correctly:
- Create a new struct type for this `CreateChangefeedStatement`
- Make it so that the `.String()` method returns the result of the changefeed statement
- Move this to its own utility package?

ryanluu12345 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1042)* status: 0 of 4 files reviewed, 15 unresolved discussions (waiting on @bobvawter and @Jeremyyang920)

___
*[`internal/source/cdc/server/integration_test.go` line 723 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3P-Od2uP9Ezd5V0bz:-OA44Xfk1iFCG2-f2o66:bddy3ri) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L723)):*

Why doesn't it?
</blockquote></details>

I adapted this from `testIntegration` that delegates to the `Preflight` to set defaults for configurations they don't set. From my understanding, there are certain configs that the cdc server integration test cares about, and the pre flight is to catch all the rest that are not relevant to configuring this test but are necessary for the fixture to function properly.

ryanluu12345 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1042)* status: 0 of 4 files reviewed, 15 unresolved discussions (waiting on @bobvawter and @Jeremyyang920)

___
*[`internal/source/cdc/server/integration_test.go` line 723 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3P-Od2uP9Ezd5V0bz:-OA44rYM9_hdE81g3BYD:b12v00a) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L723)):*

I adapted this from `testIntegration` that delegates to the `Preflight` to set defaults for configurations they don't set. From my understanding, there are certain configs that the cdc server integration test cares about, and the pre flight is to catch all the rest that are not relevant to configuring this test but are necessary for the fixture to function properly.
</blockquote></details>

If you take a look at the `getConfig` method, you can see that I set the same values that are set in `testIntegration`.

bobvawter (changes requested):
> Reviewed 2 of 3 files at r1, 1 of 2 files at r5, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1042)* status: all files reviewed, 5 unresolved discussions (waiting on @Jeremyyang920 and @ryanluu12345)

___
*[`internal/source/cdc/server/integration_test.go` line 458 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3KD_l8Z8cf6PH1zHt:-OA4-xyH7BPUJjdPtID0:bmwtcq4) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L458)):*

Ok got it, so if I'm understanding this comment correctly:
- Create a new struct type for this `CreateChangefeedStatement`
- Make it so that the `.String()` method returns the result of the changefeed statement
- Move this to its own utility package?
</blockquote></details>

That's correct.
___
*[`internal/source/cdc/server/integration_test.go` line 668 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3No_c9qd46Ixv8dzq:-OA48hER5d9FXkEBIzeP:b-metxuj) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L668)):*

Decided to test the simplest case, webhook, since the setup for the changefeed is simpler (doesn't require multiple changefeed queries for multiple tables). Also, we have coverage for the other flavors already inside of `TestIntegration` above.
</blockquote></details>

If you were to try to use your changefeed creator to instantiate the changefeed query case, you would find that the API is incomplete.  There's no way to specify the projection, so it couldn't be used, for example, with the child table.  There's no real difference in complexity between creating a single changefeed or two.  You'd set `WorkloadConfig.DisableFK` and the test should run the same way.

Also, `diff` exercises some slightly different code paths.  While that shouldn't make a difference, making this test comprehensive might shake loose a corner case today or in the future.
___
*[`internal/source/cdc/server/integration_test.go` line 677 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3NzamA73hiZMgDeqU:-OA41kIs3P2PLsYeQ1gj:b29xwf3) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L677)):*

Removing the `cfg.queries` since you're right, it's always false
</blockquote></details>

Does this test need to be guarded by query support if it's not being tested (even though it should be)?
___
*[`internal/source/cdc/server/integration_test.go` line 723 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3P-Od2uP9Ezd5V0bz:-OA46jRfEMX_AnYi7esW:b4hkx6i) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L723)):*

If you take a look at the `getConfig` method, you can see that I set the same values that are set in `testIntegration`.
</blockquote></details>

Let. me rephrase, is there any reason that `getConfig()` shouldn't call `Preflight()` so that its return value is ready-to-use?

ryanluu12345 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1042)* status: 3 of 4 files reviewed, 5 unresolved discussions (waiting on @bobvawter and @Jeremyyang920)

___
*[`internal/source/cdc/server/integration_test.go` line 677 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3NzamA73hiZMgDeqU:-OA4NM2e2ano4tMC6_iq:bwgdio3) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L677)):*

Does this test need to be guarded by query support if it's not being tested (even though it should be)?
</blockquote></details>

Yes, ti should be guarded. Especially because we've established above that we want to test different flavors of this. We can test the case where `queries` are used.
___
*[`internal/source/cdc/server/integration_test.go` line 723 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3P-Od2uP9Ezd5V0bz:-OA4NVBP3RjvfR6wjMo0:b-xyf0y0) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L723)):*

Let. me rephrase, is there any reason that `getConfig()` shouldn't call `Preflight()` so that its return value is ready-to-use?
</blockquote></details>

Good point, let me do this.

ryanluu12345 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1042)* status: 3 of 4 files reviewed, 5 unresolved discussions (waiting on @bobvawter and @Jeremyyang920)

___
*[`internal/source/cdc/server/integration_test.go` line 668 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3No_c9qd46Ixv8dzq:-OA4O3HVBq9GUll6e-Lv:b3t98ip) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L668)):*

If you were to try to use your changefeed creator to instantiate the changefeed query case, you would find that the API is incomplete.  There's no way to specify the projection, so it couldn't be used, for example, with the child table.  There's no real difference in complexity between creating a single changefeed or two.  You'd set `WorkloadConfig.DisableFK` and the test should run the same way.

Also, `diff` exercises some slightly different code paths.  While that shouldn't make a difference, making this test comprehensive might shake loose a corner case today or in the future.
</blockquote></details>

Ok got it, so we can test these flavors by setting `DisableFK`? However, does this mean child data won't make it's way to the target? or does it mean that parent and child data will make it's way to the target but just without any FKs?

Alright, I think it's worthwhile here, let's do that then.

ryanluu12345 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1042)* status: 1 of 6 files reviewed, 5 unresolved discussions (waiting on @bobvawter and @Jeremyyang920)

___
*[`internal/source/cdc/server/integration_test.go` line 458 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OA3KD_l8Z8cf6PH1zHt:-OAK4aPV2glzMyXOidjc:b-djridv) ([raw file](https://github.com/cockroachdb/replicator/blob/40b20e12cb23ed6014d387bf7d8f88ebd1256c12/internal/source/cdc/server/integration_test.go#L458)):*

That's correct.
</blockquote></details>

Yep, as discussed offline I've just moved this under `internal/sinktest` and moved over all relevant helpers. The code should be cleaner now.

I also added support for `key_in_value` now that that is merged in.

ryanluu12345 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1042)* status: 3 of 6 files reviewed, 15 unresolved discussions (waiting on @bobvawter and @Jeremyyang920)

___
*[`internal/sinktest/changefeed.go` line 1 at r10](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OANsF1c5hofefIRY38X:-OAO05ejDHZ1J8fk8A-K:b-896fix) ([raw file](https://github.com/cockroachdb/replicator/blob/40974710563750bc1c1579454cec7e72a93fa620/internal/sinktest/changefeed.go#L1)):*

Copyright.
</blockquote></details>

Done.
___
*[`internal/sinktest/changefeed.go` line 14 at r10](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OANsKHOCoxt3EE5JfZm:-OAO06RVEBMXmkOSpxkM:b-896fix) ([raw file](https://github.com/cockroachdb/replicator/blob/40974710563750bc1c1579454cec7e72a93fa620/internal/sinktest/changefeed.go#L14)):*

Reflow comments to 72 chars, here and elsewhere.
</blockquote></details>

Done.
___
*[`internal/sinktest/changefeed.go` line 16 at r10](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OANsNMrBdvBOc8f68pC:-OAO0LHE1bFO4Jsxa0mA:b-896fix) ([raw file](https://github.com/cockroachdb/replicator/blob/40974710563750bc1c1579454cec7e72a93fa620/internal/sinktest/changefeed.go#L16)):*

Sort.
</blockquote></details>

Done.
___
*[`internal/sinktest/changefeed.go` line 18 at r10](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OANsOYM8uFSW9JjzCrT:-OAO0LaJ7rcUE7Eyg-sX:b-896fix) ([raw file](https://github.com/cockroachdb/replicator/blob/40974710563750bc1c1579454cec7e72a93fa620/internal/sinktest/changefeed.go#L18)):*

Sort.
</blockquote></details>

Done.
___
*[`internal/sinktest/changefeed.go` line 46 at r10](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OANsUfC3SICied5xp1b:-OAO0TPVB45o5YceOphB:bn6bb9u) ([raw file](https://github.com/cockroachdb/replicator/blob/40974710563750bc1c1579454cec7e72a93fa620/internal/sinktest/changefeed.go#L46)):*

Uncapitalize local variables.
</blockquote></details>

Sorry, this was a good catch. I did a find and replace in order to update the `table` field to `Table`. I think this was an unintended effect.
___
*[`internal/sinktest/changefeed_test.go` line 1 at r10](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OANteLK0k0ZllrCvg9V:-OAO0cpc1BKhq3KCUJK1:b-896fix) ([raw file](https://github.com/cockroachdb/replicator/blob/40974710563750bc1c1579454cec7e72a93fa620/internal/sinktest/changefeed_test.go#L1)):*

Copyright.
</blockquote></details>

Done.
___
*[`internal/source/cdc/server/integration_test.go` line 425 at r10](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OANuTRTCK9djumfsmGL:-OAO0ux3-u2MUSPmfGr5:b-4hvi0z) ([raw file](https://github.com/cockroachdb/replicator/blob/40974710563750bc1c1579454cec7e72a93fa620/internal/source/cdc/server/integration_test.go#L425)):*

Return `(*Config, error)`.  Passing in the `*require.Assertions` mixes concerns of producing a valid configuration, versus test flow.
</blockquote></details>

Got it!
___
*[`internal/source/cdc/server/integration_test.go` line 459 at r10](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OANubV5-_5sVtoaK_7M:-OAO0wtp9kKx3qcKQW8M:b5j0n71) ([raw file](https://github.com/cockroachdb/replicator/blob/40974710563750bc1c1579454cec7e72a93fa620/internal/source/cdc/server/integration_test.go#L459)):*

You can `return fixtureCfg, fixtureCfg.Preflight()`
</blockquote></details>

Yep, cleaner! I'll do it this way
___
*[`internal/source/cdc/server/integration_test.go` line 462 at r10](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OANugaK4fBzNhiPp120:-OAO118lDL9tH-lEa5LB:b3pn2ic) ([raw file](https://github.com/cockroachdb/replicator/blob/40974710563750bc1c1579454cec7e72a93fa620/internal/source/cdc/server/integration_test.go#L462)):*

If this is specific only to the `testWork()`, declare it at the top of the function.
</blockquote></details>

Yep only used here, I'll update this to be at the top of the function.
___
*[`internal/source/cdc/server/integration_test.go` line 623 at r10](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OANxEdYCDFrdw0TF8Ff:-OAO1paR6k4hBulyVavj:b-896fix) ([raw file](https://github.com/cockroachdb/replicator/blob/40974710563750bc1c1579454cec7e72a93fa620/internal/source/cdc/server/integration_test.go#L623)):*

Unnecessary assignment of zero value.
</blockquote></details>

Done.

bobvawter (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1042)* status: 1 of 6 files reviewed, 12 unresolved discussions (waiting on @Jeremyyang920 and @ryanluu12345)

___
*[`internal/sinktest/changefeed.go` line 1 at r10](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OANsF1c5hofefIRY38X:-OAO35Cb-kwWW4uZVYAY:bn3meyy) ([raw file](https://github.com/cockroachdb/replicator/blob/40974710563750bc1c1579454cec7e72a93fa620/internal/sinktest/changefeed.go#L1)):*

Done.
</blockquote></details>

It's 2024.

ryanluu12345 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1042)* status: 1 of 6 files reviewed, 12 unresolved discussions (waiting on @bobvawter and @Jeremyyang920)

___
*[`internal/sinktest/changefeed.go` line 1 at r10](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OANsF1c5hofefIRY38X:-OAO4VRjA5AgeDCFZiEr:b-896fix) ([raw file](https://github.com/cockroachdb/replicator/blob/40974710563750bc1c1579454cec7e72a93fa620/internal/sinktest/changefeed.go#L1)):*

It's 2024.
</blockquote></details>

Done.
___
*[`internal/sinktest/changefeed.go` line 26 at r10](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OANssUYDRWcW3V6Eq2k:-OAO2Bqr-0CekubPangs:b-qb6kcd) ([raw file](https://github.com/cockroachdb/replicator/blob/40974710563750bc1c1579454cec7e72a93fa620/internal/sinktest/changefeed.go#L26)):*

This seems like an unnecessary separation of concerns.  Why would some fields go into `ChangefeedConfig`, but not into `ChangefeedStatement` ?  Does `ChangefeedConfig` need to exist at all?
</blockquote></details>

Good point, I'll go ahead and remove this struct wrapper. Not needed here anymore.
___
*[`internal/sinktest/changefeed.go` line 32 at r10](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OANu9IaBvCxEplRDiSO:-OAO6pOk3wTVUc_Gnt2J:bch1o5j) ([raw file](https://github.com/cockroachdb/replicator/blob/40974710563750bc1c1579454cec7e72a93fa620/internal/sinktest/changefeed.go#L32)):*

Is is only columns that can be in this field?  If so, a `[]ident.Ident` seems like the right choice.  If it's an arbitrary projection, just call it `QueryProjection`.
</blockquote></details>

For now I'm going to support just columns. If we need more flexibility later, I'll make this an arbitrary projection. I think `[]ident.Ident` is a good move here and will support this + update the code and tests to reflect this.
___
*[`internal/source/cdc/server/integration_test.go` line 469 at r10](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OANurdl5DCXYbZgAlye:-OAP1GU43OQuiiKiLV0k:b-896fix) ([raw file](https://github.com/cockroachdb/replicator/blob/40974710563750bc1c1579454cec7e72a93fa620/internal/source/cdc/server/integration_test.go#L469)):*

Reflow to 100 columns.
</blockquote></details>

Done.
___
*[`internal/source/cdc/server/integration_test.go` line 471 at r10](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OANvB7c96nxCMDuxzAj:-OAP1H4W3LkouRvEl4sC:b-uy0gg5) ([raw file](https://github.com/cockroachdb/replicator/blob/40974710563750bc1c1579454cec7e72a93fa620/internal/source/cdc/server/integration_test.go#L471)):*

This should also test non-webhook delivery.  Webhook versus ndjson  _shouldn't_ make any difference, but it's good to verify across the combinatorial explosion that is changefeed configurations.
</blockquote></details>

Yep good call out here. I'm also planning to test against:
- webhook == false, queries == false, diff = false | diff = true
- webhook == false, queries == diff == key_in_value == true (since these must be set together in order to emit all deletes)
___
*[`internal/source/cdc/server/integration_test.go` line 514 at r10](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OANvjIF4TFRzjdD_iLk:-OAP1XHp9ElX58ZzrkOk:b-n30iuh) ([raw file](https://github.com/cockroachdb/replicator/blob/40974710563750bc1c1579454cec7e72a93fa620/internal/source/cdc/server/integration_test.go#L514)):*

Shouldn't this be conditional on changefeed queries being enabled?
</blockquote></details>

Yep, it should, I made it key off of `cfg.queries`

ryanluu12345 (commented):
Ok, so I've gone through the log lines for the test cases that failed and I've noticed that this the common thread between all the failed runs is the test case: `TestWorkload/webhook\_true\_diff\_true\_queries\_true\_key\_in\_value\_true` and the following logging:
```
"error": "proposed checkpoint timestamp for group=\"tgt-15510-396.public\" [ \"tgt-15510-396\".\"public\".\"tbl-69\", \"tgt-15510-396\".\"public\".\"tbl-70\" ], partition=\"tgt-15510-396.public\" is going backwards: 1730245103746546739.0000000000; verify changefeed cursor or remove already-applied checkpoint timestamp entries",
```
We see several log messages of this form for every failed run. However, for all runs that succeeded, we did not see any failures of this types (grepped for `proposed checkpoint timestamp`). 

I have seen these failures on the following targets:
-   MySQL 5.6 as target
-   MySQL Maria DB v10
-   Postgres v12
-   Oracle v21.3
-   Objstore
-   Postgres v13

I know that we previously implemented the configuration option to be able to disable this check, but I don't want to paper over this if this is a legitimate issue. I'm wondering if you have any insights into this noted behavior?

I'll attach some raw logs of a few failure cases below:
https://productionresultssa17.blob.core.windows.net/actions-results/265f65ce-26a7-4892-848d-c634a7627b7c/workflow-job-run-9651b4f4-7e76-59bc-40cd-903b1abdbf95/logs/job/job-logs.txt?rsct=text%2Fplain&se=2024-10-30T00%3A21%3A38Z&sig=qG8o1kliQWmvUQ77cYpkZE%2FgReI9w2Mtarz4xrlajDY%3D&ske=2024-10-30T08%3A46%3A07Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2024-10-29T20%3A46%3A07Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2024-08-04&sp=r&spr=https&sr=b&st=2024-10-30T00%3A11%3A33Z&sv=2024-08-04

https://productionresultssa17.blob.core.windows.net/actions-results/265f65ce-26a7-4892-848d-c634a7627b7c/workflow-job-run-8e0db1fe-7ff6-5c40-1ba5-2022544b595a/logs/job/job-logs.txt?rsct=text%2Fplain&se=2024-10-30T00%3A21%3A27Z&sig=kA1E1j%2BqwIXuvbX9mXcZ6ocgzs1L5ZsECKRgOP2sliM%3D&ske=2024-10-30T08%3A54%3A12Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2024-10-29T20%3A54%3A12Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2024-08-04&sp=r&spr=https&sr=b&st=2024-10-30T00%3A11%3A22Z&sv=2024-08-04

https://productionresultssa17.blob.core.windows.net/actions-results/265f65ce-26a7-4892-848d-c634a7627b7c/workflow-job-run-116d4943-12ce-5a1c-6e9a-64c4c5013f12/logs/job/job-logs.txt?rsct=text%2Fplain&se=2024-10-30T00%3A21%3A00Z&sig=UieORS09NbOHCOkX3SDaWqBxwHv5xrYIksKJVUq0UFE%3D&ske=2024-10-30T08%3A46%3A38Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2024-10-29T20%3A46%3A38Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2024-08-04&sp=r&spr=https&sr=b&st=2024-10-30T00%3A10%3A55Z&sv=2024-08-04

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1042#-:-OAPl5I348NUE4jkDkkg:bmew627)* status: 1 of 6 files reviewed, 12 unresolved discussions (waiting on @bobvawter and @Jeremyyang920)

ryanluu12345 (commented):
I did an experiment to set the `SkipBackwardsDataCheck` to true and now all CI is passing. Thinking about it more offline, I'm wondering if the cases I ran into above are consistent with the reason why a customer wanted the `SkipBackwardsDataCheck` functionality in the first place. Happy to discuss more here just to get more context and ensure I'm not missing anything here.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1042#-:-OAQ-y-r9RJTWl6_aSFc:b14377m)* status: 1 of 6 files reviewed, 12 unresolved discussions (waiting on @bobvawter and @Jeremyyang920)

bobvawter:
That message indicates a legitimate problem where, for some target schema, the resolved timestamps received from the changefeed aren't proceeding in a monotonic ordering.  I filed https://github.com/cockroachdb/cockroach/issues/112880 a year ago to make this less of an issue for Replicator.

ryanluu12345 (commented):
Thanks for linking the relevant ticket. Makes sense to me, and I'll keep this in mind. Given this context, enabling the `SkipBackwardsDataCheck` will ensure that tests pass in this scenario where resolved timestamps aren't arriving in monotonic ordering. However, it's something we must do until the ticket you linked is resolved and we use the upstream message data within replicator to mitigate or fix this issue.

What are your thoughts on keeping this test as is right now with the `SkipBackwardsDataCheck` enabled since this exercises and validates the e2e CDC server flow? Then, whenever the changefeedccl issue gets resolved, we can make the follow up change and reset `SkipBackwardsDataCheck` to false.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1042#-:-OAT2x9e73YcjDoWG1GU:bfblglx)* status: 1 of 6 files reviewed, 12 unresolved discussions (waiting on @bobvawter and @Jeremyyang920)

ryanluu12345 (commented):
Rereading the ticket, it seems that even with the changes requested in that ticket, we'll still potentially hit this issue but just be able to fail with more informative detail. So in this case I think it would still require the `SkipBackwardsDataCheck`. Later, can add testing that verifies we get the correct informative error logging.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1042#-:-OAT59a0BJrUQtSov3fm:b8kfg2v)* status: 1 of 6 files reviewed, 12 unresolved discussions (waiting on @bobvawter and @Jeremyyang920)

bobvawter:
The check is telling you there's a problem with your test.

ryanluu12345 (commented):
Discussed this offline. Because in the multiple changefeed to same target schema case we run into the data moving backwards issue, since resolved timestamps are not currently differentiated right now, we need to disable the backwards data check. Added in a conditional setting of this `SkipBackwardsDataCheck` to true.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1042#-:-OA_CfYZAi1nooVy-8F3:b26ir68)* status: 1 of 9 files reviewed, 12 unresolved discussions (waiting on @bobvawter and @Jeremyyang920)

bobvawter (approved):
:lgtm: w/ nits

> Reviewed 2 of 5 files at r10, 1 of 3 files at r12, 5 of 5 files at r17, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1042#-:-OArZBn1ANEvqrIJ-KVk:b-rjhpvx)* status: all files reviewed, 4 unresolved discussions (waiting on @Jeremyyang920 and @ryanluu12345)

___
*[`internal/sinktest/changefeed.go` line 35 at r17](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OArXsM25jOwqMCIhH6i:-OArXsM25jOwqMCIhH6j:b-q05scb) ([raw file](https://github.com/cockroachdb/replicator/blob/75d010690607161f75c24c1eb6191bfbe3b41ed0/internal/sinktest/changefeed.go#L35)):*
> ```Go
> 	Host                   string
> 	KeyInValue             bool
> 	Queries                bool
> ```

Isn't this redundant with a non-empty projection?
___
*[`internal/sinktest/changefeed_test.go` line 27 at r17](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OArW0pa2mRuNBt-vml_:-OArW0pa2mRuNBt-vmla:b-dll3d8) ([raw file](https://github.com/cockroachdb/replicator/blob/75d010690607161f75c24c1eb6191bfbe3b41ed0/internal/sinktest/changefeed_test.go#L27)):*
> ```Go
> 
> // Note that the `sourceVersion` must contain a space at the end to satisfy the
> // semver regex (i.e. "CockroachDB CCL v24.2.1 ").
> ```

Your test string is missing the `(platform)` portion of a CRDB `version()` output.

https://github.com/cockroachdb/replicator/blob/master/internal/util/stdpool/crdb_hints_test.go#L48
___
*[`internal/source/cdc/server/integration_test.go` line 432 at r17](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OArXB6s11SowR2YDUak:-OArXB6s11SowR2YDUal:bhlaa8o) ([raw file](https://github.com/cockroachdb/replicator/blob/75d010690607161f75c24c1eb6191bfbe3b41ed0/internal/source/cdc/server/integration_test.go#L432)):*
> ```Go
> 				// different streams. Therefore, the backwards data check
> 				// should be disabled in the multiple changefeed case since we
> 				// know the backwards data check will fail without disabling it.
> ```

Add links to https://github.com/cockroachdb/cockroach/issues/112880 and https://github.com/cockroachdb/replicator/issues/574  to close the loop later on.
___
*[`internal/source/cdc/server/integration_test.go` line 632 at r17](https://reviewable.io/reviews/cockroachdb/replicator/1042#-OArYdNh5GvTQfSI42uS:-OArYdNh5GvTQfSI42uT:b-bq2ujv) ([raw file](https://github.com/cockroachdb/replicator/blob/75d010690607161f75c24c1eb6191bfbe3b41ed0/internal/source/cdc/server/integration_test.go#L632)):*
> ```Go
> 			Target:        targetParent,
> 			Token:         token,
> 			Tables:        tables,
> ```

Sort. Here and above.

Issue #1044, opened 2024-10-14
Prefer changefeed queries to have key_in_value option set


bobvawter:
As presently implemented, we perform a bit of work when parsing changefeed query payloads to create a replication key based on columns in the target table (https://github.com/cockroachdb/replicator/blob/master/internal/source/cdc/primary_key.go#L26). Due to https://github.com/cockroachdb/cockroach/issues/131856 we may not have the information necessary to do so when processing deletes, when the target primary key requires a projected value.

If changefeed queries can support the `key_in_value` option, this back-parsing would be unnecessary. The source changefeed would simply provide a replica identity (the PK of the row that triggered the message). Given the changes to support `REPLICA IDENTITY FULL`, the replica key does not require any relationship to the target PK.

ryanluu12345:
Hi Bob, just want to make sure I understand the ask here clearly. I did a quick search on the changefeed queries docs and found this:
<img width="1214" alt="image" src="https://github.com/user-attachments/assets/f14be0d7-3f6d-4030-bdbb-c8156ba04b93">

The snippet above seems to imply that if we are using a webhook sink, this is already passed by default. Is the ask here then to unmarshal the `primary key array` into an appropriate format (maybe `*ident.Map[int]`) and use that in lieu of the current `getPrimaryKey` ? From an initial look at the codebase it looks like I need to update the `query_payload` processors to pull this data out of the data coming from the source CDC message when `key_in_value` is specified.

ryanluu12345:
Made some progress here, I'm at least able to dump out the received JSON coming out of the request body from the webhook. I see that the data now shows the key in the message:
```
Received JSON: map[length:1 payload:[map[after:map[id:1.204560795e+09 t:yo] before:<nil> key:[1.204560795e+09] updated:1728948380857547000.0000000000]]]
```

**Problem**
One thing to note here is that the `key` seems to be sent with the value of the key, not the name of the column. Given that `getPrimaryKeys` adds to a map that has a key of ident type and value of integer, it doesn't seem like the webhook payload gives us what we need. We'll still need to introspect the table to get the column names.

**Assuming we get the correct data from the webhook (can check to see if I can set another setting to column names)**
Given that we modify the values of `req.keys` via reference, I'm thinking we can embed the logic here to key off of the "key" field and get that data into the `ident.Map`. Alternatively, we can do this after `getPrimaryKey` is called in the webhook handler.

bobvawter:
[...]

PR #1045, opened 2024-10-17
cdcjson/query_payload: updated the unmarshal JSON logic so that it uses the key values from the CDC message if it's provided; otherwise, it does the back-processing to obtain the key values

ryanluu12345:
Previously, query payloads without key_in_value may not have the information
necessary to do so when processing deletes, when the target PK requires a
projected value. This change supports messages that come in with the
key_in_value data. This obviates back-parsing if it's explicitly provided.
Otherwise, we fallback and use the back-parsing in the case where this data is
not provided.
    
This change should help unblock customers we are currently blocked by
performing deletes in certain circumstances.
    
Resolves: #1044
Release Note: None

**TODO**
- [x] Remove duplicated unmarshalling we do for the getPrimaryKey check
- [x] Add final testing to capture changes in the changefeed behavior
- [ ] Update wiki to note the support for `key_in_value` and recommend the best practice of using `key_in_value` 
- [x] Verify with CDC team about the shape of messages for `key_in_value` ([link](https://cockroachlabs.slack.com/archives/C9TGBJB44/p1729270418624779?thread_ts=1729007453.000399&cid=C9TGBJB44)) -> confirmed that they always come as the datums (i.e. the array of PK values in JSON ordered in the same way a composite PK is defined)

**Testing**
- Verified that when PK vals are passed in `key_in_value`, we skip the back-parsing logic to get the values from PK projections (since this may not be provided by the changefeed message on deletes)
- Verified that existing tests still pass with the default logic to not use `key_in_value`
- Verified that the minimal test using webhook queries and key in value succeeded for `integration_test.go`
- In failure case where we don't have data in the `key` key and we also don't parse (I forced this error), the message fails to process -> expected here
```
CREATE CHANGEFEED INTO 'webhook-https://127.0.0.1:53188/tgt-20738-6/public/tbl-1?insecure_tls_skip_verify=true'  WITH updated,     resolved='1s',     webhook_auth_header='Bearer <token>',diff,min_checkpoint_frequency='1s',envelope='wrapped',format='json' AS SELECT pk, val FROM %s
{
  "message": "\"tgt-20738-6.public\" [ \"tgt-20738-6\".\"public\".\"tbl-1\" ]: initial mode is ModeConsistent",
  "severity": "INFO",
  "timestamp": {
    "seconds": 1729188410,
    "nanos": 104524000
  }
}
=== RUN   TestIntegration/diff_transactional_queries_serial_webhook/check_diag_endpoint
{
  "error": "context canceled",
  "message": "did not shut down cleanly",
  "severity": "ERROR",
  "timestamp": {
    "seconds": 1729188415,
    "nanos": 481720000
  }
}
--- PASS: TestIntegration (0.00s)
    --- PASS: TestIntegration/diff_transactional_queries_serial_webhook (5.92s)
        --- PASS: TestIntegration/diff_transactional_queries_serial_webhook/check_diag_endpoint (0.00s)
PASS
ok      github.com/cockroachdb/replicator/internal/source/cdc/server    6.341s
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1045)


internal/util/cdcjson/query_payload.go
```diff
@@ -126,14 +138,20 @@ func (q *queryPayload) UnmarshalJSON(data []byte) error {
...
-	q.keyValues = make([]json.RawMessage, q.keys.Len())
-	for k, pos := range q.keys.All() {
-		v, ok := msg.Get(k)
-		if !ok {
-			return errors.Errorf("missing primary key: %s", k)
+
+	// Extract PK values only if they have not already been
+	// extracted previously due to the key that "key_in_value"
+	// options provide.
+	if len(q.keyValues) == 0 {
```
ryanluu12345, line 145: Open question of if this should only apply to the initial data message we stage or if we should also do this logic for the `resolved` message. Need to clarify this behavior with @bobvawter

internal/util/cdcjson/query_payload.go
```diff
@@ -98,6 +99,17 @@ func (q *queryPayload) UnmarshalJSON(data []byte) error {
 	if !ok {
 		return errors.Errorf("could not find timestamp in field %s while attempting to parse envelope=wrapped", updated)
 	}
+
+	// Get the PK values from the "key" part of the payload.
+	// This is only done for CDC queries that specify the key_in_value option.
+	// In the case that the key_in_value option is not specified, this will
+	// be empty.
+	if key, ok := msg.Get(keyLabel); ok {
```
ryanluu12345, line 107: Important note to make here is that we still do the `getPrimaryKey` data from the schema, but that's because we potentially need this in the case the webhook isn't sending the `keyInValue` metadata. Wondering if we want to key off this data even earlier so we can conditionally pull primary keys. This makes the solution less clean but has the benefit of not doing introspection queries on the schema to get primary key columns.

BramGruneir (commented):
Once this behaviour is complete, we should update the wiki to specify the need for the key_in_value.  

And as a separate issue, perhaps we should deprecate the use of a changefeed query without the key_in_value specified.  

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1045#-:-O9VKN3L7HXD8nJPWzBz:bjr4isu)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @bobvawter and @Jeremyyang920)

bobvawter (commented):
The manual test is nice, but is not sufficient.  Please add a complete test to detect potential future changes in changefeed behavior.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1045#-:-O9VKAwcD5PDAkOWnFZD:btszel1)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @Jeremyyang920 and @ryanluu12345)

___
*[`internal/util/cdcjson/query_payload.go` line 107 at r2](https://reviewable.io/reviews/cockroachdb/replicator/1045#gh-1805653738:-O9VK-RaEA98Yo2U7SXq:b-9q6o0o) ([raw file](https://github.com/cockroachdb/replicator/blob/bbeb0b6025b432cd154f194608dcea19f659634a/internal/util/cdcjson/query_payload.go#L107)):*

Important note to make here is that we still do the `getPrimaryKey` data from the schema, but that's because we potentially need this in the case the webhook isn't sending the `keyInValue` metadata. Wondering if we want to key off this data even earlier so we can conditionally pull primary keys. This makes the solution less clean but has the benefit of not doing introspection queries on the schema to get primary key columns.
</blockquote></details>

It seems like if there is a key in the payload much of the rest of this method can be skipped, since unnecessary work is being performed for every incoming message; it's unmarshaling a bunch of stuff to possibly compute the key later, even though it's not going to be used, and then re-marshaling the data into a fully-formed mutation up above in `AsMutation()`.
___
*[`internal/util/cdcjson/query_payload.go` line 145 at r2](https://reviewable.io/reviews/cockroachdb/replicator/1045#gh-1805651215:-O9V5hoE5SUJG0cv0Npu:b-mdiw18) ([raw file](https://github.com/cockroachdb/replicator/blob/bbeb0b6025b432cd154f194608dcea19f659634a/internal/util/cdcjson/query_payload.go#L145)):*

Open question of if this should only apply to the initial data message we stage or if we should also do this logic for the `resolved` message. Need to clarify this behavior with @bobvawter 
</blockquote></details>

A resolved timestamp message wouldn't have any data associated with it,  beyond the timestamp.

ryanluu12345 (commented):
Sorry, I should have clarified in the PR that I did this manual testing as a sanity check, but I've also added a new test permutation for the `cdc/server/integration_test.go` to support `key_in_value` as an option.

This test should test both paths e2e here. Wondering if you think more coverage is needed. Happy to add as needed. I think this test should also cover the updates Bob and I are talking about below where we should remove some of the duplicate processing as needed.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1045#-:-O9VWnAa7h2qjEHruMqF:bgafuxr)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @Jeremyyang920)

ryanluu12345 (commented):
@BramGruneir noted, I'll make sure there is a list of TODOs here for this PR so we flesh this out in our Wiki + tests

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1045#-:-O9VX8cI5Uy9c4-fupHp:bvigk0z)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @Jeremyyang920)

internal/source/cdc/server/integration_test.go
```diff
@@ -60,16 +60,18 @@ const (
...
-	immediate bool
-	parallel  bool
-	queries   bool
-	webhook   bool
+	diff       bool
+	immediate  bool
+	parallel   bool
+	queries    bool
+	webhook    bool
+	keyInValue bool
```
ryanluu12345, line 74: @bobvawter @BramGruneir automated test update I've made so far. My logic here is that it will test the case where the changefeed doesn't have key_in_value and when it does, for a variety of configurations. This should test that for both cases, the message will be written to the source and correctly reach the target. This can only happen if the primary key values are properly fetched from the payload.

ryanluu12345 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1045)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @bobvawter and @Jeremyyang920)

___
*[`internal/util/cdcjson/query_payload.go` line 107 at r2](https://reviewable.io/reviews/cockroachdb/replicator/1045#gh-1805653738:-O9kRolY8kB-gFMTRWGN:b-h3dztk) ([raw file](https://github.com/cockroachdb/replicator/blob/bbeb0b6025b432cd154f194608dcea19f659634a/internal/util/cdcjson/query_payload.go#L107)):*

It seems like if there is a key in the payload much of the rest of this method can be skipped, since unnecessary work is being performed for every incoming message; it's unmarshaling a bunch of stuff to possibly compute the key later, even though it's not going to be used, and then re-marshaling the data into a fully-formed mutation up above in `AsMutation()`.
</blockquote></details>

I ended updating the logic so that it only calls `getPrimaryKey` if the `key` parameter is not specified in the webhook payload. 

We still need a lot of this method since we populate the resolved and updated fields in the message. But we at least can do away with one set of processing for the `primary key` portion.

ryanluu12345 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1045)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @bobvawter and @Jeremyyang920)

___
*[`internal/util/cdcjson/query_payload.go` line 145 at r2](https://reviewable.io/reviews/cockroachdb/replicator/1045#gh-1805651215:-O9kSFsJ2cuYr3eTfAEx:b8og5wg) ([raw file](https://github.com/cockroachdb/replicator/blob/bbeb0b6025b432cd154f194608dcea19f659634a/internal/util/cdcjson/query_payload.go#L145)):*

A resolved timestamp message wouldn't have any data associated with it,  beyond the timestamp.
</blockquote></details>

Ok thanks for clarifying here!

ryanluu12345 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1045)* status: 0 of 4 files reviewed, 4 unresolved discussions (waiting on @bobvawter and @Jeremyyang920)

___
*[`internal/util/cdcjson/query_payload_test.go` line 124 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1045#-O9l1sNiCx-xqGy98id_:-O9l1sNiCx-xqGy98ida:b-xdtq2d) ([raw file](https://github.com/cockroachdb/replicator/blob/1d53646f4a263497352e9e723845c4d119246a36/internal/util/cdcjson/query_payload_test.go#L124)):*
> ```Go
> 			"could not find timestamp in field \"updated\" while attempting to parse envelope=wrapped",
> 		},
> 		{"primary key values are present in the key field",
> ```

After reviewing the existing testing here, I think this is the simplest test that will test the conditional logic inside of UnmarshalJSON. The logic is spelled out in the comment, but as an additional point of context: since I make it so that `tt.keys` is an empty map, there is no way that the data from the projections can be mapped to key values. This means that this test passing entirely depends on the `key` field being passed from the changefeed message, thus giving us the coverage we need as simply as possible.

In order to address changes in changefeed behavior, we already have the integration_test that includes the new `key_in_value` option and tests all permutations of it. I think this should cover this behavior just fine.

ryanluu12345:
@bobvawter @BramGruneir added the final testing and clean ups that we discussed offline. I think this is ready for a final look. Once I merge this in I'll update the wiki then close out this ticket!

bobvawter (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1045)* status: 0 of 4 files reviewed, 11 unresolved discussions (waiting on @BramGruneir, @Jeremyyang920, and @ryanluu12345)

___
*[`internal/source/cdc/webhook_query.go` line 62 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1045#-O9ohHfmEaYSmpkAS2Oi:-O9ohHfnB1kiBS-ObXAC:bexs0wa) ([raw file](https://github.com/cockroachdb/replicator/blob/1d53646f4a263497352e9e723845c4d119246a36/internal/source/cdc/webhook_query.go#L62)):*
> ```Go
> 	}
> 
> 	keys := &ident.Map[int]{}
> ```

Can this be `nil` or a sentinel value to indicate a `key_in_value` condition?
___
*[`internal/source/cdc/server/integration_test.go` line 63 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1045#-O9ohOr35qwk-mIzpZ0f:-O9ohOr35qwk-mIzpZ0g:b-80xfat) ([raw file](https://github.com/cockroachdb/replicator/blob/1d53646f4a263497352e9e723845c4d119246a36/internal/source/cdc/server/integration_test.go#L63)):*
> ```Go
> 	testModeParallel
> 	testModeWebhook
> 	testModeKeyInValue
> ```

Sort.
___
*[`internal/source/cdc/server/integration_test.go` line 74 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1045#-O9ogkG2C9HQbew-hM83:-O9ogkG33VJtWEiIfQyZ:b-80xfat) ([raw file](https://github.com/cockroachdb/replicator/blob/1d53646f4a263497352e9e723845c4d119246a36/internal/source/cdc/server/integration_test.go#L74)):*
> ```Go
> 	queries    bool
> 	webhook    bool
> 	keyInValue bool
> ```

Sort.
___
*[`internal/source/cdc/server/integration_test.go` line 121 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1045#-O9ogq951fbw0j4Rpvr-:-O9ogq951fbw0j4Rpvr0:b1zkr6k) ([raw file](https://github.com/cockroachdb/replicator/blob/1d53646f4a263497352e9e723845c4d119246a36/internal/source/cdc/server/integration_test.go#L121)):*
> ```Go
> 	// cases that don't apply to a particular target.
> 	tcs := make([]testConfig, testModeMax)
> 	tcs[0] = testConfig{
> ```

Unused assignment.
___
*[`internal/source/cdc/server/integration_test.go` line 125 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1045#-O9ogyBv61Qk3aXA9AER:-O9ogyBv61Qk3aXA9AES:b-80xfat) ([raw file](https://github.com/cockroachdb/replicator/blob/1d53646f4a263497352e9e723845c4d119246a36/internal/source/cdc/server/integration_test.go#L125)):*
> ```Go
> 		queries:    true,
> 		webhook:    true,
> 		keyInValue: false,
> ```

Sort.
___
*[`internal/source/cdc/server/integration_test.go` line 135 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1045#-O9ogyz90J7wbNm5c_Yk:-O9ogyz90J7wbNm5c_Yl:b-80xfat) ([raw file](https://github.com/cockroachdb/replicator/blob/1d53646f4a263497352e9e723845c4d119246a36/internal/source/cdc/server/integration_test.go#L135)):*
> ```Go
> 			parallel:   i&testModeParallel == testModeParallel,
> 			webhook:    i&testModeWebhook == testModeWebhook,
> 			keyInValue: i&testModeKeyInValue == testModeKeyInValue,
> ```

Sort.
___
*[`internal/source/cdc/server/integration_test.go` line 320 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1045#-O9oh1-NBz-8byCed3Za:-O9oh1-NBz-8byCed3Zb:b-x85r3p) ([raw file](https://github.com/cockroachdb/replicator/blob/1d53646f4a263497352e9e723845c4d119246a36/internal/source/cdc/server/integration_test.go#L320)):*
> ```Go
> 
> 	log.Debugf("changefeed URL is %s", feedURL.String())
> 	fmt.Println(createStmt)
> ```

Add to `log` call.
___
*[`internal/util/cdcjson/query_payload.go` line 34 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1045#-O9oh4tTE38ViCKvsTEd:-O9oh4tTE38ViCKvsTEe:b-80xfat) ([raw file](https://github.com/cockroachdb/replicator/blob/1d53646f4a263497352e9e723845c4d119246a36/internal/util/cdcjson/query_payload.go#L34)):*
> ```Go
> 	crdbLabel   = ident.New("__crdb__")
> 	updated     = ident.New("updated")
> 	keyLabel    = ident.New("key")
> ```

Sort.
___
*[`internal/util/cdcjson/query_payload_test.go` line 129 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1045#-O9ohcavB37jm8-jlkJI:-O9ohcavB37jm8-jlkJJ:b-txb038) ([raw file](https://github.com/cockroachdb/replicator/blob/1d53646f4a263497352e9e723845c4d119246a36/internal/util/cdcjson/query_payload_test.go#L129)):*
> ```Go
> 			// Mutation struct is populated from the logic for unmarshalling the
> 			// "key" field in the JSON.
> 			&ident.Map[int]{},
> ```

Using `nil` or a sentinel instance would clean this up.

ryanluu12345 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1045)* status: 0 of 4 files reviewed, 11 unresolved discussions (waiting on @bobvawter, @BramGruneir, and @Jeremyyang920)

___
*[`internal/source/cdc/server/integration_test.go` line 63 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1045#-O9ohOr35qwk-mIzpZ0f:-O9qBXgg8Oa5rtr8J-vK:b-896fix) ([raw file](https://github.com/cockroachdb/replicator/blob/1d53646f4a263497352e9e723845c4d119246a36/internal/source/cdc/server/integration_test.go#L63)):*

Sort.
</blockquote></details>

Done.
___
*[`internal/source/cdc/server/integration_test.go` line 74 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1045#-O9ogkG2C9HQbew-hM83:-O9qBb2M2DI6Tls-in_U:b-896fix) ([raw file](https://github.com/cockroachdb/replicator/blob/1d53646f4a263497352e9e723845c4d119246a36/internal/source/cdc/server/integration_test.go#L74)):*

Sort.
</blockquote></details>

Done.
___
*[`internal/source/cdc/server/integration_test.go` line 121 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1045#-O9ogq951fbw0j4Rpvr-:-O9qBn5S1JYyj11H8iWL:b-g7i742) ([raw file](https://github.com/cockroachdb/replicator/blob/1d53646f4a263497352e9e723845c4d119246a36/internal/source/cdc/server/integration_test.go#L121)):*

Unused assignment.
</blockquote></details>

Good catch, remnant of local testing.
___
*[`internal/source/cdc/server/integration_test.go` line 125 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1045#-O9ogyBv61Qk3aXA9AER:-O9qBvW3CW_ShhKCseLy:b-896fix) ([raw file](https://github.com/cockroachdb/replicator/blob/1d53646f4a263497352e9e723845c4d119246a36/internal/source/cdc/server/integration_test.go#L125)):*

Sort.
</blockquote></details>

Done.
___
*[`internal/source/cdc/server/integration_test.go` line 135 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1045#-O9ogyz90J7wbNm5c_Yk:-O9qBxX9--2au5513p2A:b-896fix) ([raw file](https://github.com/cockroachdb/replicator/blob/1d53646f4a263497352e9e723845c4d119246a36/internal/source/cdc/server/integration_test.go#L135)):*

Sort.
</blockquote></details>

Done.
___
*[`internal/source/cdc/server/integration_test.go` line 320 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1045#-O9oh1-NBz-8byCed3Za:-O9qC9SX482gRanTKQrn:b-896fix) ([raw file](https://github.com/cockroachdb/replicator/blob/1d53646f4a263497352e9e723845c4d119246a36/internal/source/cdc/server/integration_test.go#L320)):*

Add to `log` call.
</blockquote></details>

Done.
___
*[`internal/util/cdcjson/query_payload.go` line 34 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1045#-O9oh4tTE38ViCKvsTEd:-O9qCFFA1jJiYl4z91mj:b-896fix) ([raw file](https://github.com/cockroachdb/replicator/blob/1d53646f4a263497352e9e723845c4d119246a36/internal/util/cdcjson/query_payload.go#L34)):*

Sort.
</blockquote></details>

Done.
___
*[`internal/util/cdcjson/query_payload_test.go` line 129 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1045#-O9ohcavB37jm8-jlkJI:-O9qCOJn2ShSHqu56IpH:belef8f) ([raw file](https://github.com/cockroachdb/replicator/blob/1d53646f4a263497352e9e723845c4d119246a36/internal/util/cdcjson/query_payload_test.go#L129)):*

Using `nil` or a sentinel instance would clean this up.
</blockquote></details>

Yep, going with this method!

ryanluu12345 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1045)* status: 0 of 4 files reviewed, 11 unresolved discussions (waiting on @bobvawter, @BramGruneir, and @Jeremyyang920)

___
*[`internal/source/cdc/webhook_query.go` line 62 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1045#-O9ohHfmEaYSmpkAS2Oi:-O9qE5gHEx7QpzcK4Ce-:b-9f087o) ([raw file](https://github.com/cockroachdb/replicator/blob/1d53646f4a263497352e9e723845c4d119246a36/internal/source/cdc/webhook_query.go#L62)):*

Can this be `nil` or a sentinel value to indicate a `key_in_value` condition?
</blockquote></details>

Yep, done!

bobvawter (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1045)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @BramGruneir, @Jeremyyang920, and @ryanluu12345)

___
*[`internal/source/cdc/webhook_query.go` line 68 at r6](https://reviewable.io/reviews/cockroachdb/replicator/1045#-OA-Ptc3DWsf9p4tEmUI:-OA-Ptc3DWsf9p4tEmUJ:b-pq9niu) ([raw file](https://github.com/cockroachdb/replicator/blob/5355b0e9b19c6ba4503a85f11749f02cd56c55eb/internal/source/cdc/webhook_query.go#L68)):*
> ```Go
> 	// the primary key from the schema, since the values are provided by webhook
> 	// message.
> 	if len(message.Payload) > 0 && !strings.Contains(string(message.Payload[0]), `"key"`) {
> ```

What happens if the byte sequence `"key"` is embedded within the mutation data?

For instance, `UPDATE foo SET bar = '"key"'`

ryanluu12345 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1045)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @bobvawter, @BramGruneir, and @Jeremyyang920)

___
*[`internal/source/cdc/webhook_query.go` line 68 at r6](https://reviewable.io/reviews/cockroachdb/replicator/1045#-OA-Ptc3DWsf9p4tEmUI:-OA07dQu0pQDdid_iITB:bi1wrag) ([raw file](https://github.com/cockroachdb/replicator/blob/5355b0e9b19c6ba4503a85f11749f02cd56c55eb/internal/source/cdc/webhook_query.go#L68)):*

What happens if the byte sequence `"key"` is embedded within the mutation data?

For instance, `UPDATE foo SET bar = '"key"'`
</blockquote></details>

Good call out. This is very brittle. I ended up attempting to unmarshal they paylod to a map of strings to interfaces and ensure that I can find a `key` key in the map. This should give us the intended behavior. I've also gone ahead and refactored that to a helper method along with tests that exercise the paths where we skip the key extraction.

bobvawter (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1045)* status: 0 of 5 files reviewed, 5 unresolved discussions (waiting on @BramGruneir, @Jeremyyang920, and @ryanluu12345)

___
*[`internal/source/cdc/webhook_query.go` line 107 at r7](https://reviewable.io/reviews/cockroachdb/replicator/1045#-OA3HXCX-aekI0QSYzKJ:-OA3HXCX-aekI0QSYzKK:b94fjso) ([raw file](https://github.com/cockroachdb/replicator/blob/4f20ea7476786a11e7b8052a4b1551b20092056a/internal/source/cdc/webhook_query.go#L107)):*
> ```Go
> }
> 
> func (h *Handler) getKeyData(req *request, message *message) (*ident.Map[int], error) {
> ```

That's not what it does, though.
___
*[`internal/source/cdc/webhook_query.go` line 113 at r7](https://reviewable.io/reviews/cockroachdb/replicator/1045#-OA3A7YPBNQkLYxKLnyU:-OA3A7YPBNQkLYxKLnyV:b-c1p4w3) ([raw file](https://github.com/cockroachdb/replicator/blob/4f20ea7476786a11e7b8052a4b1551b20092056a/internal/source/cdc/webhook_query.go#L113)):*
> ```Go
> 
> 	var payloadMap map[string]interface{}
> 	if err := json.Unmarshal(message.Payload[0], &payloadMap); err != nil {
> ```

This requires reification of arbitrarily large payloads, only to throw the objects away.  It also can't early-out.
___
*[`internal/source/cdc/webhook_query.go` line 125 at r7](https://reviewable.io/reviews/cockroachdb/replicator/1045#-OA3Hx-D1gM0HbWJle-2:-OA3Hx-D1gM0HbWJle-3:b2rlc3z) ([raw file](https://github.com/cockroachdb/replicator/blob/4f20ea7476786a11e7b8052a4b1551b20092056a/internal/source/cdc/webhook_query.go#L125)):*
> ```Go
> 	// In the case we don't have any "key" data from the request paylod,
> 	// then we need to get the primary key from the schema.
> 	keys, err := h.getPrimaryKey(req)
> ```

Just `return`.
___
*[`internal/util/cdcjson/query_payload_test.go` line 124 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1045#-O9l1sNiCx-xqGy98id_:-OA3IVYU9jPFpJZP5t2P:bcgb5n8) ([raw file](https://github.com/cockroachdb/replicator/blob/1d53646f4a263497352e9e723845c4d119246a36/internal/util/cdcjson/query_payload_test.go#L124)):*

After reviewing the existing testing here, I think this is the simplest test that will test the conditional logic inside of UnmarshalJSON. The logic is spelled out in the comment, but as an additional point of context: since I make it so that `tt.keys` is an empty map, there is no way that the data from the projections can be mapped to key values. This means that this test passing entirely depends on the `key` field being passed from the changefeed message, thus giving us the coverage we need as simply as possible.

In order to address changes in changefeed behavior, we already have the integration_test that includes the new `key_in_value` option and tests all permutations of it. I think this should cover this behavior just fine.
</blockquote></details>

That's reasonable; these utility test cases just need to ensure adequate coverage.

ryanluu12345 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1045)* status: 0 of 5 files reviewed, 5 unresolved discussions (waiting on @bobvawter, @BramGruneir, and @Jeremyyang920)

___
*[`internal/source/cdc/webhook_query.go` line 107 at r7](https://reviewable.io/reviews/cockroachdb/replicator/1045#-OA3HXCX-aekI0QSYzKJ:-OA3YASB8MOH6C7Zna0C:b-qbj84f) ([raw file](https://github.com/cockroachdb/replicator/blob/4f20ea7476786a11e7b8052a4b1551b20092056a/internal/source/cdc/webhook_query.go#L107)):*

That's not what it does, though.
</blockquote></details>

You're right. Technically this is getting the PK columns and ordinal positions. Technically, this is something like `maybeGetPKColumns` or `getPKColumns`.
___
*[`internal/source/cdc/webhook_query.go` line 125 at r7](https://reviewable.io/reviews/cockroachdb/replicator/1045#-OA3Hx-D1gM0HbWJle-2:-OA3YKOt7sbbPu7mV9yT:b-83od3n) ([raw file](https://github.com/cockroachdb/replicator/blob/4f20ea7476786a11e7b8052a4b1551b20092056a/internal/source/cdc/webhook_query.go#L125)):*

Just `return`.
</blockquote></details>

Noted
___
*[`internal/util/cdcjson/query_payload_test.go` line 124 at r5](https://reviewable.io/reviews/cockroachdb/replicator/1045#-O9l1sNiCx-xqGy98id_:-OA3Y97R7w4JKH6FcB1D:btojbfd) ([raw file](https://github.com/cockroachdb/replicator/blob/1d53646f4a263497352e9e723845c4d119246a36/internal/util/cdcjson/query_payload_test.go#L124)):*

That's reasonable; these utility test cases just need to ensure adequate coverage.
</blockquote></details>

Great!

ryanluu12345 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1045)* status: 0 of 5 files reviewed, 5 unresolved discussions (waiting on @bobvawter, @BramGruneir, and @Jeremyyang920)

___
*[`internal/source/cdc/webhook_query.go` line 113 at r7](https://reviewable.io/reviews/cockroachdb/replicator/1045#-OA3A7YPBNQkLYxKLnyU:-OA3bpqV3jStjio397gG:b-3z58ju) ([raw file](https://github.com/cockroachdb/replicator/blob/4f20ea7476786a11e7b8052a4b1551b20092056a/internal/source/cdc/webhook_query.go#L113)):*

This requires reification of arbitrarily large payloads, only to throw the objects away.  It also can't early-out.
</blockquote></details>

Good point here, since it unmarshals all the data to the map. I've gone ahead and followed the convention used elsewhere in the codebase and made it so that we use `json.NewDecoder` for the `message.Payload` and decode into this:
```
type keyPayload struct {
	Key []any `json:"key"`
}
```

ryanluu12345 (commented):
After final review, I'll squash before merge

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1045#-:-OA3c6-B7PP3SBdPttsM:bektud)* status: 0 of 5 files reviewed, 5 unresolved discussions (waiting on @bobvawter, @BramGruneir, and @Jeremyyang920)

bobvawter (changes requested):
> Reviewed 1 of 4 files at r6.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1045)* status: 1 of 5 files reviewed, 4 unresolved discussions (waiting on @BramGruneir, @Jeremyyang920, and @ryanluu12345)

___
*[`internal/source/cdc/webhook_query.go` line 113 at r7](https://reviewable.io/reviews/cockroachdb/replicator/1045#-OA3A7YPBNQkLYxKLnyU:-OA4F3tk-cUfHZCcQkEQ:b-v92li0) ([raw file](https://github.com/cockroachdb/replicator/blob/4f20ea7476786a11e7b8052a4b1551b20092056a/internal/source/cdc/webhook_query.go#L113)):*

Good point here, since it unmarshals all the data to the map. I've gone ahead and followed the convention used elsewhere in the codebase and made it so that we use `json.NewDecoder` for the `message.Payload` and decode into this:
```
type keyPayload struct {
	Key []any `json:"key"`
}
```
</blockquote></details>

The other pattern is to declare the type locally (i.e. in the method) since it's not useful to any other function.  Also, you don't need `[]any`; use `json.RawMessage` instead.  That allows the decoder to do the absolute minimum amount of work.
___
*[`internal/source/cdc/webhook_query.go` line 52 at r9](https://reviewable.io/reviews/cockroachdb/replicator/1045#-OA4DdWE982FS57-Clwf:-OA4DdWF0111tT4A01Gp:b1j8urc) ([raw file](https://github.com/cockroachdb/replicator/blob/102d2de9e023abc0350ef9aaf72aa7a5d5dcb20e/internal/source/cdc/webhook_query.go#L52)):*
> ```Go
> 	}
> 
> 	var message message
> ```

Stutters.

```
msg := &message{}
```

Then you don't have to keep taking the address.  Also, `message` is a bit vague when looking at the contents of the package.
___
*[`internal/source/cdc/webhook_query.go` line 112 at r9](https://reviewable.io/reviews/cockroachdb/replicator/1045#-OA4EBugEAC6HMrZlY6S:-OA4EBugEAC6HMrZlY6T:brpqpd4) ([raw file](https://github.com/cockroachdb/replicator/blob/102d2de9e023abc0350ef9aaf72aa7a5d5dcb20e/internal/source/cdc/webhook_query.go#L112)):*
> ```Go
> 
> func (h *Handler) getPKColumns(req *request, message *message) (*ident.Map[int], error) {
> 	if len(message.Payload) == 0 {
> ```

This should return an error, since it doesn't make sense to get PK columns for an empty mutation.  What you should do is to move the call to this method below the `if message.Resolved != ""` case above.  That is, you've pushed what should be a (slightly) higher-level flow control decision down into a leaf method and return a nonsensical value.
___
*[`internal/util/cdcjson/query_payload.go` line 125 at r9](https://reviewable.io/reviews/cockroachdb/replicator/1045#-OA4GLT7BpjiqrTmnTg-:-OA4GLT7BpjiqrTmnTg0:bhttfeo) ([raw file](https://github.com/cockroachdb/replicator/blob/102d2de9e023abc0350ef9aaf72aa7a5d5dcb20e/internal/util/cdcjson/query_payload.go#L125)):*
> ```Go
> 			return errors.Wrap(err, "could not unmarshal 'before' field")
> 		}
> 	}
> ```

You ought to be able to return here if `keyValues` has been set.  The rest of this method is just about the extraction case and has no import on the `queryPayload` being constructed.

ryanluu12345 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1045)* status: 1 of 5 files reviewed, 4 unresolved discussions (waiting on @bobvawter, @BramGruneir, and @Jeremyyang920)

___
*[`internal/source/cdc/webhook_query.go` line 52 at r9](https://reviewable.io/reviews/cockroachdb/replicator/1045#-OA4DdWE982FS57-Clwf:-OA4JS2M75t7NJt3JAbq:b-a91zp3) ([raw file](https://github.com/cockroachdb/replicator/blob/102d2de9e023abc0350ef9aaf72aa7a5d5dcb20e/internal/source/cdc/webhook_query.go#L52)):*

Stutters.

```
msg := &message{}
```

Then you don't have to keep taking the address.  Also, `message` is a bit vague when looking at the contents of the package.
</blockquote></details>

I'll call it changefeed message.

ryanluu12345 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1045)* status: 1 of 5 files reviewed, 4 unresolved discussions (waiting on @bobvawter, @BramGruneir, and @Jeremyyang920)

___
*[`internal/source/cdc/webhook_query.go` line 113 at r7](https://reviewable.io/reviews/cockroachdb/replicator/1045#-OA3A7YPBNQkLYxKLnyU:-OA4LAPA4f366HmVhzEN:b-75m1lu) ([raw file](https://github.com/cockroachdb/replicator/blob/4f20ea7476786a11e7b8052a4b1551b20092056a/internal/source/cdc/webhook_query.go#L113)):*

The other pattern is to declare the type locally (i.e. in the method) since it's not useful to any other function.  Also, you don't need `[]any`; use `json.RawMessage` instead.  That allows the decoder to do the absolute minimum amount of work.
</blockquote></details>

Updated it to look like this:
```
payload := &struct {
		Key json.RawMessage `json:"key"`
	}{}
	decoder := json.NewDecoder(bytes.NewReader(message.Payload[0]))
	if err := decoder.Decode(payload); err != nil {
		return nil, err
	}
```
Good point, didn't need the type to be used by any other function.
___
*[`internal/source/cdc/webhook_query.go` line 112 at r9](https://reviewable.io/reviews/cockroachdb/replicator/1045#-OA4EBugEAC6HMrZlY6S:-OA4KaTK2KZHSjtQfIT5:bmntnlq) ([raw file](https://github.com/cockroachdb/replicator/blob/102d2de9e023abc0350ef9aaf72aa7a5d5dcb20e/internal/source/cdc/webhook_query.go#L112)):*

This should return an error, since it doesn't make sense to get PK columns for an empty mutation.  What you should do is to move the call to this method below the `if message.Resolved != ""` case above.  That is, you've pushed what should be a (slightly) higher-level flow control decision down into a leaf method and return a nonsensical value.
</blockquote></details>

Done. Returning an error for the empty payload case: `errors.New("cannot get PK columns for an empty mutation")` and moved it to below that check.
___
*[`internal/util/cdcjson/query_payload.go` line 125 at r9](https://reviewable.io/reviews/cockroachdb/replicator/1045#-OA4GLT7BpjiqrTmnTg-:-OA4MQte2iki8EFD9RTi:b-896fix) ([raw file](https://github.com/cockroachdb/replicator/blob/102d2de9e023abc0350ef9aaf72aa7a5d5dcb20e/internal/util/cdcjson/query_payload.go#L125)):*

You ought to be able to return here if `keyValues` has been set.  The rest of this method is just about the extraction case and has no import on the `queryPayload` being constructed.
</blockquote></details>

Done.

bobvawter (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1045)* status: 1 of 5 files reviewed, 7 unresolved discussions (waiting on @BramGruneir, @Jeremyyang920, and @ryanluu12345)

___
*[`internal/source/cdc/handler_test.go` line 624 at r10](https://reviewable.io/reviews/cockroachdb/replicator/1045#-OAIkASaBgPss3AaqTU7:-OAIkASaBgPss3AaqTU8:br5vt6s) ([raw file](https://github.com/cockroachdb/replicator/blob/b211e6e7f2d718c2a5486df915bf47db964f6d3e/internal/source/cdc/handler_test.go#L624)):*
> ```Go
> 		request   *request
> 		wantError error
> 		keys      *ident.Map[int]
> ```

This field is always `nil`.  Are you missing a test case?
___
*[`internal/source/cdc/webhook_query.go` line 32 at r10](https://reviewable.io/reviews/cockroachdb/replicator/1045#-OAIk-QR0VJ2v54dQeFq:-OAIk-QR0VJ2v54dQeFr:bm5lu7) ([raw file](https://github.com/cockroachdb/replicator/blob/b211e6e7f2d718c2a5486df915bf47db964f6d3e/internal/source/cdc/webhook_query.go#L32)):*
> ```Go
> )
> 
> // message represents the webhook message from the changefeed.
> ```

Update type name.
___
*[`internal/source/cdc/webhook_query.go` line 37 at r10](https://reviewable.io/reviews/cockroachdb/replicator/1045#-OAIjzKm2lf1RwM_wCP1:-OAIjzKm2lf1RwM_wCP2:b-80xfat) ([raw file](https://github.com/cockroachdb/replicator/blob/b211e6e7f2d718c2a5486df915bf47db964f6d3e/internal/source/cdc/webhook_query.go#L37)):*
> ```Go
> 	Length  int               `json:"length"`
> 	// With envelope="bare" (default for queries), there is a  `__crdb__` property.
> 	Bare json.RawMessage `json:"__crdb__"`
> ```

Sort.
___
*[`internal/source/cdc/webhook_query.go` line 63 at r10](https://reviewable.io/reviews/cockroachdb/replicator/1045#-OAIinVu6vCGK6Bd3lvO:-OAIinVv3TZONI4HrISp:b-pc7nn9) ([raw file](https://github.com/cockroachdb/replicator/blob/b211e6e7f2d718c2a5486df915bf47db964f6d3e/internal/source/cdc/webhook_query.go#L63)):*
> ```Go
> 		return errors.Wrap(err, "could not decode payload")
> 	}
> ```

Whitespace diff.
___
*[`internal/source/cdc/webhook_query.go` line 83 at r10](https://reviewable.io/reviews/cockroachdb/replicator/1045#-OAIjL4XEOm53Qo-u4oQ:-OAIjL4XEOm53Qo-u4oR:b5igdry) ([raw file](https://github.com/cockroachdb/replicator/blob/b211e6e7f2d718c2a5486df915bf47db964f6d3e/internal/source/cdc/webhook_query.go#L83)):*
> ```Go
> 	// the primary key from the schema, since the values are provided by webhook
> 	// message.
> 	keys, err := h.getPKColumns(req, msg)
> ```

Rename the `keys` variable to `keysToExtract`; that makes it more obvious that the work is conditional.
___
*[`internal/source/cdc/webhook_query.go` line 124 at r10](https://reviewable.io/reviews/cockroachdb/replicator/1045#-OAIjUZX34bjEvktBVYo:-OAIjUZX34bjEvktBVYp:bolknbu) ([raw file](https://github.com/cockroachdb/replicator/blob/b211e6e7f2d718c2a5486df915bf47db964f6d3e/internal/source/cdc/webhook_query.go#L124)):*
> ```Go
> 	// schema.
> 	if len(payload.Key) > 0 {
> 		return nil, nil
> ```

This return is somewhat atypical, so it should be documented on the method.
___
*[`internal/util/cdcjson/query_payload.go` line 47 at r10](https://reviewable.io/reviews/cockroachdb/replicator/1045#-OAIjoZU8RBEi-e1wLVz:-OAIjoZU8RBEi-e1wLW-:b-b1u8q4) ([raw file](https://github.com/cockroachdb/replicator/blob/b211e6e7f2d718c2a5486df915bf47db964f6d3e/internal/util/cdcjson/query_payload.go#L47)):*
> ```Go
> 	after     *ident.Map[json.RawMessage]
> 	before    *ident.Map[json.RawMessage]
> 	keys      *ident.Map[int]
> ```

Also rename this field `keysToExtract`

ryanluu12345 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1045)* status: 1 of 5 files reviewed, 7 unresolved discussions (waiting on @bobvawter, @BramGruneir, and @Jeremyyang920)

___
*[`internal/source/cdc/webhook_query.go` line 32 at r10](https://reviewable.io/reviews/cockroachdb/replicator/1045#-OAIk-QR0VJ2v54dQeFq:-OAJ1qpy8YVDFLzTnEWQ:b-896fix) ([raw file](https://github.com/cockroachdb/replicator/blob/b211e6e7f2d718c2a5486df915bf47db964f6d3e/internal/source/cdc/webhook_query.go#L32)):*

Update type name.
</blockquote></details>

Done.
___
*[`internal/source/cdc/webhook_query.go` line 37 at r10](https://reviewable.io/reviews/cockroachdb/replicator/1045#-OAIjzKm2lf1RwM_wCP1:-OAJ1wkl37VwqS1W1lZW:b-896fix) ([raw file](https://github.com/cockroachdb/replicator/blob/b211e6e7f2d718c2a5486df915bf47db964f6d3e/internal/source/cdc/webhook_query.go#L37)):*

Sort.
</blockquote></details>

Done.
___
*[`internal/source/cdc/webhook_query.go` line 63 at r10](https://reviewable.io/reviews/cockroachdb/replicator/1045#-OAIinVu6vCGK6Bd3lvO:-OAJ22IqBDWE-rDbY_qs:b-896fix) ([raw file](https://github.com/cockroachdb/replicator/blob/b211e6e7f2d718c2a5486df915bf47db964f6d3e/internal/source/cdc/webhook_query.go#L63)):*

Whitespace diff.
</blockquote></details>

Done.
___
*[`internal/source/cdc/webhook_query.go` line 83 at r10](https://reviewable.io/reviews/cockroachdb/replicator/1045#-OAIjL4XEOm53Qo-u4oQ:-OAJ26DI38yuyeW_o1PO:b-896fix) ([raw file](https://github.com/cockroachdb/replicator/blob/b211e6e7f2d718c2a5486df915bf47db964f6d3e/internal/source/cdc/webhook_query.go#L83)):*

Rename the `keys` variable to `keysToExtract`; that makes it more obvious that the work is conditional.
</blockquote></details>

Done.
___
*[`internal/source/cdc/webhook_query.go` line 124 at r10](https://reviewable.io/reviews/cockroachdb/replicator/1045#-OAIjUZX34bjEvktBVYo:-OAJ3Ey22zoeQ6bzDPSF:b-896fix) ([raw file](https://github.com/cockroachdb/replicator/blob/b211e6e7f2d718c2a5486df915bf47db964f6d3e/internal/source/cdc/webhook_query.go#L124)):*

This return is somewhat atypical, so it should be documented on the method.
</blockquote></details>

Done.
___
*[`internal/util/cdcjson/query_payload.go` line 47 at r10](https://reviewable.io/reviews/cockroachdb/replicator/1045#-OAIjoZU8RBEi-e1wLVz:-OAJ2Iui23QhkQ4RA1cL:b-896fix) ([raw file](https://github.com/cockroachdb/replicator/blob/b211e6e7f2d718c2a5486df915bf47db964f6d3e/internal/util/cdcjson/query_payload.go#L47)):*

Also rename this field `keysToExtract`
</blockquote></details>

Done.

ryanluu12345 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1045)* status: 1 of 6 files reviewed, 7 unresolved discussions (waiting on @bobvawter, @BramGruneir, and @Jeremyyang920)

___
*[`internal/source/cdc/handler_test.go` line 624 at r10](https://reviewable.io/reviews/cockroachdb/replicator/1045#-OAIkASaBgPss3AaqTU7:-OAJ8HpD1WbqMEvYGFes:b-uu3gr3) ([raw file](https://github.com/cockroachdb/replicator/blob/b211e6e7f2d718c2a5486df915bf47db964f6d3e/internal/source/cdc/handler_test.go#L624)):*

This field is always `nil`.  Are you missing a test case?
</blockquote></details>

Added a final test case for when the `key` field is not present and we need to get the column to index mapping.

bobvawter (approved):
:lgtm: w/ nit

> Reviewed 5 of 5 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1045#-:-OAJCH2G9-3t9lPEIAmR:b-uobk8w)* status: all files reviewed, 1 unresolved discussion (waiting on @BramGruneir, @Jeremyyang920, and @ryanluu12345)

___
*[`internal/source/cdc/handler_test.go` line 619 at r11](https://reviewable.io/reviews/cockroachdb/replicator/1045#-OAJAxjt9RH2yuvDBOM0:-OAJAxjt9RH2yuvDBOM1:biqxu1x) ([raw file](https://github.com/cockroachdb/replicator/blob/20f8cd4866c605ac5fd632fa58f14a63976a0fbb/internal/source/cdc/handler_test.go#L619)):*
> ```Go
> 	h := fixture.Handler
> 
> 	pkMap := ident.Map[int]{}
> ```

https://pkg.go.dev/github.com/cockroachdb/replicator/internal/util/ident#MapOf makes this easer:

```
pkMap := ident.MapOf[int]("pk", 0)
```

ryanluu12345 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1045)* status: all files reviewed, 1 unresolved discussion (waiting on @bobvawter, @BramGruneir, and @Jeremyyang920)

___
*[`internal/source/cdc/handler_test.go` line 619 at r11](https://reviewable.io/reviews/cockroachdb/replicator/1045#-OAJAxjt9RH2yuvDBOM0:-OAJCnZV7jGKHYRaMrnY:b-x8ujw) ([raw file](https://github.com/cockroachdb/replicator/blob/20f8cd4866c605ac5fd632fa58f14a63976a0fbb/internal/source/cdc/handler_test.go#L619)):*

https://pkg.go.dev/github.com/cockroachdb/replicator/internal/util/ident#MapOf makes this easer:

```
pkMap := ident.MapOf[int]("pk", 0)
```
</blockquote></details>

Oh nice! Thanks! I went ahead and just put this inline so I don't have to assign to a variable:
```
keys:      ident.MapOf[int]("pk", 0),
```

ryanluu12345 (commented):
Great, thanks!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1045#-:-OAJCtMs4eFvVImJy4K2:b-6mhyx0)* status: all files reviewed, 1 unresolved discussion (waiting on @bobvawter, @BramGruneir, and @Jeremyyang920)

Issue #1047, opened 2024-10-18
Make the schema watch refresh rate configurable


ryanluu12345:
Currently, the schema watch refresh is configured to 1 minute by default and is not able to be overridden. The ask here is to support being able to set this time from a cmd line flag. The default should still be 1 minute for backwards compatibility reasons.

PR #1048, opened 2024-10-19
internal/target/schemawatch: added in configuration wiring for the RefreshDelay option

ryanluu12345:
Previously, the RefreshDelay was set to a constant value of 1 minute, which means that customers did not have a way to opt out or update this behavior so it happens less or more frequently. This change introduces the wiring from each frontend to the schemawatch package to control this schema refresh timing.

Resolves: #1047
Release Note: Added in the ability to configure the schema watcher refresh delay via a CLI flag.

**TODO**
- [x] Fix all broken tests: note that all tests that broke are because of: `Error: The operation was canceled.`, which indicates there was a timeout for the test, which matches what we see
- [x] Wiki updates with the latest new flag here

**Manual Testing**
Put log lines right before the schema watch go routine kicks off and checks on an interval and see that it takes the value we pass in.
```
go run . start ..... --schemaRefresh '2m'
...
time="Oct 18 17:05:18" level=info msg="Replicator starting" -buildmode=exe -compiler=gc -gcflags="all=-N -l" CGO_CFLAGS="-O0 -g" CGO_CPPFLAGS= CGO_CXXFLAGS= CGO_ENABLED=1 CGO_LDFLAGS= GOARCH=arm64 GOARM64=v8.0 GOOS=darwin
time="Oct 18 17:05:18" level=info msg="authentication disabled, any caller may write to the target database"
time="Oct 18 17:05:18" level=info msg="Server listening" address="[::]:30004"
time="Oct 18 17:05:18" level=info msg="Version 8.0.39."
time="Oct 18 17:05:18" level=debug msg="Mode REAL_AS_FLOAT,PIPES_AS_CONCAT,ANSI_QUOTES,IGNORE_SPACE,ONLY_FULL_GROUP_BY,ANSI. Ssl_cipher "
2m0s
delay is:  2m0s
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1048)


ryanluu12345 (commented):
For changes that touch configurations and how a user interfaces with the replicator CLI, what do you recommend as the best way to  add test coverage for this?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1048#-:-O9XEtjw2WNZokyXsXMs:b5gmkdj)* status: 0 of 24 files reviewed, all discussions resolved (waiting on @bobvawter and @Jeremyyang920)

ryanluu12345:
Just wanted to make a note here that I've tried fixing test in various ways with the following results:
1. Default configuration where the ProvideSchemaWatcher is initialized with refresh of 1s -> the tests all just time out at around 10 minutes; something here is causing the tests to hang
2. Updating the tests to use the default 1 minute refresh timeout -> the schemawatch test predictably takes 1 minute and one test case times out on the channel
3. Making the ProvideSchemaWatcher duration configurable from the fixture.NewBase etc. -> pending (however this is the same exact behavior as before where each test has 1 minute refresh time and the schema watch test only has 1 second to make it operate fast)

I'm wondering why setting the blanket refresh of 1s makes it so the test times out. Does the schema reloading end up locking things within replicator so that data movement cannot advance? That's my most reasonable theory right now.

Furthermore, I don't like the configurable schema refresh duration that much since it causes a change to all the fixture call sites. Although if this is a pattern we want to standardize and extend my implementation so it considers a fixtureConfig struct instead of just the "duration", this may be more flexible for future helpers we may want.

ryanluu12345 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1048)* status: 0 of 49 files reviewed, 2 unresolved discussions (waiting on @bobvawter and @Jeremyyang920)

___
*[`internal/cmd/workload/workload_test.go` line 42 at r3](https://reviewable.io/reviews/cockroachdb/replicator/1048#-O9_zrju7AXNASO47Er6:-O9_zrjvBnX8VP7jOHJg:b-pkk417) ([raw file](https://github.com/cockroachdb/replicator/blob/debf6ecc27c53e3550c1fccaa0836f8edb27a799/internal/cmd/workload/workload_test.go#L42)):*
> ```Go
> 	r := require.New(t)
> 
> 	fixture, err := all.NewFixture(t, time.Minute)
> ```

Later on, I intend to make this the same variable from `schema_watch` or export it so we don't copy/paste this.
___
*[`internal/sinktest/all/injector.go` line 32 at r3](https://reviewable.io/reviews/cockroachdb/replicator/1048#-O9_zyBB9QAyPBGYidaN:-O9_zyBC5H8wrMtkWsDT:bppazwv) ([raw file](https://github.com/cockroachdb/replicator/blob/debf6ecc27c53e3550c1fccaa0836f8edb27a799/internal/sinktest/all/injector.go#L32)):*
> ```Go
> // NewFixture constructs a self-contained test fixture for all services
> // in the target sub-packages.
> func NewFixture(t testing.TB, dur time.Duration) (*Fixture, error) {
> ```

Later on I plan to make this more general and use a `Config` option that contains different options for configuring fixtures.

bobvawter (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1048)* status: 0 of 49 files reviewed, 3 unresolved discussions (waiting on @Jeremyyang920 and @ryanluu12345)

___
*[`internal/sinktest/all/injector.go` line 32 at r3](https://reviewable.io/reviews/cockroachdb/replicator/1048#-O9_zyBB9QAyPBGYidaN:-O9jV8hNE1MbzlJ1-nM7:b-uize1f) ([raw file](https://github.com/cockroachdb/replicator/blob/debf6ecc27c53e3550c1fccaa0836f8edb27a799/internal/sinktest/all/injector.go#L32)):*

Later on I plan to make this more general and use a `Config` option that contains different options for configuring fixtures.
</blockquote></details>

The right way to do this is to leave the `NewFixture` signature alone and to introduce a binding type per https://github.com/google/wire/blob/main/docs/best-practices.md#distinguishing-types

This avoids needing to update all injector constructions when the change is relevant only to a single configuration.

```go
type RefreshDelay time.Duration

func NewFixture(t testing.TB) (*Fixture, error) {
  panic(wire.Build(TestSet, wire.Value(RefreshDelay(time.Minute)))
}

func NewFixtureWithRefresh(t testing.TB, d RefreshDelay) (*Fixture, error) {
  panic(wire.Build(TestSet))
}
```
___
*[`internal/source/objstore/config.go` line 142 at r3](https://reviewable.io/reviews/cockroachdb/replicator/1048#-O9jWTGg8Mckhjjy-N3V:-O9jWTGg8Mckhjjy-N3W:bi1pz3o) ([raw file](https://github.com/cockroachdb/replicator/blob/debf6ecc27c53e3550c1fccaa0836f8edb27a799/internal/source/objstore/config.go#L142)):*
> ```Go
> 	Target            sinkprod.TargetConfig
> 	TLS               secure.Config
> 	SchemaWatchConfig schemawatch.Config
> ```

Keep fields and references alphabetized; it makes it easier in the future to double-check that all calls to preflight, bind, or whatnot are in sync.

ryanluu12345 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1048)* status: 0 of 49 files reviewed, 3 unresolved discussions (waiting on @bobvawter and @Jeremyyang920)

___
*[`internal/sinktest/all/injector.go` line 32 at r3](https://reviewable.io/reviews/cockroachdb/replicator/1048#-O9_zyBB9QAyPBGYidaN:-O9k7Ux709IPI7Cj_uev:b-xhxc58) ([raw file](https://github.com/cockroachdb/replicator/blob/debf6ecc27c53e3550c1fccaa0836f8edb27a799/internal/sinktest/all/injector.go#L32)):*

The right way to do this is to leave the `NewFixture` signature alone and to introduce a binding type per https://github.com/google/wire/blob/main/docs/best-practices.md#distinguishing-types

This avoids needing to update all injector constructions when the change is relevant only to a single configuration.

```go
type RefreshDelay time.Duration

func NewFixture(t testing.TB) (*Fixture, error) {
  panic(wire.Build(TestSet, wire.Value(RefreshDelay(time.Minute)))
}

func NewFixtureWithRefresh(t testing.TB, d RefreshDelay) (*Fixture, error) {
  panic(wire.Build(TestSet))
}
```
</blockquote></details>

Great recommendation here. A lot cleaner and makes it so I don't have to update all call sites. I went ahead and made this change and the diffs are a lot cleaner now.
___
*[`internal/source/objstore/config.go` line 142 at r3](https://reviewable.io/reviews/cockroachdb/replicator/1048#-O9jWTGg8Mckhjjy-N3V:-O9k7a_aEjMPu3K-gfWE:b6srk9f) ([raw file](https://github.com/cockroachdb/replicator/blob/debf6ecc27c53e3550c1fccaa0836f8edb27a799/internal/source/objstore/config.go#L142)):*

Keep fields and references alphabetized; it makes it easier in the future to double-check that all calls to preflight, bind, or whatnot are in sync.
</blockquote></details>

Noted, I went ahead and did a pass on all the definitions and references and calls to bind/preflight so that they occur alphabetically.

ryanluu12345:
@bobvawter just a heads up I think this is ready for a final review. I just cleaned things up and rebased + squashed so the commit history is cleaner.

bobvawter (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1048)* status: 0 of 49 files reviewed, 5 unresolved discussions (waiting on @Jeremyyang920 and @ryanluu12345)

___
*[`internal/sinktest/all/injector.go` line 32 at r7](https://reviewable.io/reviews/cockroachdb/replicator/1048#-O9trHwB3NN3b16PMJZz:-O9trHwB3NN3b16PMJ_-:bhbh4a5) ([raw file](https://github.com/cockroachdb/replicator/blob/a5f3de8721f4d7ddc8a54d043ef7e1c404eefd67/internal/sinktest/all/injector.go#L32)):*
> ```Go
> // This is needed as a variable so that we can get the pointer
> // to this time.Duration for our `NewFixture` methods.
> var refreshDelay = time.Minute
> ```

This doesn't make sense and potentially contaminates across test cases.
___
*[`internal/sinktest/all/provider.go` line 82 at r7](https://reviewable.io/reviews/cockroachdb/replicator/1048#-O9trWCa0dSjOZy-CGHW:-O9trWCa0dSjOZy-CGHX:bjipf05) ([raw file](https://github.com/cockroachdb/replicator/blob/a5f3de8721f4d7ddc8a54d043ef7e1c404eefd67/internal/sinktest/all/provider.go#L82)):*
> ```Go
> // that has a refresh delay of one second since it test scenarios,
> // we want the schema watch to be fast.
> func ProvideSchemaWatchConfig(refreshDelay *time.Duration) (*schemawatch.Config, error) {
> ```

Use a named type, as previously discussed.  It doesn't need to be a pointer, either.

bobvawter (commented):
It would also make sense for an explicit 0 value to disable the query entirely, since it's only for metrics purposes.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1048#-:-O9tvtIBC7S2vpCcBZ1V:b-89nox9)* status: 0 of 49 files reviewed, 5 unresolved discussions (waiting on @Jeremyyang920 and @ryanluu12345)

bobvawter (approved):
:lgtm: w/ nits

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1048#-:-OA3krWKCV3o9_QUPDgp:b-rjhpvx)* status: 0 of 49 files reviewed, 3 unresolved discussions (waiting on @Jeremyyang920 and @ryanluu12345)

___
*[`internal/sinktest/all/provider.go` line 93 at r10](https://reviewable.io/reviews/cockroachdb/replicator/1048#-OA3jyK75GhTuQBezsej:-OA3jyK75GhTuQBezsek:b-ill4uq) ([raw file](https://github.com/cockroachdb/replicator/blob/44ae1629b3cf70b818f427b33bef930397942cd9/internal/sinktest/all/provider.go#L93)):*
> ```Go
> // ProvideSchemaWatchConfig emits a default configuration, for testing
> // that has a refresh delay of one second since it test scenarios,
> // we want the schema watch to be fast.
> ```

This comment seems out of date.
___
*[`internal/source/kafka/config.go` line 52 at r10](https://reviewable.io/reviews/cockroachdb/replicator/1048#-OA3k4CzCLN8qg7CyQwy:-OA3k4CzCLN8qg7CyQwz:b-9hmxk2) ([raw file](https://github.com/cockroachdb/replicator/blob/44ae1629b3cf70b818f427b33bef930397942cd9/internal/source/kafka/config.go#L52)):*
> ```Go
> 	Conveyor          conveyor.Config
> 	DLQ               dlq.Config
> 	SchemaWatchConfig schemawatch.Config
> ```

In this and the other `Config` types, name this field `SchemaWatch` for consistency with surrounding fields.
___
*[`internal/target/schemawatch/config.go` line 34 at r10](https://reviewable.io/reviews/cockroachdb/replicator/1048#-OA3ja6d6_NtSg7YKVYP:-OA3ja6d6_NtSg7YKVYQ:b-pcgnmd) ([raw file](https://github.com/cockroachdb/replicator/blob/44ae1629b3cf70b818f427b33bef930397942cd9/internal/target/schemawatch/config.go#L34)):*
> ```Go
> // Bind adds configuration flags to the set.
> func (c *Config) Bind(f *pflag.FlagSet) {
> 	f.DurationVar(&c.RefreshDelay, "schemaRefresh", defaultRefreshDelay, `controls how often a watcher will refresh its schema. If this value is zero or negative, refresh behavior will be disabled.`)
> ```

Reflow to 100 chars.

ryanluu12345:
Adding this to the merge queue and just updated the wiki so that it contains the `schemaRefresh` flag now (along with all of the latest flags)

Issue #1051, opened 2024-10-23
Make the CheckConsistency run less frequently


ryanluu12345:
The default should run less frequently than it currently does. This should also be configurable. The impact here is that if it runs less frequently, alerting analytics will be slower to fire leading to a bigger window for inconsistency.

ryanluu12345:
@bobvawter I've taken an initial look at this and have a high level idea of how to implement this. The current thinking I have here is to:
1. Drill the check consistency duration inside of the `factory` struct inside of the stage package. The key reason for this is that we can pass in a new config object inside of `stage.ProvideFactory` that lets the frontends configure this via a flag.
2. Update the `newStage` so that it takes in this value from the factory struct field
3. Update the hardcoded duration in the polling logic so that it uses this drilled in value

I don't have questions about the above methodology, but I do have open questions about alternatives I'm considering:
1. Currently we have this query running every minute. I'm wary there is an implication of reporting consistency errors slower if we dial this down to be less frequent. I'm currently debating between 1.5 minutes to 2 minutes (leaning towards the latter). What are your thoughts here? Or do you think it should be even less frequent?
2. I see that in some of the other frontend configs (i.e. kafka, mylogical, pglogical) we already have a `Staging` configuration object. Originally I was thinking I could make a new `config.go` inside of the `internal/staging/stage` package, but after seeing there is already a `Staging` package, wondering if I should put it inside of there? The one consideration I have here is that if we do it this way, then we'll need to have `stage`'s `ProvideFactory` reference the `sinkprod.StagingConfig`.

bobvawter:
I'm just about done implementing this.  I'd like to present it to the customer this afternoon.

ryanluu12345:
Noted. Thanks! Please let me know when it is ready and I can review the PR.

PR #1053, opened 2024-10-24
stage: Add a configuration object to the stage package.

bobvawter:
This change brings the stage package in line with other packages that provide a configuration object and flag bindings.  Several timing and sizing values around the consistency chack are added to make it less intensive on larger deployments.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1053)


ryanluu12345 (approved):
> Reviewed 27 of 27 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1053)* status: all files reviewed, 2 unresolved discussions (waiting on @bobvawter, @BramGruneir, and @sravotto)

___
*[`internal/staging/stage/config.go` line 27 at r1](https://reviewable.io/reviews/cockroachdb/replicator/1053#-O9zg-ehD496n4dDCIpg:-O9zg-ehD496n4dDCIph:b4apma4) ([raw file](https://github.com/cockroachdb/replicator/blob/c289ec9f30e7843eaf562cae9bb77f7401a1803b/internal/staging/stage/config.go#L27)):*
> ```Go
> const (
> 	defaultMarkAppliedBatchSize = 100_000
> 	defaultSanityCheckPeriod    = 10 * time.Minute
> ```

Previously, if I'm not mistaken the sanity check period was 1 minute. How did we arrive at this value that is 10x less frequent? Since the implication is that alerts can fire later for inconsistent data, wondering what potential window of inconsistency is acceptable for customers.

Although, since it's configurable now, customers can just set the value they deem appropriate.
___
*[`internal/staging/stage/stage.go` line 196 at r1](https://reviewable.io/reviews/cockroachdb/replicator/1053#-O9zh4WT-xF-eOmlFqPX:-O9zh4WT-xF-eOmlFqPY:bxtmy82) ([raw file](https://github.com/cockroachdb/replicator/blob/c289ec9f30e7843eaf562cae9bb77f7401a1803b/internal/staging/stage/stage.go#L196)):*
> ```Go
> 	// Report unapplied mutations on a periodic basis.
> 	ctx.Go(func(ctx *stopper.Context) error {
> 		if f.cfg.UnappliedPeriod <= 0 {
> ```

Just to clarify, if any of these configurations are lte 0, we just disable this check right?

bobvawter (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1053)* status: all files reviewed, 2 unresolved discussions (waiting on @BramGruneir, @ryanluu12345, and @sravotto)

___
*[`internal/staging/stage/config.go` line 27 at r1](https://reviewable.io/reviews/cockroachdb/replicator/1053#-O9zg-ehD496n4dDCIpg:-O9zu0FX9lf0ITLs4Ksl:b7zua1r) ([raw file](https://github.com/cockroachdb/replicator/blob/c289ec9f30e7843eaf562cae9bb77f7401a1803b/internal/staging/stage/config.go#L27)):*

Previously, if I'm not mistaken the sanity check period was 1 minute. How did we arrive at this value that is 10x less frequent? Since the implication is that alerts can fire later for inconsistent data, wondering what potential window of inconsistency is acceptable for customers.

Although, since it's configurable now, customers can just set the value they deem appropriate.
</blockquote></details>

All defaults are wrong for somebody.  This query turns out to be much heavier weight than anticipated, so reducing the frequency seemed like a good idea.
___
*[`internal/staging/stage/stage.go` line 196 at r1](https://reviewable.io/reviews/cockroachdb/replicator/1053#-O9zh4WT-xF-eOmlFqPX:-O9ztzvnCMWkgFQ0jmeJ:btqqu5l) ([raw file](https://github.com/cockroachdb/replicator/blob/c289ec9f30e7843eaf562cae9bb77f7401a1803b/internal/staging/stage/stage.go#L196)):*

Just to clarify, if any of these configurations are lte 0, we just disable this check right?
</blockquote></details>

Correct.

ryanluu12345:
This resolves #1051

ryanluu12345 (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1053)* status: all files reviewed, 1 unresolved discussion (waiting on @bobvawter, @BramGruneir, and @sravotto)

___
*[`internal/staging/stage/config.go` line 27 at r1](https://reviewable.io/reviews/cockroachdb/replicator/1053#-O9zg-ehD496n4dDCIpg:-OA-URex33rtw6pcm6GC:be08et8) ([raw file](https://github.com/cockroachdb/replicator/blob/c289ec9f30e7843eaf562cae9bb77f7401a1803b/internal/staging/stage/config.go#L27)):*

All defaults are wrong for somebody.  This query turns out to be much heavier weight than anticipated, so reducing the frequency seemed like a good idea.
</blockquote></details>

Yep, agreed that we should reduce the frequency. Well, it's configurable, so no concerns here.

PR #1065, opened 2024-11-06
source: use pull model for mylogical and pglogical

ZhouXing19:
This PR is based on the branch
https://github.com/cockroachdb/replicator/tree/bob_core_open to modify the mylogical and pglogical frontend to use a pull based model which is enabled by the BatchReader interface.


internal/sequencer/besteffort/best_effort.go
```diff
@@ -147,46 +148,68 @@ func (s *bestEffort) Start(
...
-					log.WithError(err).Warn("error while shutting down previous BestEffort")
-				}
 				return nil
 			})
 		return err
 	})
 
-	return ret, stats, nil
+	ctx.Go(func(ctx *stopper.Context) error {
+		return ret.Fanout(ctx)
```
ryanluu12345, line 174: What's the purpose of doing a fanout here? Does this allow us to perform batching across different tables? How is this fan out affected if they have FK constraints?

internal/sequencer/besteffort/router.go
```diff
@@ -31,74 +31,100 @@ type router struct {
...
 
-	for comp, acc := range componentBatches {
-		destination, ok := cfg.routes[comp]
-		if !ok {
-			return errors.Errorf("encountered unmapped schema component %s", comp)
-		}
-		if err := destination.AcceptMultiBatch(ctx, acc, opts); err != nil {
-			return err
-		}
+// AcceptMultiBatch splits the batch based on destination tables.
```
ryanluu12345, line 62: Thanks for the helpful note. So that means for the PG and MY case, previously we would have had to send updates to any of the target tables single threaded:
```
table1 / mut 1 -> table 2 / mut 1 -> table2 / mut 2 -> table 1 / mut 2
```

But now we can fan it out so that we do in separate goroutines.
```
table1 / mut 1 -> table1 / mut 2
table2 / mut 1 -> table2 / mut 2
```

internal/sequencer/besteffort/router.go
```diff
@@ -31,74 +31,100 @@ type router struct {
...
+		return err
 	}
+	cursorCnt := 0
+	for {
+		componentBatches := make(map[*types.SchemaComponent]*types.TemporalBatch, len(cfg.routes))
+		select {
+		// Read from the source input.
+		case cursor, ok := <-inputChan:
+			cursorCnt++
+			for name, mut := range cursor.Batch.Data.All() {
```
ryanluu12345, line 76: Beyond fanning out, do we do any batch applying for the single table case? Like within single tables, can we do like:
```
(table1 / mut 1 + table1 / mut 2 + table1 / mut 3) as one batch apply
```

internal/sequencer/seqtest/seqtest.go
```diff
@@ -67,7 +67,7 @@ func (f *Fixture) SequencerFor(
 	case switcher.ModeConsistent:
 		return f.Staging.Wrap(ctx, f.Core)
 	case switcher.ModeImmediate:
-		return f.Immediate, nil
+		return f.Core, nil
```
ryanluu12345, line 70: Are we now using `.Core` since there is no more `Immediate` after this change?

internal/sequencer/staging/staging.go
```diff
@@ -72,3 +87,70 @@ func (s *staging) Start(
 	_, stats, err := s.delegate.Start(ctx, opts)
 	return &acceptor{s.Staging}, stats, err
 }
+
+func (s *staging) StageMutations(
```
ryanluu12345, line 91: Just to clarify, but is this `StageMutations` only relevant for modes that touch the staging table for mutations storage? C2X cases I mean.

internal/source/mylogical/conn.go
```diff
@@ -183,33 +200,15 @@ func (c *conn) accumulateBatch(
...
-			if err != nil {
-				return nil, errors.WithStack(err)
-			}
-			defer tx.Rollback()
-
-			if err := c.acceptor.AcceptTemporalBatch(ctx, batch, &types.AcceptOptions{
-				TargetQuerier: tx,
-			}); err != nil {
-				return nil, err
+			cursor := &types.BatchCursor{
```
ryanluu12345, line 203: So previously, we were accepting a single batch at once and then flushing single threaded on the target. What is the implication of using the `BatchCursor`? Does this now get us to use multibatches? 

Who is the receiver of this `out` channel where cursors are written too? Is that the core sequencer that will then do all the batching within and across tables?

ryanluu12345 (commented):
Left some comments here. My remaining thought is this: because this is such an all encompassing change, we risk regressions across all operating modes.

Once we continue this work, I don't think we should merge and release all of this at once. Here is how I'm thinking about the sequencing (pun intended):
- Make the core changes to the sequencer and relevant interfaces + implementations as their own standalone change
- Make the changes for PG and MySQL first since those will see the most immediate performance benefit
- Perform testing and verification of the PG and MySQL cases
- Slowly add each mode of operation like objstore, kafka, and C2C

I don't want to embroil us in support tickets across the board in case something goes awry.

I'm wondering if it's possible for us to take this piecemeal approach -- starting with the PG and MySQL sources?

ZhouXing19:
> Left some comments here. My remaining thought is this: because this is such an all encompassing change, we risk regressions across all operating modes.
> 
> 
> 
> Once we continue this work, I don't think we should merge and release all of this at once. Here is how I'm thinking about the sequencing (pun intended):
> 
> - Make the core changes to the sequencer and relevant interfaces + implementations as their own standalone change
> 
> - Make the changes for PG and MySQL first since those will see the most immediate performance benefit
> 
> - Perform testing and verification of the PG and MySQL cases
> 
> - Slowly add each mode of operation like objstore, kafka, and C2C
> 
> 
> 
> I don't want to embroil us in support tickets across the board in case something goes awry.
> 
> 
> 
> I'm wondering if it's possible for us to take this piecemeal approach -- starting with the PG and MySQL sources?

That was the plan -- having smaller PRs to make smaller progress. The tricky thing is how the integration tests are written -- it's convoluted with the script wrapper. It means if we changed the pg/mysql front end, to make sure the CI can be green, we need to modify the interface of the script wrapper too. However, the script wrapper is shared by other frontends and thus other integrations too, meaning we need to change other frontends to accommodate the change too. We make break it into smaller commits, but given how the tests are structured, I don't think there's an easy way to break it into smaller PRs.

PR #1075, opened 2024-11-19
target: Support MySQL 5.7 as a Replication Target

### LINKED ISSUE #988: Support MySQL 5.X as a target
These older versions of MySQL don't support CTE's, which, at a minimum, prevents the schema-inspection queries from running. There may be other unknown-unknown's in terms of SQL workload changes.

noelcrl:
- Updates various Replicator SQL queries, query templates, code, and tests across Replicator to not use unsupported features in older MySQL target versions like 5.7. These unsupported features include:
  - common table expressions (CTEs).
  - VALUES syntax to create a temporary table.
  - expressions for default column values.
- Fixes issues caused by the following:
    - JSON YEAR columns displayed as [base64-encoded strings](https://bugs.mysql.com/bug.php?id=93253) instead of text.
    - MySQL geometry column functions complaining about the type of placeholder string arguments.
    - JSON string output produced by a test had inconsistent trailing zeroes or not for numbers, depending on the MySQL version, causing the test to fail
- Changes the default collation for Go-MySQL-Driver connections to older MySQL versions (< 8.0.1) to be the `@@collation_database` system variable, since older versions use latin1_swedish_ci as the default collation instead of the default for the driver (utf8mb4_general_ci).
  - Without it, certain Replicator queries fail due to illegal collation mixing.
- Adds integration tests to ensure that edge cases for the conditional upsert query templates are covered for all supported target dialects (except oracle for now).
- Fixes an issue with the .gitignore ignoring unintended directories.

**CI Changes:**
- Adds MySQL 5.7 as a migration backfill case for the testing matrix.

**Side note:** Externalizing the schemawatch queries with embed.FS and the apply package instead of using string constants should be addressed in a future PR. Refer to the resolved issue comments for more information.

Resolves #988. Jira ticket: [CC-29766](https://cockroachlabs.atlassian.net/browse/CC-29766)

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1075)


ryanluu12345:
You can probably rebase here now: https://github.com/cockroachdb/replicator/pull/1081

internal/target/apply/apply.go
```diff
@@ -48,6 +48,7 @@ type apply struct {
 	dlqs    types.DLQs
 	loader  *load.Loader
 	product types.Product
+	version string
```
ryanluu12345, line 51: It may be helpful to be more specific here: `DatabaseVersion` for example.

internal/target/apply/apply_test.go
```diff
@@ -654,7 +654,6 @@ var (
 			sourceType: `GEOGRAPHY`,
 			columnType: `GEOMETRY`,
 			sqlValue:   `0101000020E6100000000000000000F03F0000000000000040`,
-			expectJSON: `{"coordinates":[1.0,2.0],"type":"Point"}`,
```
ryanluu12345, line 657: Why is this `expectJSON` now not set? Doesn't this reduce our coverage? And shouldn't it lead to failures?

internal/target/apply/apply_test.go
```diff
@@ -1193,7 +1200,6 @@ func TestAllDataTypes(t *testing.T) {
 func normalizeJSON(data string) (string, error) {
 	var value any
 	dec := json.NewDecoder(strings.NewReader(data))
-	dec.UseNumber()
```
ryanluu12345, line 1196: Can you explain why this is not needed anymore?

internal/target/apply/column_mapping.go
```diff
@@ -46,6 +46,7 @@ type columnMapping struct {
 	Merger               merge.Merger                 // Conflict-resolution callback.
 	Positions            *ident.Map[positionalColumn] // Map of idents to column info and position.
 	Product              types.Product                // Target database product.
+	Version              string                       // Target database version. Useful to check for unsupported features.
```
ryanluu12345, line 49: Again, would be clearer if you can do `DatabaseVersion` so it's clearer.

internal/target/apply/queries/my/common.tmpl
```diff
@@ -8,7 +8,15 @@
     {{- end -}}
 {{- end -}}
 
-
+{{- /* Produce SELECT, UNION SELECT statements to create a temporary table containing data rows to upsert.
```
ryanluu12345, line 11: Thanks for the comment here. I'm wondering if you know if there are any downsides to creating this temporary table in terms of performance?

internal/target/apply/queries/my/conditional_compat.tmpl
```diff
@@ -0,0 +1,67 @@
+{{- /*gotype: github.com/cockroachdb/replicator/internal/target/apply.templates*/ -}}
+{{- /*
+This template implements the conditional upsert flow (compare-and-set,
```
ryanluu12345, line 3: And just to clarify, we're keeping the MySQL >= 8.0.1 code right? This is just the template it falls into if we are in an older MySQL version?

internal/target/apply/templates.go
```diff
@@ -206,12 +207,30 @@ func newTemplates(mapping *columnMapping) (*templates, error) {
...
 		ret.tmpl = tmplCRDB
 
-	case types.ProductMariaDB, types.ProductMySQL:
+	case types.ProductMariaDB:
 		ret.conditional = tmplMy.Lookup("conditional.tmpl")
 		ret.delete = tmplMy.Lookup("delete.tmpl")
 		ret.upsert = tmplMy.Lookup("upsert.tmpl")
 		ret.tmpl = tmplMy
 
+	case types.ProductMySQL:
```
ryanluu12345, line 216: On one hand, I like that you're being explicit about the differences between MySQL and MariaDB. On the other hand, I see that for the `else` case we just duplicate the first three lines of the MariaDB case. I think this is fine, since I prefer explicitness.

internal/target/apply/templates_conditional_test.go
```diff
@@ -0,0 +1,458 @@
...
+
+// Returns true if an expected upsert row for a test case matches the actual row
+// read from the target table.
+func expectedRowEqualsActual(expectedRow conditionalUpsertTestRow, actualRow []any) bool {
+	return expectedRow.pk0 == actualRow[0] &&
+		expectedRow.pk1 == actualRow[1] &&
+		expectedRow.ver0 == actualRow[2] &&
+		expectedRow.ver1 == actualRow[3] &&
+		expectedRow.hasDefault == actualRow[4]
+	// Skip ts comparison; we just need to know whether the other columns updated or not.
```
ryanluu12345, line 457: Nit: comment can go above the `return`

internal/target/dlq/dlq_schema.go
```diff
@@ -54,7 +54,7 @@ data_after JSONB NOT NULL,
 data_before JSONB NOT NULL
 )`
 	basicMySQLSchema = `CREATE TABLE %[1]s (
-event binary(16) DEFAULT (uuid()) PRIMARY KEY,
+event INT NOT NULL AUTO_INCREMENT PRIMARY KEY,
```
ryanluu12345, line 57: If I understand correctly, are we making this `event` an INT because older versions of MySQL do not support `binary` and `uuid` ?

My concern with this change is that it will be a breaking change for customers using the MySQL flow that already have DLQs created. I think we may need conditional logic here for the MySQL older versions case so we can comply with language features and not break existing users. Let's think on this more and discuss in more detail either here or with the team in standup.

internal/target/load/provider.go
```diff
@@ -47,7 +48,17 @@ func ProvideLoader(statements *types.TargetStatements, target *types.TargetPool)
...
 	case types.ProductMySQL:
-		l.selectTemplate = templates.Lookup("my.tmpl")
+		needsCompatQueries, err := stdpool.MySQLVersionNeedsCompatibilityQueries(target.Version)
+		if err != nil {
+			return nil, err
+		}
+
+		// MySQL version should at least support CTEs to use the my.tmpl query.
+		if needsCompatQueries {
+			l.selectTemplate = templates.Lookup("my_compat.tmpl")
```
ryanluu12345, line 58: Is there a reason the template file is named `my_compat` and not something like `my_legacy` or something that indicates it's an older version template? Again, really a nit here.

internal/target/schemawatch/coldata.go
```diff
@@ -363,10 +431,27 @@ func getColumns(
...
-			parts[1].Raw(),
+
+		needsCompatQueries, err := stdpool.MySQLVersionNeedsCompatibilityQueries(tx.Version)
+		if err != nil {
+			return nil, err
+		}
+
+		if needsCompatQueries {
+			// Support older versions of MySQL with unsupported query features (ex. CTEs).
+			stmt = sqlColumnsQueryMySQLCompatibility
```
ryanluu12345, line 442: I understand this is to support older features, but it's unclear why your args slice has 4 items in it and what each of these items mean. Can you please explain what each of the args represents?

Also, is it meant to be hardcoded to 4 items and 2 items respectively? If so why?

ryanluu12345: I'm guessing it's related to this?
```
// Parameters:
// - $1: Table catalog (always "def" in MySQL).
// - $2: Table schema ("my_database").
// - $3: Table catalog (same as $1).
// - $4: Table schema (same as $2).
```
I think it deserves a comment here.

internal/target/schemawatch/coldata_test.go
```diff
@@ -39,6 +40,12 @@ func TestGetColumns(t *testing.T) {
 
 	ctx := fixture.Context
 
+	// Only execute specific tests that support expressions in default column values.
+	var mySQLVersionSupportsExpressionsForDefaultColVals bool
+	if fixture.TargetPool.Product == types.ProductMySQL {
+		mySQLVersionSupportsExpressionsForDefaultColVals, _ = stdpool.MySQLVersionSupportsExpressionsForDefaultColVals(fixture.TargetPool.Version)
```
ryanluu12345, line 46: I'd recommend still returning the error and just using `r.NoError(err)`

internal/target/schemawatch/dependencies.go
```diff
@@ -256,6 +308,27 @@ func getDependencyRefs(
 		// Catalog names are hardcoded to "def" in MySQL.
 		args = []any{`def`, parts[0].Raw()}
 
+	case types.ProductMySQL:
+		parts := db.Idents(make([]ident.Ident, 0, 1))
+		if len(parts) != 1 {
+			return nil, errors.Errorf("expecting one schema parts, had %d", len(parts))
```
ryanluu12345, line 314: Super nit: `expecting one schema part` (singular)

internal/target/schemawatch/dependencies.go
```diff
@@ -256,6 +308,27 @@ func getDependencyRefs(
...
+
+		needsCompatQueries, err := stdpool.MySQLVersionNeedsCompatibilityQueries(tx.Version)
+		if err != nil {
+			return nil, err
+		}
+
+		if needsCompatQueries {
+			// Support older versions of MySQL with unsupported query features like CTEs.
+			stmt = depOrderTemplateMySQLCompatibility
+			args = []any{`def`, parts[0].Raw(), `def`, parts[0].Raw()}
```
ryanluu12345, line 325: Explaining why `def` is hardcoded here would be helpful. Can leave a short comment

internal/util/stdpool/my.go
```diff
@@ -146,29 +148,84 @@ func OpenMySQLAsTarget(
 
 	ping:
 		if err := ret.Ping(); err != nil {
-			// For some errors, we retry.
+			// For some ping errors, we retry, since some tests will wait longer for DB startup.
```
ryanluu12345, line 151: Thanks for the clarification comments. Really helpful!

internal/util/stdpool/my.go
```diff
@@ -146,29 +148,84 @@ func OpenMySQLAsTarget(
...
+			// Try a different tlsConfig.
 			continue
 		}
 		// Testing that connection is usable.
 		if err := ret.QueryRow("SELECT VERSION();").Scan(&ret.Version); err != nil {
 			return nil, errors.Wrap(err, "could not query version")
 		}
-		log.Infof("Version %s.", ret.Version)
+
+		var mySQLVersionUsesLatinDefaultCollation bool
```
ryanluu12345, line 171: I need to give this a deeper look when I have another moment to just sit and review after upcoming meetings.

ryanluu12345 (commented):
Great start on this! I want to give the MY connection logic + collations updates and the actual queries a closer look. 

I'd recommend having @ZhouXing19 take a look at the queries since she has expertise on the schema introspection, given she implemented a lot of that for `molt`.

internal/target/apply/apply_test.go
noelcrl: `expectJSON` is being used an "expected value override". If it is set, we use this string as the expected value instead of the mutation payload value `jsonValue` that we construct during the test using the staging database and `sqlValue`.

In this test case (and for MySQL 8 specifically), the `expectJSON` override was being used to account for trailing zero differences when we read back the resulting mutation from the target. For example, the mutation being applied to the target is: `{"coordinates":[1,2],"type":"Point"}`. When we read it back from the target table however, it is `{"coordinates":[1.0,2.0],"type":"Point"}`. As JSON objects, these should be equal. But if we compare them as strings, they are unequal.

`normalizeJSON()` is supposed to re-encode these JSON strings so string formatting differences shouldn't affect the equality test. However, since the function used .Number(), it would actually keep the formatting of the numbers as is, so `1` in `jsonValue` would stay as `1`, and `1.0` in the actual value would stay as `1.0`, so the test case would fail.

Because of this, the original test case used the trailing zero `expectJSON` override value as a workaround. After changing the `normalizeJSON()` function, there was no longer a need for the `expectJSON` field as an override, since trailing zero differences in JSON strings should no longer matter for JSON object equality.

**However**, after testing normalizeJSON() a bit more (and the golang json decoding), it seems to convert numbers to float64 by default during decoding. This causes `bigint` numbers to lose some information when re-encoded, and we don't want to change the actual data during the test. Using `UseNumber()` decodes them as strings instead, so it will keep the precision.

Because of this new issue, I'm thinking about keeping the `UseNumber()` like before, but finding another way to remove those trailing zeros. If we were to use the `expectJSON` workaround like before, we would have to have separate logic in the test for older versions of MySQL like 5.7, since those are read back without the trailing zero, unlike MySQL 8.0. Any thoughts on how I should approach this issue? Thank you!

internal/target/apply/apply_test.go
noelcrl: Please see my reply to your comment above, as it explains there.

internal/target/apply/queries/my/conditional_compat.tmpl
noelcrl: Yes. The compatibility queries will only be used for MySQL versions < 8.0.1, which don't support CTEs.

internal/target/apply/queries/my/common.tmpl
noelcrl: Not entirely sure, but this template is used so we can just use one INSERT statement and SELECT FROM this temporary table instead of using multiple insert statements. I think MySQL does also support multiple value insert, but this may change the overall query structure, as we refer back to this temporary table when joining and applying filtering logic in the query to determine which rows to upsert.

internal/target/dlq/dlq_schema.go
noelcrl: I see. It seems like this schema is logged to the user so it gives them a reference on what their DLQ schema should look similar to if they don't have a DLQ table created already. It essentially uses it to create a schema only for testing, so the user can make the PK column whatever they want since it is not in the minimum set of expected columns `expectedColumns`. However, it is a good point that even logging an example table schema for the user to create themselves should be different based on the versions, based on what is recommended and supported (uuids recommended in supported versions).

I'll add some conditional logic to change the logged/testing schema based on version here.

internal/target/load/provider.go
noelcrl: I think the "compat" naming reasoning is that the query is a compatibility version that is compatible with older versions (also some Android development bias haha). What should we prefer here?

internal/target/schemawatch/coldata.go
noelcrl: Yes, due to the structure of the new compatibility query, it needs to pass in the same placeholder args twice, as I could not find a way to reuse the same argument for different placeholder strings in the query. I will add a clarification comment.

internal/target/apply/apply_test.go
ryanluu12345: Thanks for the thorough explanation here. I'm curious if this is only a test issue. Do we have cases in the actual code where we use this methodology? I want to make sure we're not exposed to cases where bigint numbers lose information

ryanluu12345: How simple is it to determine if it's MySQL 5.7 in the test case and just do conditional logic? I think we can approach it that way. And then put a comment on why the logic is different.

What were you thinking for another way to remove trailing zeroes?

internal/target/apply/queries/my/common.tmpl
ryanluu12345: Ok got it. If I understand this correctly, this temporary table solution is the way we approach the other MySQL cases already?

Let's opt for consistency with the previous implementation (as much as possible). Can evaluate other options if we end up seeing perf be an issue here.

internal/target/apply/queries/my/conditional_compat.tmpl
ryanluu12345: Ok great, thanks for clarifying. That bit of logic looks good to me.

internal/target/apply/queries/my/common.tmpl
```diff
@@ -31,6 +39,36 @@
     {{- end -}}
 {{- end -}}
 
+{{- /* Similar as cas-exprs, but also adds column names to the first row of the
```
ryanluu12345, line 42: Since this seems relevant for the 5.7 case, why are we putting this in `common.tmp`? Wouldn't this also affect > mysql 8.0?

internal/target/dlq/dlq_schema.go
ryanluu12345: Ok, so just to clarify a few things:
1. This is only logged out as an example to follow but is not actually applied for the user on their MySQL or CRDB cluster?
2. It is used for testing purposes to create a testing DLQ?

Yes, I think the right move here is to do conditional logic so that we log out the best suggestion we can give per dialect + version. I don't think it makes sense to give a one-size fits all solution and cater to the non-ideal case (using an INT), when UUID is a better option, especially for large volumes of data where the DLQ INT PK could be a limitation.

internal/target/load/provider.go
ryanluu12345: Let's just keep compat. I understand what it means. Maybe @ZhouXing19 has a better idea of naming here. Again, just a nit

internal/target/schemawatch/coldata.go
ryanluu12345: ok understood. Yeah, a comment would help here just to clarify.

internal/util/stdpool/my.go
```diff
@@ -146,29 +148,84 @@ func OpenMySQLAsTarget(
...
+			if err != nil {
+				return nil, errors.Wrap(err, "could not parse MySQL version")
+			}
 		}
+
+		if mySQLVersionUsesLatinDefaultCollation {
+			// Using an older version of MySQL, be sure to use the same collation for
+			// driver connections to avoid illegal collation mixing errors with queries.
+			var collation string
+			err = ret.QueryRow("SELECT @@collation_server").Scan(&collation)
```
ryanluu12345, line 189: Are there any particular permissions a user needs to be able to have in order to run this `SELECT @@collation_server` ?

internal/util/stdpool/my.go
```diff
@@ -146,29 +148,84 @@ func OpenMySQLAsTarget(
...
 		if err := ret.QueryRow("SELECT VERSION();").Scan(&ret.Version); err != nil {
 			return nil, errors.Wrap(err, "could not query version")
 		}
-		log.Infof("Version %s.", ret.Version)
+
+		var mySQLVersionUsesLatinDefaultCollation bool
 		if strings.Contains(ret.Version, "MariaDB") {
 			ret.PoolInfo.Product = types.ProductMariaDB
+		} else {
+			// Older MySQL versions (< 8.0.1) may use latin1_swedish_ci as the
```
ryanluu12345, line 175: What happens if it is MySQL version >= 8.0.1 and it is using the latin default collation? Do we still need to do the below collation driver connection logic?

It's clear to me that for < 8.0.1 and latin, we want to use the same collation for the driver connections though.

internal/util/stdpool/my.go
```diff
@@ -146,29 +148,84 @@ func OpenMySQLAsTarget(
...
+			// driver connections to avoid illegal collation mixing errors with queries.
+			var collation string
+			err = ret.QueryRow("SELECT @@collation_server").Scan(&collation)
+			if err != nil || collation == "" {
+				return nil, errors.Wrap(err, "could not query collation_server")
+			}
+			cfg.Collation = collation
+
+			// Re-open the DB to use the new collation config option.
+			_ = ret.Close()
```
ryanluu12345, line 196: So if I understand this correctly, we close the old pool and then we create a new pool with the new latin collation to make sure they're consistent collations?

internal/target/apply/apply_test.go
noelcrl: > Thanks for the thorough explanation here. I'm curious if this is only a test issue. Do we have cases in the actual code where we use this methodology? I want to make sure we're not exposed to cases where bigint numbers lose information

Good point. I checked the occurrences in the non-test Replicator code where `.UseNumber()` is not called after creating a new JSON decoder. In these cases, a struct with a predefined type is used as the destination for the decoded JSON. The fields in these structs have either a `string` or `[]byte` data type, so they shouldn't lose precision if they are numbers. All other JSON decoder creations call `.UseNumber()` afterwards.

noelcrl: > How simple is it to determine if it's MySQL 5.7 in the test case and just do conditional logic? I think we can approach it that way. And then put a comment on why the logic is different.
> 
> What were you thinking for another way to remove trailing zeroes?

I would have to modify the testing code and add a version check after determining it's MySQL, then either add in the test case with the expectedJSON for each version.

Right now, I made the normalizeJSON() handle JSON objects with number properties by recursively going through the JSON object and stripping trailing zeros from any encountered property that is a json.Number. I made it extensible by allowing formatting for other data types by adding another switch case. I like this approach since it makes `normalizeJSON()` more likely to accomplish what it intends, and we don't have to add some workarounds based on the DB version for trailing zeros.

internal/target/dlq/dlq_schema.go
noelcrl: > Ok, so just to clarify a few things:
> 
> 1. This is only logged out as an example to follow but is not actually applied for the user on their MySQL or CRDB cluster?
> 2. It is used for testing purposes to create a testing DLQ?
> 
> Yes, I think the right move here is to do conditional logic so that we log out the best suggestion we can give per dialect + version. I don't think it makes sense to give a one-size fits all solution and cater to the non-ideal case (using an INT), when UUID is a better option, especially for large volumes of data where the DLQ INT PK could be a limitation.

1. Yes. If Replicator detects that the user has not created a DLQ table yet, it will log a message which contains an example reference schema that they can follow to create their own. Replicator pushes the DLQ table creation responsibility onto the user to allow them to be more custom/extensible with their DLQ table and workloads.
2. Yes. The example schema that is logged out to the user is also used to create a DLQ table during tests only.

Agreed. It is worth noting that UUIDs **are** supported by MySQL 5.7, but expressions for default column values aren't, so UUID() can't be used as a default value for the `event` column. Theoretically, we could use UUIDs for older versions, but we would have to explicitly include the PK column ('event' in this case) in the insert statement like this:
```
INSERT INTO dlq_table (event, ...) VALUES (UNHEX(REPLACE(uuid(), "-","")), ...);
```

1. We would either have to add 'event' into the require dlq table schema columns (potential breaking changes, since it wasn't required before) or add extra logic to check for its existence before including it specifically in the query. Do you think it is worth all these changes/effort to support UUIDs for MySQL 5.7? Or should we just use BIGINT or INT as the recommended PK type for 5.7?

2. Also something strange I noticed, typically just using a `uuid()` through the MySQL shell to store into a table with a `binary(16)` causes an error since a 36 character uuid() string is too long for binary(16). But if it is inserted via the golang mysql driver, there is no error, even without converting it to binary with UUID_TO_BIN(). Is the driver somehow converting the uuid() string in the query to binary before storing it? Just wondering why the same error doesn't occur via the go driver.

internal/target/apply/queries/my/common.tmpl
noelcrl: It appears that `common.tmpl` is used as file which contains helper templates that can be used across different `my/X.tmpl` files. Since `cas-exprs` is only used by `my/conditional.tmpl`, and `cas-exprs-compat` is only used by `my/conditional_compat.tmpl`, we could just put those helper templates in the same conditional template file that references them, or put them in their own file which contains helper templates for each `my/conditionalX.tmpl`. Which option do we prefer here?

It won't affect the functionality for MySQL 8.0, since the conditional upsert query template file used for those versions (`my/conditional.tmpl`) only uses `cas-exprs` and **not** `cas-exprs-compat`.

internal/util/stdpool/my.go
noelcrl: Yes. Every new connection that is opened for the pool will now have the consistent collation that was read from the `collation_server` variable.

internal/util/stdpool/my.go
noelcrl: I tested the offending query using MySQL v8.0.34 and set the global default collation variables to latin and it seemed to work fine despite the difference. We could also just update the collation for each new driver database connection to be whatever the default is on the server, regardless of the MySQL version to be safe? What should we prefer here?

internal/target/load/provider.go
ZhouXing19: compat lgtm

internal/target/apply/apply_test.go
ryanluu12345: Ok got it. So sounds like this is DB version agnostic now, which means even if in the future we add a new database version that has different behavior, the trailing zeroes should still be stripped right?

internal/target/apply/queries/my/common.tmpl
```diff
@@ -8,14 +8,14 @@
     {{- end -}}
 {{- end -}}
 
-
-{{- define "cas-exprs" -}}
+{{- define "exprs" -}}
```
ryanluu12345, line 11: Hmm, don't we use `cas-exprts` for the other database versions? Is this renamed also where it's referenced elsewhere? Just make sure we don't regress the MySQL 8.0 and up versions.

internal/target/apply/queries/my/delete_compat.tmpl
```diff
@@ -0,0 +1,11 @@
+{{- /*gotype: github.com/cockroachdb/replicator/internal/target/apply.templates*/ -}}
+{{- /*
+DELETE FROM "database"."schema"."table"
```
ryanluu12345, line 3: Could you do a quick search on if there is a behavior difference between:
1. `DELETE FROM ... WHERE ... IN ()`
2. `DELETE FROM ... WHERE condition1 OR condition2 OR condition3`

internal/target/dlq/dlq_schema.go
```diff
@@ -60,6 +60,16 @@ source_nanos INT8 NOT NULL,
 source_logical INT8 NOT NULL,
 data_after JSON NOT NULL,
 data_before JSON NOT NULL
+)`
+	// Suggested schema to use for the older MySQL version 5.7
+	// that doesn't support expressions for default column values
+	BasicMySQL5_7Schema = `CREATE TABLE %[1]s (
```
ryanluu12345, line 66: It may be worth leaving an explicit comment for the non 5_7 schemas so it's clear for 8.0 and up you recommend to use UUID. But yeah, makes sense for us to use BIGINT here. You may even want to change this to a `BIGINT UNSIGNED` since the range is larger on the positive side: 2^64 - 1.

internal/util/stdpool/my.go
ryanluu12345: I think what's appropriate here is to do the following:
1. Leave a comment explaining that we have this option to check the DB connection and see what the collation is and then making sure we're consistent. Please create a ticket here to track our re-evaluation of this behavior: https://cockroachlabs.atlassian.net/browse/CC-30175
2. Keep this behavior as is. I want to limit the amount of aspects we change here. But now we know that if we run into collation sync issues, there's a path forward :)

internal/util/stdpool/my.go
ryanluu12345: Excellent, no concerns :)

internal/target/apply/apply_test.go
noelcrl: Yes, and yes. No matter if the JSON string we get back from the DB has trailing zeroes for its numbers or not, we should now expect trailing zero numbers vs. non-trailing zero numbers to be equal in the test.

internal/target/apply/queries/my/common.tmpl
noelcrl: Oh yes, `cas-exprs` is used for MySQL 8.0+. I moved it into the `conditional.tmpl` file (which is used by MySQL 8.0+), since it is only referenced in that template file, and is not really a common helper template used amongst multiple files. I also moved `cas-exprs-compat` into the `conditional_compat.tmpl` file for the same reason. The `cas-exprs` helper template is still being used, it was just moved into a different file essentially, so MySQL 8.0+ should have no regressions. `exprs` is actually a different template altogether from `cas-exprs`, which is a more specialized template for the `conditional.tmpl` file, so there is no renaming going on here.

internal/target/apply/queries/my/common.tmpl
```diff
@@ -8,14 +8,14 @@
...
     {{- range $groupIdx, $pairs := $.Vars -}}
-        {{- if $groupIdx }}{{- nl }}  UNION SELECT {{ end -}}
```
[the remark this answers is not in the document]
noelcrl: The `exprs` helper template is actually unchanged from previous commits, I believe these diffs may be caused by moving the `cas-exprs` helper template into the `conditional.tmpl` file, and how similar the `cas-exprs` definition is to `exprs`. The difference between `exprs` and `cas-exprs` is that `exprs` is used to construct the data rows in queries like the following INSERT:

```
INSERT INTO "schema"."table"
("pk0","pk1","val0","val1","has_default")
VALUES
(?,?,?,?,CASE WHEN ? THEN ? ELSE expr() END),
(?,?,?,?,CASE WHEN ? THEN ? ELSE expr() END)
ON DUPLICATE KEY UPDATE
"val0"=VALUES("val0"),"val1"=VALUES("val1"),"has_default"=VALUES("has_default")
```
The `exprs` helper template is used in this case to generate the following data row placeholders in the above query.
```
(?,?,?,?,CASE WHEN ? THEN ? ELSE expr() END),
(?,?,?,?,CASE WHEN ? THEN ? ELSE expr() END)
```
The `cas-exprs` template on the other hand is used to construct the data part of queries like this:
```
INSERT
INTO "schema"."table"("pk0","pk1","val0","val1","has_default")
WITH data  ("pk0","pk1","val0","val1","has_default") AS (
  SELECT ?,?,?,?,CASE WHEN ? = 1 THEN ? ELSE expr() END
  UNION SELECT ?,?,?,?,CASE WHEN ? = 1 THEN ? ELSE expr() END
),
current AS (
SELECT "pk0","pk1", "table"."val1","table"."val0"
FROM "schema"."table"
JOIN data
USING ("pk0","pk1")),
action AS (
SELECT data.* FROM data
LEFT JOIN current
USING ("pk0","pk1")
WHERE current."pk0" IS NULL OR
(data."val1",data."val0") > (current."val1",current."val0"))
SELECT * FROM action
ON DUPLICATE KEY UPDATE
"val0"=VALUES("val0"),"val1"=VALUES("val1"),"has_default"=VALUES("has_default")
```
Where the following data rows in the above query are generated by `cas-exprs`:
```
SELECT ?,?,?,?,CASE WHEN ? = 1 THEN ? ELSE expr() END
UNION SELECT ?,?,?,?,CASE WHEN ? = 1 THEN ? ELSE expr() END
```
Notice how `exprs` doesn't need to use UNION SELECT since it is used in the context of a multi-row insert statement, and how `cas-exprs` does need it since it is constructing a CTE that will be referenced later in the query with conditional logic based on whether the row versions are greater or not.

Overall, both `cas-exprs` and `exprs` should be unchanged (so no regressions from before), `cas-exprs` was just moved into the `conditional.tmpl` file.

internal/target/apply/queries/my/delete_compat.tmpl
noelcrl: There wasn't any behavior difference I could find when searching, but it seems that the `IN` syntax may be a bit more concise of course. I also did some manual testing on 5.7 to make sure:

```
mysql> create table t1 (pk0 int, pk1 int, PRIMARY KEY(pk0,pk1));
Query OK, 0 rows affected (0.02 sec)

mysql> insert into t1 values (0, 0);
Query OK, 1 row affected (0.03 sec)

mysql> insert into t1 values (0, 1);
Query OK, 1 row affected (0.00 sec)

mysql> insert into t1 values (1, 0);
Query OK, 1 row affected (0.01 sec)

mysql> insert into t1 values (1, 1);
Query OK, 1 row affected (0.00 sec)

mysql> delete from t1 where (pk0, pk1) in ((0, 0), (0, 1));
Query OK, 2 rows affected (0.01 sec)

mysql> select * from t1;
+-----+-----+
| pk0 | pk1 |
+-----+-----+
|   1 |   0 |
|   1 |   1 |
+-----+-----+
2 rows in set (0.00 sec)

mysql> delete from t1 where (pk0 = 1 AND pk1 = 0) OR (pk0 = 1 AND pk1 = 1);
Query OK, 2 rows affected (0.01 sec)

mysql> select * from t1;
Empty set (0.00 sec)
```

internal/target/apply/queries/my/conditional_compat.tmpl
```diff
@@ -0,0 +1,99 @@
...
+      SELECT ? AS "pk0",? AS "pk1",? AS "val0",? AS "val1",CASE WHEN ? = 1 THEN ? ELSE expr() END AS "has_default"
+      UNION SELECT ?,?,?,?,CASE WHEN ? = 1 THEN ? ELSE expr() END
+    ) AS data
+(3) LEFT JOIN (SELECT "pk0","pk1", "table"."val1","table"."val0" FROM "schema"."table") AS current ON ("pk0","pk1")
+(4) WHERE ("val0"> now() - INTERVAL '3600' SECOND)
+    AND ("val1"> now() - INTERVAL '1' SECOND)
+(5) AND (current."pk0" IS NULL OR
+        ((data."val1",data."val0") > (current."val1",current."val0")))
+(6) ON DUPLICATE KEY UPDATE
+    "val0"=VALUES("val0"),"val1"=VALUES("val1"),"has_default"=VALUES("has_default")
```
ZhouXing19, line 21: super nit: might be helpful to comment that `VALUES(col_name) in the UPDATE clause refers to the value of col_name that would be inserted, had no duplicate-key conflict occurred., See also https://dev.mysql.com/doc/refman/8.4/en/miscellaneous-functions.html#function_values`

ZhouXing19 (approved):
Stamping for the tmpl files. The comprehensive comments inline is super helpful, excellent job!

internal/target/apply/queries/my/common.tmpl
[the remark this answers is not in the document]
ryanluu12345: Thanks for the detailed reply. I see what you mean. I see that it's just been moved but it's largely unchanged, with the exception of clarification comments (which I love btw).

internal/target/apply/queries/my/common.tmpl
```diff
@@ -43,7 +44,7 @@
             {{- if $pair.Expr -}}
                 ({{ $pair.Expr }})
             {{- else if eq $pair.Column.Type "geometry" -}}
-                st_geomfromgeojson(?)
```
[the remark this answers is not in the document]
ryanluu12345: Ok understood, so in order to make it work for all versions, we just `CAST ... AS JSON` which works on both versions and guarantees we are working with JSON under the hood right?

internal/target/apply/queries/my/delete_compat.tmpl
ryanluu12345: Ah sorry, I mistyped above, I mean "performance impact" not "behavior"

ryanluu12345 (approved):

ryanluu12345 (approved):

internal/target/apply/apply_test.go
```diff
@@ -1197,10 +1204,45 @@ func normalizeJSON(data string) (string, error) {
...
 	}
+	processJSON(value)
 	buf, err := json.Marshal(value)
 	return string(buf), err
 }
 
+// Takes a JSON object and recusively processes each
+// property to ensure standardardized formatting for different data types.
+// For instance, ensuring that all number properties don't have trailing zeros.
+func processJSON(obj interface{}) interface{} {
```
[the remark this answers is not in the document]
ryanluu12345: LGTM!

internal/target/apply/queries/my/delete_compat.tmpl
noelcrl: Oh wow, that is a interesting optimization for larger clauses.

sravotto (approved):
:lgtm: - Just a couple of nits re: alphabetization in structs (in the past we tried to keep fields in alpha order, if possible).
Also, in a few instances, I was wondering if we can reduce the size of code by using  `compat` queries/templates also for newer version of MySQL.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1075#-:-ODD31uF0uh2x-sGZyJW:bv1swa6)* status: 0 of 39 files reviewed, 34 unresolved discussions (waiting on @MattWhelan, @noelcrl, and @ryanluu12345)

___
*[`internal/target/apply/queries/my/common.tmpl` line 47 at r4](https://reviewable.io/reviews/cockroachdb/replicator/1075#-ODD-vKyBSD0lIplEGNs:-ODD-vKyBSD0lIplEGNt:b4q8p6y) ([raw file](https://github.com/cockroachdb/replicator/blob/7783db3d7ad7247478d8d70fb11621e95bb367ff/internal/target/apply/queries/my/common.tmpl#L47)):*
> ```
>                 ({{ $pair.Expr }})
>             {{- else if eq $pair.Column.Type "geometry" -}}
>                 st_geomfromgeojson(CAST(? as JSON))
> ```

Does this work on MySQL 8? If so, you might not need two different versions, if I'm reading this correctly.
If it doesn't work, then I'd just add a comment to explain why we have 2 versions.
___
*[`internal/target/schemawatch/coldata.go` line 212 at r4](https://reviewable.io/reviews/cockroachdb/replicator/1075#-ODD4kys38rXeHwS5rKi:-ODD4kytA2Cc5Q_rkm_c:b-jdd8eq) ([raw file](https://github.com/cockroachdb/replicator/blob/7783db3d7ad7247478d8d70fb11621e95bb367ff/internal/target/schemawatch/coldata.go#L212)):*
> ```Go
> ORDER BY ordered.ordinal_position, column_name
> `
> ```

Does this query work on MySQL 8? If it does, perhaps we can just use this.
___
*[`internal/target/apply/column_mapping.go` line 49 at r3](https://reviewable.io/reviews/cockroachdb/replicator/1075#-ODCjysM-7MXgQr_nFEr:-ODCjysM-7MXgQr_nFEs:b-bjs2op) ([raw file](https://github.com/cockroachdb/replicator/blob/64a20d40b0bee8d03dd1aa21a8611195fa22b5dc/internal/target/apply/column_mapping.go#L49)):*
> ```Go
> 	Positions            *ident.Map[positionalColumn] // Map of idents to column info and position.
> 	Product              types.Product                // Target database product.
> 	DatabaseVersion      string                       // Target database version. Useful to check for unsupported features.
> ```

nit: alphabetize.
___
*[`internal/target/apply/apply.go` line 51 at r3](https://reviewable.io/reviews/cockroachdb/replicator/1075#-ODCjWuz4yuWsU_Z8ioN:-ODCjWuz4yuWsU_Z8ioO:b-bjs2op) ([raw file](https://github.com/cockroachdb/replicator/blob/64a20d40b0bee8d03dd1aa21a8611195fa22b5dc/internal/target/apply/apply.go#L51)):*
> ```Go
> 	loader          *load.Loader
> 	product         types.Product
> 	databaseVersion string
> ```

nit: alphabetize.
___
*[`internal/target/apply/templates_conditional_test.go` line 32 at r4](https://reviewable.io/reviews/cockroachdb/replicator/1075#-ODD2pqQ2dvVcYl2G6Zg:-ODD2pqQ2dvVcYl2G6Zh:b-sx2xfw) ([raw file](https://github.com/cockroachdb/replicator/blob/7783db3d7ad7247478d8d70fb11621e95bb367ff/internal/target/apply/templates_conditional_test.go#L32)):*
> ```Go
> 	"github.com/stretchr/testify/require"
> )
> ```

Nice set of new tests. Very useful.
___
*[`internal/target/schemawatch/dependencies.go` line 95 at r4](https://reviewable.io/reviews/cockroachdb/replicator/1075#-ODD1ixX9hXEO-6MxCvP:-ODD1ixX9hXEO-6MxCvQ:b-k6l3t8) ([raw file](https://github.com/cockroachdb/replicator/blob/7783db3d7ad7247478d8d70fb11621e95bb367ff/internal/target/schemawatch/dependencies.go#L95)):*
> ```Go
> SELECT * FROM refs
> `
> ```

Does this query work on MySQL 8 and MariaDB? If it does, perhaps we can just use this for all of them.

internal/target/apply/queries/my/common.tmpl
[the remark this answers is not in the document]
noelcrl: Yes. If there are no issues with both old and new versions with using `CAST ... AS JSON`, we could use it for both and just have one `exprs` helper template that can be used for old and new versions without needing `exprs-compat`.

noelcrl (commented):
Thank you! As a follow up for @ryanluu12345 , What do you think about using the same `compat` queries for both new and older DB versions? Should we keep them separate for now and use one later on?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1075#-:-ODIpRG216edkOCWHYnX:bk6xfjt)* status: 0 of 39 files reviewed, 32 unresolved discussions (waiting on @MattWhelan, @ryanluu12345, and @sravotto)

___
*[`internal/target/apply/apply.go` line 51 at r3](https://reviewable.io/reviews/cockroachdb/replicator/1075#-ODCjWuz4yuWsU_Z8ioN:-ODIpLgJDiZUkTHGMd_n:bhy912y) ([raw file](https://github.com/cockroachdb/replicator/blob/64a20d40b0bee8d03dd1aa21a8611195fa22b5dc/internal/target/apply/apply.go#L51)):*

nit: alphabetize.
</blockquote></details>

Will Do.
___
*[`internal/target/apply/column_mapping.go` line 49 at r3](https://reviewable.io/reviews/cockroachdb/replicator/1075#-ODCjysM-7MXgQr_nFEr:-ODIpGpw6470NUkNv5lq:bhy9ot6) ([raw file](https://github.com/cockroachdb/replicator/blob/64a20d40b0bee8d03dd1aa21a8611195fa22b5dc/internal/target/apply/column_mapping.go#L49)):*

nit: alphabetize.
</blockquote></details>

Will do.
___
*[`internal/target/apply/queries/my/common.tmpl` line 47 at r4](https://reviewable.io/reviews/cockroachdb/replicator/1075#-ODD-vKyBSD0lIplEGNs:-ODIoZdg3Y2BJf1lAlOB:b-l7o1cb) ([raw file](https://github.com/cockroachdb/replicator/blob/7783db3d7ad7247478d8d70fb11621e95bb367ff/internal/target/apply/queries/my/common.tmpl#L47)):*

Does this work on MySQL 8? If so, you might not need two different versions, if I'm reading this correctly.
If it doesn't work, then I'd just add a comment to explain why we have 2 versions.
</blockquote></details>

I'll test it more thoroughly with both. If it does, I can just have `exprs`.
___
*[`internal/target/schemawatch/coldata.go` line 212 at r4](https://reviewable.io/reviews/cockroachdb/replicator/1075#-ODD4kys38rXeHwS5rKi:-ODImFd24pslxde_sxTS:b-kukkhj) ([raw file](https://github.com/cockroachdb/replicator/blob/7783db3d7ad7247478d8d70fb11621e95bb367ff/internal/target/schemawatch/coldata.go#L212)):*

Does this query work on MySQL 8? If it does, perhaps we can just use this.
</blockquote></details>

It should, but please refer to my other reply to your comment for the new compatibility dependencies query.
___
*[`internal/target/schemawatch/dependencies.go` line 95 at r4](https://reviewable.io/reviews/cockroachdb/replicator/1075#-ODD1ixX9hXEO-6MxCvP:-ODIki-jEt5NcTju_xj6:bbehgfq) ([raw file](https://github.com/cockroachdb/replicator/blob/7783db3d7ad7247478d8d70fb11621e95bb367ff/internal/target/schemawatch/dependencies.go#L95)):*

Does this query work on MySQL 8 and MariaDB? If it does, perhaps we can just use this for all of them.
</blockquote></details>

It should theoretically, will wait for @ryanluu12345 on their thoughts about this potential change to use the same query.

ryanluu12345 (commented):
I think it's reasonable for us to use separate the compat queries for now since I would like to reduce the scope of this change and because this change is fairly large already. I'd like the existing MySQL 8 path to use the existing queries so that there is no net change there and less risk of regression.

I'm open to consolidating this in a later change. Let's track this in a ticket in the replicator epic: CC-30175. When we do this change, let's assess the query plan for these queries and see if there is a performance hit we are taking by doing the new compat queries for MySQL 8.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1075#-:-ODMmCfDAkMZC8XRULIF:bl46nvs)* status: 0 of 39 files reviewed, 32 unresolved discussions (waiting on @MattWhelan, @noelcrl, and @sravotto)

noelcrl (commented):
> Reviewed 13 of 28 files at r1, 7 of 17 files at r2, 14 of 16 files at r3, 2 of 2 files at r4, 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1075)* status: all files reviewed, 32 unresolved discussions (waiting on @MattWhelan, @ryanluu12345, and @sravotto)

internal/target/apply/apply.go
noelcrl: Done.

internal/target/apply/column_mapping.go
noelcrl: Done.

internal/target/apply/queries/my/conditional_compat.tmpl
noelcrl: Done.

internal/target/dlq/dlq_schema.go
noelcrl: Done.

internal/target/schemawatch/coldata_test.go
noelcrl: Done.

internal/target/schemawatch/dependencies.go
noelcrl: Done.

internal/target/schemawatch/dependencies.go
noelcrl: Done.

internal/util/stdpool/my.go
noelcrl: Done.

noelcrl (commented):
> Dismissed @ZhouXing19 and @ryanluu12345 from 27 discussions.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1075)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @MattWhelan)

noelcrl (commented):
> Reviewed 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1075)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @MattWhelan)

PR #1079, opened 2024-11-26
util: internalize stopvar lib and add WaitForUpdate

ZhouXing19:
This commit internalizes the `stopvar` lib which we used to import from the field-eng-power-tool repo. We also added the `WaitForUpdate` functionality, which differs from the existing `WaitForChange` that it will return regardless of the new value is different or not with the original value.

This is needed for the pull-based sequencer model, so that the main frontend can be notified even when a batch cursor is successfully processed by the backend without an error.

Release note: None


internal/util/stopvar/stopvar.go
```diff
@@ -0,0 +1,236 @@
...
+func DoWhenChanged[T comparable](
+	ctx *stopper.Context,
+	start T,
+	source *notify.Var[T],
+	fn func(ctx *stopper.Context, old, new T) error,
+) (last T, err error) {
+	last = start
+	for {
+		next, _ := WaitForChange(ctx, last, source)
+		// fmt.Printf("[DoWhenChanged]: [%v -> %v]]\n", last, next)
```
ryanluu12345, line 45: Please remove dead code

internal/util/stopvar/stopvar.go
```diff
@@ -0,0 +1,236 @@
...
+			return last, nil
+		}
+		if err := fn(ctx, last, next); err != nil {
+			return last, fmt.Errorf("changed [%v -> %v]: %w", last, next, err)
+		}
+		last = next
+	}
+}
+
+// DoWhenUpdated executes the callbackwhen an update attempt to the
```
ryanluu12345, line 56: Nit: callback when

internal/util/stopvar/stopvar.go
```diff
@@ -0,0 +1,236 @@
+// Copyright 2024 The Cockroach Authors
```
ryanluu12345, line 1: I'm trying to understand the context here: this just aims to be a utility wrapper around the `stopper` shared package? And is this something we need in order to simplify our usage of `stop vars` ?

It would be helpful to understand the context in which these helpers are used.

internal/util/stopvar/stopvar_test.go
```diff
@@ -0,0 +1,91 @@
...
+			ctx := stopper.Background()
+			defer ctx.Stop(0)
+			r := require.New(t)
+
+			toWatch := notify.VarOf[error](error(nil))
+			r.NotNil(toWatch)
+
+			ctx.Go(func(ctx *stopper.Context) error {
+				for i := 0; i < delaySecondsToUpdate; i++ {
+					t.Logf("sleeping for %dth seconds before updating", i+1)
```
ryanluu12345, line 68: This reads a bit odd. In the case of i = 0 -> "sleeping for 1th seconds before updating"

ryanluu12345 (commented):
Deferring to field engineering folks since they have more context on the `stopvar` related aspects here.

internal/util/stopvar/stopvar.go
ZhouXing19: We are extending the stopvar functionalities.
Currently what supported is to wait if a var has value changed. We are extending it by adding a new function set where we wait for an update attempt, even the value is unchanged. In other words, this is only function addition, without modifying any of existing functionalities.
For more context, this addition is needed for the pull-based model, which will be corporates in the oracle PR.

internal/util/stopvar/stopvar_test.go
ZhouXing19: updated to `have slept %d seconds`

noelcrl (commented):
Looks good so far! I left a few comments.

internal/util/stopvar/stopvar_test.go
```diff
@@ -0,0 +1,91 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+// Package script provides an API shim that integrates the userscript
```
noelcrl, line 17: Is this comment still relevant to `stopvar` here?

internal/util/stopvar/stopvar_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+	"github.com/cockroachdb/field-eng-powertools/notify"
+	"github.com/cockroachdb/field-eng-powertools/stopper"
+	"github.com/stretchr/testify/require"
+)
+
+// TestDoWhenUpdated tests the DoWhenUpdatedOrInterval function, which
+// should return if there is an update *attempt* made to the monitored
+// notify.Var, even when the actual notify.Var.mu.data is not changed.
+func TestDoWhenUpdated(t *testing.T) {
```
noelcrl, line 37: Should we also include a `TestDoWhenChangedOrInterval` test to add a bit of verification to the "changed" category of functions?

Jeremyyang920:
Does it make sense to update the power tools repo instead and vendor the change in?

internal/util/stopvar/stopvar.go
```diff
@@ -0,0 +1,235 @@
...
+// successfully-processed value will be returned.
+func DoWhenUpdatedOrInterval[T comparable](
+	ctx *stopper.Context,
+	start T,
+	source *notify.Var[T],
+	period time.Duration,
+	loggerForTest *strings.Builder,
+	fn func(ctx *stopper.Context, old, new T) error,
+) (last T, err error) {
+	next := WaitForUpdateOrDuration(ctx, last, source, period, loggerForTest)
```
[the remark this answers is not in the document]
ZhouXing19: Yup, updated.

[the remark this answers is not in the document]
ZhouXing19: Good point, I misunderstood the usage of `DoWhenChanged`. Updated it with for loop.

internal/util/stopvar/stopvar_test.go
ZhouXing19: Good point, copy-pasted from the original repo: https://github.com/cockroachdb/field-eng-powertools/blob/main/stopvar/stopvar_test.go

sravotto (commented):
I would suggest to open a PR in field-eng-powertools.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1079#-:-ODHFxAc64a0HM_yMo3D:b-u5dniy)* status: 0 of 17 files reviewed, 9 unresolved discussions (waiting on @Jeremyyang920, @noelcrl, @ryanluu12345, and @ZhouXing19)

PR #1080, opened 2024-11-26
gzip: Use sync.pool

Jeremyyang920:
This commit uses a sync.Pool for botht he gzip writer and reader so that we reduce the number of allocations and time GC takes as previously every mutation that required to be gzipped would make a call to newWriter and allocate a new object. This in turn spent a lot of time and created extra objects on the heap that were un needed which drove up GC time


internal/staging/stage/gzip.go
```diff
@@ -55,9 +72,13 @@ func maybeGunzip(data []byte) ([]byte, error) {
 	if len(data) < 2 || data[0] != 0x1f || data[1] != 0x8b {
 		return data, nil
 	}
-	r, err := gzip.NewReader(bytes.NewReader(data))
+
+	gzReader := gzipReaderPool.Get().(*gzip.Reader)
+	defer gzipReaderPool.Put(gzReader)
```
stevendanna, line 77: If Reset() failed below, are we sure we want to put this back in the pool?

Jeremyyang920: Hmm thats a good question. Im looking at the Reset code, and I think it just calls bufio.NewReader(r) to reset it back. So the put call is really just putting the base gzip.Reader{} object back and reset internally resets z.r which is the actual reader.

Does that check out with your understand of reset as well?

ryanluu12345:
Unrelated to the change itself, but I think @noelcrl mentioned this in standup that the container images for `v20.1` and `v20.2` are no longer present. You can just rebase after this goes in:
https://github.com/cockroachdb/replicator/pull/1081

internal/staging/stage/gzip.go
```diff
@@ -29,14 +30,30 @@ import (
...
 // maybeGZip compresses the given data if it is larger than gzipMinSize.
 func maybeGZip(data []byte) ([]byte, error) {
 	if len(data) <= gzipMinSize {
 		return data, nil
 	}
 
 	var buf bytes.Buffer
-	gzWriter := gzip.NewWriter(&buf)
+	gzWriter := gzipWriterPool.Get().(*gzip.Writer)
+	gzWriter.Reset(&buf)
```
ryanluu12345, line 54: From the docs, I see that:
```
Reset discards the [Reader](https://pkg.go.dev/compress/gzip#Reader) z's state and makes it equivalent to the result of its original state from [NewReader](https://pkg.go.dev/compress/gzip#NewReader), but reading from r instead. This permits reusing a [Reader](https://pkg.go.dev/compress/gzip#Reader) rather than allocating a new one.
```
Do we happen to have any idea of how expensive this state reset costs relative to allocations? Definitely understand that extra allocations here will probably be much more expensive.

ryanluu12345: Following up on the above message, is there any way we can run a simple workload and see what the profile looks like before this change, and then another test with a profile after this change and just show the profile charts?

internal/staging/stage/gzip.go
```diff
@@ -29,14 +30,30 @@ import (
...
+	defer gzipWriterPool.Put(gzWriter)
```
ryanluu12345, line 55: Just for my own knowledge, does `Get` remove a gzip writer from the pool? And then after we use this gzip writer, we want to `Put` this back so that it's available for usage by others? How does this exactly work?

ryanluu12345 (commented):
General implementation looks good to me. Pending Steven's question and light testing to see the profile changes before/after change.

internal/staging/stage/gzip.go
Jeremyyang920: https://victoriametrics.com/blog/go-sync-pool/
This is a good overview of how it works.

Jeremyyang920: But yes its exactly as you describe it in that it is a thread safe way for various go routines to share an object without consitent allocations.

internal/staging/stage/gzip.go
Jeremyyang920: The reset should be much cheaper than allocating the object each time. If you dig into gzip.NewWriter, there is a call to z := new(Writer) which will always require malloc to do work to allocate the new object entirely. Reset now would be reusing the same object and just reinstantiating the struct fields pretty much.

Jeremyyang920: I don't have a workload. I was just going to give the customer the build and ask for a new profile since we dont have the exact workload they were using. But certainly any workload that has mutations above 1024 bytes will trigger the zip path so could test that as well.

ryanluu12345: Awesome, thanks for clarifying. Sounds like it's very inexpensive: just reinstantiating struct fields like you mentioned.

Ok, that's totally fine. We can probably just send a locally built version or one from this PR build to Fidelity to send us a profile back.

internal/staging/stage/gzip.go
ryanluu12345: Great, thanks for the deep dive. Clear article and helpful examples.

internal/staging/stage/gzip.go
ryanluu12345: Yep makes sense, let's follow back up with the customer for a new profile with this new build then.

internal/staging/stage/gzip.go
stevendanna: I've never looked at the inside of Reset before today :D.

Looking at this code, I'd be a little concerned that this could result in us keeping `data` alive for longer than is necessary because the gzipReader will still have a reference to it.

Given that your profile only shows the writer being a problem, I wonder if it would be better to start there.

Jeremyyang920: It's a good call out. The reader doesn't show up in this profile since I think the other things dominated it.

We're doing a lot of profile work on this now, so I can certainly test that as well. The new heap profiles after this pool change don't seem to point to a big issue of `data` living for longer than necessary. But it could also be that GC is running very aggressively due to other poorly managed allocations.

Jeremyyang920: image
For a reference, here is a new CPU profile taken after teh pooling changes. Im highlighting over where the reader shows up which is much smaller that the writer. But certainly there are way more places to look a for bad allocations or things remaining on the heap leading to tha giant GC block still.

PR #1082, opened 2024-11-27
mylogical: added in handling for unsigned bigints

ryanluu12345:
Previously, bigint unsigned ould be interpreted as int64, which meant that over-large values would overflow and be interpreted as negative values. This change aims to make mylogical aware of signedness and treat the bigint unsigned case as uint64, which means that over-large values can now be interpreted properly. There is also logic introduced to get the signedness from the schema for older MySQL dialects.
    
Resolves: CC-30721
Release Note: Fixing issue with unsigned bigints overflowing.

This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1082)


ZhouXing19:
I tested this locally and seems working fine:

- MySQL source:

```sql

mysql> CREATE TABLE outranget (x int PRIMARY KEY, v bigint unsigned);
Query OK, 0 rows affected (0.02 sec)

mysql> select @@GLOBAL.gtid_executed;
+-------------------------------------------+
| @@GLOBAL.gtid_executed                    |
+-------------------------------------------+
| 848e34b6-adb9-11ef-ad31-32f46cde3d73:1-10 |
+-------------------------------------------+
1 row in set (0.01 sec)

mysql> INSERT INTO outranget (x, v) VALUES (1, 9823372057939841974);
Query OK, 1 row affected (0.02 sec)
```

- Replicator
```bash
 go run . mylogical --sourceConn="mysql://root:SoupOrSecret@localhost:3306/janedb/?sslmode=disable" --targetConn="postgresql://root@localhost:26258/defaultdb?sslmode=disable" --targetSchema=defaultdb.public --verbose=1 --stagingC
reateSchema --defaultGTIDSet="848e34b6-adb9-11ef-ad31-32f46cde3d73:1-10"
```

- CRDB (26258)

```sql
root@localhost:26258/defaultdb> CREATE TABLE outranget (x int PRIMARY KEY, v decimal);    
root@localhost:26258/defaultdb> select * from outranget;   
  x |          v
----+----------------------
  1 | 9823372057939841974
(1 row)

Time: 2ms total (execution 2ms / network 0ms)
```

PR #1085, opened 2024-12-03
.github/workflows: updated the binaries job to publish a version manifest

ryanluu12345:
Previously, binaries were published to the GCS bucket but weren't tracked in a version manifest. The change here introduces a version manifest that allows users to find all available release versions and download them from a single location.

Resolves: CC-30305
Release Note: None

**Testing**
**Successful Run**: https://github.com/cockroachdb/replicator/actions/runs/12144201220
**Skipped manifest run when not release**: https://github.com/cockroachdb/replicator/actions/runs/12144381340

**Release Behavior**
1. On every merge to master, we still publish a `master` build that mirrors the latest tip of the master branch
2. Whenever a release happens, we create a "latest" build for each platform-os-target combination as well as a tagged versioned build (i.e. v1.0.5) and upload it to the bucket
3. After the release upload happens, we offload the manifest generation to a new manifest job that creates the version manifest from the new uploaded builds

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1085)


.github/workflows/go-binaries.yaml
```diff
@@ -149,3 +162,89 @@ jobs:
...
+      # Copying the versions.txt file to the `artifacts` directory
+      # will result in it being synced to the GCS bucket.
+      - name: Write the versions text file with updated binary links
+        # We want to clean the replicator.tgz if it exists.
+        # rm -f doesn't return an error if it doesn't exist; works for
+        # our use case.
+        run: |
+          ls
+          mkdir -p ./artifacts
+          echo "" >> ./versions.txt
```
ZhouXing19, line 198: Is this redundant lines?

ryanluu12345: Good call out, let me fix.

ryanluu12345: Fixed!

.github/workflows/go-binaries.yaml
```diff
@@ -149,3 +162,88 @@ jobs:
...
+      - uses: actions/checkout@v4
+
+      - name: Pull the previous versions text file from GCP
+        run: |
+          curl --fail-with-body -o versions.txt -s 'https://replicator.cockroachdb.com/versions.txt' || cp ./climanifest/base-versions.txt versions.txt
+      # Copying the versions.txt file to the `artifacts` directory
+      # will result in it being synced to the GCS bucket.
+      - name: Write the versions text file with updated binary links
+        # We want to clean the replicator.tgz if it exists.
+        # rm -f doesn't return an error if it doesn't exist; works for
```
noelcrl, line 193: Not sure what this `rm -f` is referring to, is the comment outdated?

noelcrl (approved):

.github/workflows/go-binaries.yaml
ryanluu12345: Good call out, stale comment.

PR #1088, opened 2024-12-04
.github/workflows: added workflow updates to bump the replicator version on release

ryanluu12345:
Previously, the replicator version was not able to be introspected from the versions command, now that there is the ability to get the version from the binary itself, we want to add a methodology to bump the version in CI and create a PR for this behavior. This also includes replacing the API version with the new version before building.

Resolves: CC-30761
Release Note: None

**Example of created PR**
https://github.com/cockroachdb/replicator/pull/1091

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1088)


ryanluu12345:
I did a manual test where I hardcoded the replace version to 2.0.0 and ended up getting the proper version on the built binary:
```
(⎈|rancher-desktop:N/A)ryanluu@crlMBP-XFFK6Q9QL6MTk1 rep-test % ./replicator/replicator --version
replicator version v2.0.0
```

.github/workflows/go-binaries.yaml
```diff
@@ -242,5 +255,47 @@ jobs:
           parent: false # Don't include the upload directory name
           destination: ${{ vars.REPLICATOR_BUCKET }}/
           process_gcloudignore: false # Suppress warning about missing .gcloudignore file
+  
+  # TODO: in order for this CI to run properly and use
+  # create-pull-request, it must first be whitelisted in the actions
+  # permissions. This is pending dev inf help too.
+  create-pull-request:
```
noelcrl, line 262: Nit: should we add a comment here explaining that this job creates a pull request to bump the version constant in the Replicator binary?

noelcrl (approved):

.github/workflows/go-binaries.yaml
ryanluu12345: Done!

PR #1103, opened 2025-01-01
oraclelogminer: added in support for LOB replication in the LogToKV layer

ryanluu12345:
Previously, LOB data would cause a crash because of antlr parsing oddities. The logic here aims to do a few things:
1. Update the insert and update queries that have EMPTY_CLOB or EMPTY_BLOB function calls to insert empty strings instead
2. Convert the hextoraw(...) data to actual bytea strings that can be interpreted in a CRDB query
3. Determine which items in the KV map for set and where clauses need to be updated with these parsing rules

Resolves: CC-31048
Release Note: None


internal/source/oraclelogminer/log_to_kv.go
```diff
@@ -17,16 +17,51 @@
...
+// insert should be the same as EMPTY_BLOB/EMPTY_CLOB.
+func replaceEmptyLobsWithEmptyString(input string) string {
+	return emptyLobRegex.ReplaceAllString(input, "''")
+}
+
 // LogToKV parse a sql stmt log to a SetKV struct, where the key is the column to be rewritten, the value
 // is the value to override / insert. The true extraction logic can be found in the functions related
 // to oracleparser.MockListener.
 // Examples can be found in TestLogToKV().
 func LogToKV(log string) (oracleparser.SetAndWhereKVStructs, error) {
```
ryanluu12345, line 53: TODO: take out all the verbose logging

internal/source/oraclelogminer/log_to_kv_test.go
```diff
@@ -32,6 +33,66 @@ type logToKVTestCase struct {
 
 func TestLogToKV(t *testing.T) {
 	for i, tc := range []logToKVTestCase{
+		{
+			input: `insert into "C##MYADMIN"."LOB_TABLE"
+("x","y","BLOB_COL","RAW_COL","LONG_RAW_COL","CLOB_COL","NCLOB_COL") 
+values 
+('9','10115','',HEXTORAW('52415731'),NULL,'','');`,
```
ryanluu12345, line 40: For insert or update need to test out the latest LOB_TABLE and see why RAW is not being updated properly in the first INSERT.

internal/source/oraclelogminer/log_to_kv_test.go
```diff
@@ -32,6 +33,66 @@ type logToKVTestCase struct {
 
 func TestLogToKV(t *testing.T) {
 	for i, tc := range []logToKVTestCase{
+		{
+			input: `insert into "C##MYADMIN"."LOB_TABLE"
```
ryanluu12345, line 37: Need to figure out cases where there are non LOB cols and LOB cols but its majority non LOB.

ryanluu12345:
@ZhouXing19 also realized that I just based this off your latest PR branch, but that has the WIP for the new parsing scheme huh? I'm seeing certain bugs here which makes sense still it's still a WIP.

By the way, this change works fine with the branch version of oracle-source-0826 like two months ago. But we'll need to reevaluate this once the new method is in. It's not properly replicating data over.

ryanluu12345:
So I dug into this behavior a bit more closely and notice a bug with how we are doing updates for the LOB case. As previously mentioned for lobs:
1. Insert occurs with non-LOB columns
2. Update occurs with LOB columns for the record from step 1
```
insert into "C##MYADMIN"."LOB_EASY"("x","y","RAW_COL","BLOB_COL") values ('169800','1800',HEXTORAW('52415731'),'');

update "C##MYADMIN"."LOB_EASY" set "BLOB_COL" = HEXTORAW('343836353643364336463230344637323631363336433635') where "x" = '169800' and "y" = '1800' and "RAW_COL" = HEXTORAW('52415731') and ROWID = 'AAASMfAAHAAAALOAAX';
```

What we actually see here however is that the update seems to clobber what happened in the insert. So from the above, we fully expect that every field will be filled out like how it is in the source:
```
        x       y RAW_COL     BLOB_COL                                            
_________ _______ ___________ ___________________________________________________   
   169800    1800 52415731    343836353643364336463230344637323631363336433635 
```

However, if you look closer here, you'll see on the target:
```
root@localhost:26257/defaultdb> SELECT * FROM "LOB_EASY";                     
    x    |  y   |                  RAW_COL                   |                      blob_col
---------+------+--------------------------------------------+-----------------------------------------------------
  169800 | NULL | NULL                                       | \x343836353643364336463230344637323631363336433635
```

In this case, any non PK element is basically set as NULL. This seems to imply that the update behavior is off here. Instead of hydrating with existing data in the target table, it just upserts NULL.

PR #1108, opened 2025-01-07
.github/workflows: added in create docs ticket automation

ryanluu12345:
The automation here now creates a DOCs ticket based on the release notes between the last release tag and current one.

Resolves: CC-30799
Release Note: None


.github/workflows/go-binaries.yaml
```diff
@@ -28,6 +28,10 @@ on:
 jobs:
   binaries:
     name: Binaries
+    needs: [ create-docs-ticket ]
+    # Conditional "needs" blocks are not supported in GH actions.
+    # The if condition below works around this limitation.
```
noelcrl, line 33: If I recall correctly, this is necessary so that `create-docs-ticket` will run before `binaries`, and `binaries` will still run even if `create-docs-ticket` doesn't run (for a non-release)?

.github/workflows/go-binaries.yaml
```diff
@@ -286,5 +290,55 @@ jobs:
...
+          echo "tag=${GITHUB_REF_NAME#v}" >> $GITHUB_OUTPUT
+          echo "release_date=$(date '+%Y-%m-%d')" >> $GITHUB_OUTPUT
+          # We are sorting the tags by semver and getting the second highest because that is
+          # the latest released version. The highest semver is the one currently being released
+          # since the tagging happens right before this runs.
+          latest_tag=$(git tag -l | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+$' | sort -V | tail -2 | head -1)
+          echo "$latest_tag is the latest tag"
+
+          # Filters all commits from latest tag to now, looks for release notes, excludes empty ones, and adds a dash at the front of each line.
+          output=$(git log $latest_tag..HEAD | grep -i "release note:" | grep -vi 'release note: none' | sed 's/^ *Release Note: *//I' | sed 's/^/- /')
```
noelcrl, line 314: I'm assuming we will always make sure to use `release note: none` instead of something else like `release note: N/A` or just `'release note:'`? Do we have some sort of pre-commit hook in replicator to tell us to format it a certain way that this job expects?

noelcrl (approved):
LGTM! Just a few thoughts.

PR #1111, opened 2025-01-14
[WIP] oraclelogminer: support simple LOB and timezone types

ryanluu12345:
Timezones need to be handled in a custom manner because of how the data comes in as `TO_TIMSTAMPTZ` which will need the function applied to the data inside.

Additionally, we need to process simple LOBs so that we replace `EMPTY_CLOB` or `EMPTY_BLOB` directives with empty strings so they insert empty BLOB/CLOB types; this also needs specialized processing for us to handle the HEX2RAW case which will need to turn the data to a bytestring.

Resolves: CC-31049, CC-31048 (partially)
Release Note: None


ZhouXing19:
The logic LGTM!

PR #1112, opened 2025-01-17
[WIP] acceptors: added in a concept of deferred deletes for multi batches

ryanluu12345:
Previously, there was a bug that an insert and delete on the same PK would lead to the delete being executed first and then the insert, no matter the order of the insert or delete. This behavior is to handle the possibility of deletes that need to happen on a dependent table that has an FK reference. However, it breaks in the case that an insert then delete happens on the same PK, since the delete runs first, which means there is an extraneous inserted row.

The logic here aims to preserve the existing logic for the pending updates, early deletes, and final update flush, while also introducing a concept of a deferred delete. This deferredDelete batch only has items that are guaranteed to be the last delete for a given key in the multi batch (across all batches). This is essential because it means we can batch all of the updates and flush those without worrying that there is an extraneous delete.

Resolves: CC-30958
Release Note: Fixed correctness issue for insert + delete case in same TX for the same PK.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1112)


ryanluu12345:
Actually did some testing locally and was seeing that we have FK violations; this is probably largely due to the change in behavior. I need to do some testing for FK workloads here. But here is what I'm seeing as an issue while doing this:
```
time="Jan 16 22:48:03" level=warning msg="error while copying data for \"tgt-60905-3.public\" [ \"tgt-60905-3\".\"public\".\"tbl-1\", \"tgt-60905-3\".\"public\".\"tbl-2\" ]; will restart" detail="ERROR: delete on table \"tbl-1\" violates foreign key constraint \"parent_fk\" on table \"tbl-2\" (SQLSTATE 23503)\ngithub.com/cockroachdb/replicator/internal/target/apply.(*apply).deleteLocked\n\t/Users/ryanluu/go/src/github.com/cockroachdb/cdc-sink/internal/target/apply/apply.go:350\ngithub.com/cockroachdb/replicator/internal/target/apply.(*apply).Apply\n\t/Users/ryanluu/go/src/github.com/cockroachdb/cdc-sink/internal/target/apply/apply.go:234\ngithub.com/cockroachdb/replicator/internal/target/apply.(*Acceptor).AcceptTableBatch\n\t/Users/ryanluu/go/src/github.com/cockroachdb/cdc-sink/internal/target/apply/acceptor.go:48\ngithub.com/cockroachdb/replicator/internal/types.(*orderedAdapter).AcceptTableBatch\n\t/Users/ryanluu/go/src/github.com/cockroachdb/cdc-sink/internal/types/acceptors.go:94\ngithub.com/cockroachdb/r
```

ryanluu12345:
This approach scales better: https://github.com/cockroachdb/replicator/pull/1115

PR #1114, opened 2025-01-17
pglogical and mylogical: fold by key before passing data to the orderedAcceptor

ryanluu12345:
Previously, there was a bug resulting from performing a combination of inserts and deletes could lead to data inconsistency because of how deletes and inserts are handled in the AcceptMultiBatch to handle the possibility of FK updates/deletes. As a result, we do a FoldByKey for each table batch before we send it to the acceptor so that only the last mutation for a single table and single PK remains. This makes it so that it is safe to perform the deletes first and then the updates after.

The caveat here is that we don't have the same guaranteees for CRDB so we can end up in a situation where for CRDB to X, we could end up in this erroneous state for the delete + insert case.

Resolves: CC-30958
Release Note: None

**Testing**
For both pglogical and mylogical, tested with the following cases and confirmed it works fine. I confirmed that it flattens to only one mutation for a given PK.
```
# Insert delete insert on same key
molt=# BEGIN;               

INSERT INTO tbl1 (id, t) VALUES (12, 'aaa');

DELETE FROM tbl1 WHERE id = 12;

INSERT INTO tbl1 (id, t) VALUES (12, 'aaa');

COMMIT;

# Insert delete on same key
molt=# BEGIN;               

INSERT INTO tbl1 (id, t) VALUES (12, 'aaa');

DELETE FROM tbl1 WHERE id = 12;

COMMIT;

# Delete insert on same key
BEGIN;

DELETE FROM tbl1 WHERE id = 12;

INSERT INTO tbl1 (id, t) VALUES (12, 'aaa');

COMMIT;
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1114)


ryanluu12345:
Closing in favor of : https://github.com/cockroachdb/replicator/pull/1115/files

PR #1115, opened 2025-01-17
internal/types/acceptor: added an option configuration to fold a multi-batch by key

ryanluu12345:
Previously, there was a bug resulting from performing a combination of inserts and deletes could lead to data inconsistency because of how deletes and inserts are handled in the AcceptMultiBatch to handle the possibility of FK updates/deletes. As a result, we do a FoldByKey for each temporal batch in the multi batch before we flush to the target, so that only the last mutation for a single table and single PK remains. This makes it so that it is safe to perform the deletes first and then the updates after.

This is also extensible in the future since we can control the behavior for this folding via a configuration flag, instead of having to plumb in a new acceptor each time; this allows us to turn on and off this ability to fold.

Resolves: CC-30958
Release Note: None

**TODO**
- [x] Add new tests for the previously wrong behavior
- [x] Remove hack for FoldByKey (but right now may require refactor)
- [x] Performance testing at scale to see if makes a difference (@Jeremyyang920)

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1115)


internal/types/acceptors.go
```diff
@@ -252,3 +277,109 @@ func (u *unorderedAdapter) AcceptTemporalBatch(
 	}
 	return nil
 }
+
+// TODO: fix this hack since this is meant to work around the fact that
+// right now `msort` imports `internal/types` and vice versa.
+// Need to refactor this into a new place so that this and msort both
+// import from `types`.
+// This is a workaround.
+func FoldByKey(x []Mutation) ([]Mutation, error) {
```
ryanluu12345, line 286: @sravotto @ZhouXing19 any thoughts on how to best resolve this issue? I was thinking of moving the acceptors into their own package since keeping it in types seems too coupled and also leads to import cycles like this.

Any alternatives?

internal/types/utils.go
```diff
@@ -0,0 +1,122 @@
...
+import (
+	"bytes"
+	"encoding/json"
+	"slices"
+
+	"github.com/cockroachdb/replicator/internal/util/hlc"
+	"github.com/pkg/errors"
+)
+
+// FoldByKey now lives in this package since it's crucial to the
```
ryanluu12345, line 12: Before considering this approach I did a local prototype of moving `acceptor` to its own package so we avoid the import cycle. However, this ends up leading to a mass refactor across the codebase that would impact in-flight PRs. After doing more investigation, I also realized that the end result is still an import cycle because `types` is required by `acceptor` and vice versa.

Ended up just moving all references to point to `types.FoldByKey` and this is a much simpler and safer change.

internal/source/pglogical/conn.go
```diff
@@ -157,6 +157,7 @@ func (c *Conn) accumulateBatch(
 
 			if err := c.acceptor.AcceptTemporalBatch(ctx, batch, &types.AcceptOptions{
 				TargetQuerier: tx,
+				FoldKeys:      true,
```
ryanluu12345, line 160: A TODO that we must consider when we pick this back up next week is to ensure that this `FoldKeys` behavior is gated by a flag (I recommend `--collapseUpdates` or `--useBatchMode`) so that we can turn on or off the behavior. Here are the tradeoffs:
- If FoldKeys is set to on: then we do extra O(N) processing to do the collapsing and reassign the proper batches
- If FoldKeys is set to off: then we may run the risk of a correctness issue for the same PK being updated multiple times (since we have the bug in the FK logic)

What we have learned from previous customers is that it's helpful to have this escape hatch so they can turn off undesirable behavior as needed.

internal/types/utils.go
ryanluu12345: Implemented this, but the open question here is should we keep this off or on by default. If we keep this off, then we have the same behavior as before, but is exposed to this correctness issue. If we keep this on, then we may have a perf impact because we do the collapsing every time we accept a batch, which in PG and MySQL right now would do it every time commit happens.

ryanluu12345:
> The whole logic LGTM, just thinking that tests are needed. I'm thinking why we should leave the foldbykey as an option rather than always executing it, as I thought we should always prioritize correctness over performance.

I think mainly because this fixes a correctness issue in the case that you could have an update/insert/update in the same batch. However, not all customer's workloads are like that. In certain customer cases, this may regress their performance at scale because the extra O(N) operations. And also because we do expensive JSON decoding to do this folding. So, wanted to give an escape hatch here if they understand their workload.

How open are you to just making this flag true by default so it's correct behavior. But if a customer knows for a fact their workload won't hit this, they can turn it off with the flag?

sravotto (changes requested):
I would consider moving `msort.go` into the `types` package.  I think `UniqueByTimeKey` is only referenced once.
Other than that, looks good to me. 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1115#-:-OH8Xko97Y_Rbt7i6Yxn:bqklzgi)* status: 0 of 14 files reviewed, 4 unresolved discussions (waiting on @Jeremyyang920, @noelcrl, @ryanluu12345, and @ZhouXing19)

___
*[`internal/types/utils.go` line 28 at r4](https://reviewable.io/reviews/cockroachdb/replicator/1115#-OH8XQAQ5UgSVoTTw5Zn:-OH8XQAQ5UgSVoTTw5Zo:bks0h0g) ([raw file](https://github.com/cockroachdb/replicator/blob/f9ba3a4573900c0e346ccead5eb1a3a55d1e7872/internal/types/utils.go#L28)):*
> ```Go
> )
> 
> // FoldByKey now lives in this package since it's crucial to the
> ```

Would it make sense just to move the whole msort.go (and test) under types?

ryanluu12345 (commented):
I'm also in agreement that `true` should be the default. Will loop in PM review here. 

@sravotto , good callout, I'll go ahead and move msort and relevant tests to `types` and update all references.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1115#-:-OH92nPz3vkfEav0X4tk:biv4kne)* status: 0 of 17 files reviewed, 4 unresolved discussions (waiting on @Jeremyyang920, @noelcrl, and @ZhouXing19)

ryanluu12345 (commented):
One more note here: I'll work on adding the `INSERT -> DELETE` use case some time this week and will bottom out on the default behavior.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1115#-:-OH92xYx16BbwWYjBVJN:b-e3rlnu)* status: 0 of 17 files reviewed, 4 unresolved discussions (waiting on @Jeremyyang920, @noelcrl, and @ZhouXing19)

ryanluu12345:
> One more note here: I'll work on adding the `INSERT -> DELETE` use case some time this week and will bottom out on the default behavior.
> 
> > _[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1115#-:-OH92xYx16BbwWYjBVJN:b-e3rlnu)_ status: 0 of 17 files reviewed, 4 unresolved discussions (waiting on @Jeremyyang920, @noelcrl, and @ZhouXing19)

@ZhouXing19 @sravotto so I'm looking into our testing right now and see that we have `internal/source/pglogical/integration_test.go` which has a `TestPGLogical` which has a helper that does the setup for schema, create table, and the data movement.

I plan to add a new set of configurations that helps to reproduce the `INSERT -> DELETE -> INSERT` logic here and then add a new test case. Thoughts?

sravotto (commented):
Makes sense to me.

> Reviewed 1 of 5 files at r1, 4 of 10 files at r6.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1115#-:-OH9SB3hA0GJBbfKOJmN:b-a7ljwd)* status: 4 of 18 files reviewed, 3 unresolved discussions (waiting on @Jeremyyang920, @noelcrl, and @ZhouXing19)

ryanluu12345:
@sravotto @ZhouXing19 when you get a chance, wondering if you can give a review on the latest iteration? Just added in the testing and have a path forward for the flag or no flag.

@rohan-joshi and I chatted and he requested that I benchmark to understand better how much timing difference doing this collapseMutation takes vs. not doing it. If the impact is trivial, then the vote goes towards not adding the flag. I'll do this sometime early next week.

sravotto (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1115#-:-OHDTeFK2bLXxdaBDi-i:bnfp4nl)* status: 4 of 21 files reviewed, 3 unresolved discussions (waiting on @Jeremyyang920, @noelcrl, and @ZhouXing19)

internal/target/apply/apply.go
```diff
@@ -160,7 +159,7 @@ func (a *apply) Apply(ctx context.Context, tx types.TargetQuerier, muts []types.
 	// updates to rows, rather than complete rows. Instead of pushing
 	// this complexity out to the frontend, we'll solve it here by
 	// folding mutations for the same key together.
-	muts, err := msort.FoldByKey(muts)
+	muts, err := types.FoldByKey(muts)
```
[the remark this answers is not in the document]
ryanluu12345: Got it!

internal/types/acceptors.go
```diff
@@ -145,6 +165,7 @@ func (t *orderedAdapter) AcceptMultiBatch(
 		// error is returned from the callback.
 		for table, tableBatch := range temporal.Data.All() {
 			for _, mut := range tableBatch.Data {
+				fmt.Println("accepted mutations: ", table.String(), string(mut.Key), string(mut.Data))
```
ZhouXing19, line 168: nit: debug code, can remove

internal/types/msort.go
```diff
@@ -14,45 +14,44 @@
...
-func FoldByKey(x []types.Mutation) ([]types.Mutation, error) {
+func FoldByKey(x []Mutation) ([]Mutation, error) {
 	// Handle trivial cases.
 	if len(x) <= 1 {
 		return x, nil
 	}
 	// Ensure that values in the slice are well-ordered. This should be
 	// the general case.
-	sorted := slices.IsSortedFunc(x, func(a, b types.Mutation) int {
+	sorted := slices.IsSortedFunc(x, func(a, b Mutation) int {
```
ZhouXing19, line 45: Can leave it as a TODO, but I think we should mark the in-txn ordering in `Mutation.Time` as well. The current implementation is based on the assumption that the ordering of the mutation from the frontend (e.g. pglogical) is never interrupted. While this might be true, I think it will be less error-prone if we always a reliable explicit ordering mark.

ryanluu12345:
@rohan-joshi @ZhouXing19 just wanted to give a heads up here that @Jeremyyang920 tested with a scaled customer-representative workload and verified a few things:
- The throughput is about the same 
- Stability at scaled throughput is reasonable (had a false alarm but later determined it was because of another change that we've since removed)

Thoughts on path forward then? My opinion is that it's still valuable to flag-gate this in case we missed something in the testing and we regressed. Always better to be safe then sorry. Then we can deprecate this two releases from now.

internal/types/msort.go
ryanluu12345: Yeah, this is a good consideration so that we ensure that within a transaction the data comes in order. For pg and mysql, I'm not too considered since we receive the messages from the protocols as a single ordered stream. But how does this work in the Oracle case?

rohan-joshi:
flag-gate makes sense

internal/source/pglogical/config.go
```diff
@@ -82,6 +83,7 @@ func (c *Config) Bind(f *pflag.FlagSet) {
 		"how often to report WAL progress to the source server")
 	f.Var(ident.NewSchemaFlag(&c.TargetSchema), "targetSchema",
 		"the SQL database schema in the target cluster to update")
+	f.BoolVar(&c.CollapseMutations, "collapseMutations", true, "collapse mutations will fold all mutations on a given PK within a batch into a final resulting mutation")
```
ryanluu12345, line 86: @taroface wondering if I can have some copy help here? This is in 3 separate frontends, but they should all have the same message, so we only need to review one.

taroface: My suggestion:

```collapse all mutations on each primary key into a single resulting mutation```

Comment (ryanluu12345) [reply internal/source/pglogical/config.go]: I'll do this.

File: internal/types/acceptors_test.go [hunk 8]
```diff
@@ -83,7 +98,9 @@ func TestOrderedAcceptor(t *testing.T) {
 	// We expect to see exactly one call per table and the table order
 	// should respect the level order.
 	calls := rec.Calls()
+	foldCalls := foldRec.Calls()
 	r.Len(calls, len(allTables))
+	r.Len(foldCalls, len(allTables))
```
ZhouXing19, line 103: Hmm i don't understand why we add this additional `foldRec` but not checking the eventual result of `foldCalls`. What is the intention of this test? 

As per the result here, with or without FoldKeys, the eventual mutations in `Calls()` are of the same length. How can we tell the difference that `FoldKeys` brings? In other words, should we loop over the `foldCalls` and confirm we only see the collapsed mutations?

internal/types/msort.go
ZhouXing19: This will be considered and resolved in the oracle pr, with the xid and redo value marked in the time struct (we might need to reconstruct the hlc.time struct, but not a concern in this pr.)

internal/source/pglogical/integration_test.go
```diff
@@ -194,12 +198,39 @@ func testPGLogical(t *testing.T, fc *fixtureConfig) {
 				return tx.Commit(egCtx)
 			})
 		}
+
+		if fc.collapseMutations {
+			tx, err := pgPool.Begin(egCtx)
+			require.NoError(t, err)
+			_, err = tx.Exec(egCtx,
+				fmt.Sprintf("INSERT INTO %s VALUES ($1, $2)", src),
+				1000, "",
```
ZhouXing19, line 207: wouldn't 1000 possibly overlap with the for loop above? 

I would recommend using `additionalRowIdx := rowCount + 1` and insert that row, to avoid pk overlap.

ZhouXing19 (approved):
LGTM, some test related questions

internal/source/cdc/config.go
```diff
@@ -37,11 +37,12 @@ const (
 
 // Config adds CDC-specific configuration to the core logical loop.
 type Config struct {
-	ConveyorConfig  conveyor.Config
-	DLQConfig       dlq.Config
-	SequencerConfig sequencer.Config
-	SchemaWatch     schemawatch.Config
-	ScriptConfig    script.Config
+	CollapseMutations bool
```
noelcrl, line 40: Nit: Perhaps add a comment here describing `CollapseMutations` for consistency? The mylogical and pglogical configs seem to have one.

As a side note, it seems that many fields in the configs for cdc/mylogical/pglogical are duplicated. Not sure if it would be worth it to have a common Config struct in the future that each specific config can embed.

internal/target/apply/apply.go
```diff
@@ -160,7 +159,10 @@ func (a *apply) Apply(ctx context.Context, tx types.TargetQuerier, muts []types.
 	// updates to rows, rather than complete rows. Instead of pushing
 	// this complexity out to the frontend, we'll solve it here by
 	// folding mutations for the same key together.
-	muts, err := msort.FoldByKey(muts)
+	// TODO: consider removing this call since folding may not be
+	// necessary if it's already been folded previously. However, this
+	// may not always be guaranteed, especially in the unordered adapter case.
+	muts, err := types.FoldByKey(muts)
```
noelcrl, line 165: Should we also only call FoldByKey here if the `CollapseMutations` flag is set to true?

internal/source/pglogical/integration_test.go
```diff
@@ -194,12 +198,39 @@ func testPGLogical(t *testing.T, fc *fixtureConfig) {
 				return tx.Commit(egCtx)
 			})
 		}
+
+		if fc.collapseMutations {
```
noelcrl, line 202: Should we also add a test where `fc.collapseMutations` is false? This may help confirm that changing the flag back to false causes the original incorrect behavior (which the user accepts for their workload by changing it to false) and indicate that we aren't folding in `acceptMultiBatch()`. It could also help document the bug in test code.

internal/source/pglogical/integration_test.go
noelcrl: Agreed. I like having it as a variable instead of hardcoding it in different places of the test as well.

noelcrl (commented):
Looks good so far! Just some minor suggestions.

internal/target/apply/apply.go
ryanluu12345: I don't want to regress previous behavior. I think it makes sense for us to retain this

internal/source/pglogical/integration_test.go
ryanluu12345: Good catch here! I'll go ahead and take this recommendation.

internal/source/cdc/config.go
ryanluu12345: Yep, I think in the future we can create a new config object that can be embedded in each of these frontends. We do this for sequencer or stager configs if you look at the cdc case, but for the more one-offs, we do it this way. 

Not important for this PR but will be helpful when we think about future UX.

I'll go ahead and add a comment here too.

ryanluu12345 (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1115#-:-OHibGhj-CLVZ7iWHa32:bnfp4nl)* status: 4 of 21 files reviewed, 12 unresolved discussions (waiting on @Jeremyyang920, @noelcrl, @sravotto, @taroface, and @ZhouXing19)

noelcrl (approved):
LGTM!

noelcrl (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1115#-:-OHicPn-DrPzEdxFC2sb:bnfp4nl)* status: 4 of 21 files reviewed, 12 unresolved discussions (waiting on @Jeremyyang920, @ryanluu12345, @sravotto, @taroface, and @ZhouXing19)

ZhouXing19 (approved):
> Reviewed 2 of 15 files at r5, 5 of 12 files at r6, 4 of 9 files at r7, 1 of 2 files at r8, 1 of 1 files at r10, 1 of 3 files at r11, 1 of 2 files at r12, 4 of 5 files at r13, 2 of 2 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1115)* status: all files reviewed, 12 unresolved discussions (waiting on @Jeremyyang920, @ryanluu12345, and @taroface)

ZhouXing19 (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1115)* status: all files reviewed, 10 unresolved discussions (waiting on @Jeremyyang920, @noelcrl, @ryanluu12345, and @taroface)

ZhouXing19 (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1115)* status: all files reviewed, 8 unresolved discussions (waiting on @Jeremyyang920, @noelcrl, @ryanluu12345, and @taroface)

PR #1117, opened 2025-01-23
source: use core sequencer for mylogical and pglogical

ZhouXing19:
WIP


ryanluu12345 (commented):
Overall looks good. I'm wondering what aspects you think need more granular testing?

Also, I think that we should get the `FoldByKey` change in first before we get this change in. Do you think there will be any impact on the folding logic from moving to core sequencer here? My understanding is that we will still call the ordered adapter `AcceptMultiBatch`

internal/sequencer/scriptforcore/script.go
```diff
@@ -0,0 +1,133 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+// Package scriptforcore provides an API shim that integrates the userscript
+// into the [sequencer.Sequencer] API, but dedicated for the core.Core backend.
+// Eventually, scriptforcore will replace the current script package.
+package scriptforcore
```
ryanluu12345, line 20: Can you please explain why `scriptforcore` is needed? And why can't we just hook into `script` as is? Also, what is different between the implementation here and the one in `script`?

internal/sequencer/scriptforcore/provider.go
```diff
@@ -0,0 +1,37 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package scriptforcore
```
ryanluu12345, line 17: I think this package deserves a package comment here to explain the purpose and also how it differs from the existing `script` package.

internal/sequencer/scriptforcore/script.go
ryanluu12345: Like I mentioned in the comment above, will be helpful to explain in more detail this package and how it differs. More focused and why and what.

internal/sequencer/scriptforcore/script.go
```diff
@@ -0,0 +1,133 @@
...
+
+type wrapper struct {
+	*ScriptSequencer
+	delegate sequencer.Sequencer
+}
+
+var _ sequencer.Sequencer = (*wrapper)(nil)
+
+// Start injects a userscript shim into the ScriptSequencer stack.
+func (w *wrapper) Start(
```
ryanluu12345, line 55: What risks does shifting to `scriptforcore` bring versus using the existing `script`? How is our test coverage here and do we still hit the same coverage we had previously?

internal/sequencer/scriptforcore/script.go
```diff
@@ -0,0 +1,133 @@
...
+		opts.Delegate = types.OrderedAcceptorFrom(&targetAcceptor{
+			delegate:   opts.Delegate,
+			ensureTX:   ensureTX,
+			group:      opts.Group,
+			targetPool: w.targetPool,
+			userScript: scr,
+		}, w.watchers)
+	}
+
+	// Install the source-phase acceptor. This provides the user with
```
ryanluu12345, line 101: To clarify, this covers `configureSource` and the `target-phase` acceptor covers `configureTarget`?

internal/sequencer/scriptforcore/script_test.go
```diff
@@ -0,0 +1,438 @@
...
+			break
+		} else {
+			t.Logf("expected 1 row, got %d row, retrying", ct)
+			time.Sleep(100 * time.Millisecond)
+		}
+		cnt++
+	}
+}
+
+func TestUserScriptSequencer(t *testing.T) {
```
ryanluu12345, line 249: Solid testing that covers various transformations. I think at some point, we should aim to have test coverage that covers all the examples we have in `main.ts`.

internal/sequencer/scriptforcore/source_acceptor.go
```diff
@@ -0,0 +1,157 @@
...
+type sourceAcceptorForScript struct {
+	batchReader    types.BatchReader
+	group          *types.TableGroup
+	sourceBindings *script.Source
+	watcher        types.Watcher
+}
+
+// Read preprocesses the batch cursor from the parent source pipe and
+// push the processed batch cursor to the returned batch cursor channel.
+func (a *sourceAcceptorForScript) Read(ctx *stopper.Context) (<-chan *types.BatchCursor, error) {
```
ryanluu12345, line 44: Just wanted to confirm, but for this approach, once we accept the batch, we still call `AcceptMultiBatch` right? I want to ensure our `FoldByKey` logic is still good. We'll want to rebase this on that merged change.

internal/sequencer/scriptforcore/source_acceptor.go
```diff
@@ -0,0 +1,157 @@
...
+		}
+		for {
+			select {
+			case batchCursor, open := <-srcCh:
+				// The channel has been closed. This will happen when the
+				// stopper is being stopped.
+				if !open {
+					return nil
+				}
+				//if batchCursor.Batch != nil {
```
ryanluu12345, line 59: Remove dead code.

internal/sequencer/scriptforcore/source_acceptor.go
```diff
@@ -0,0 +1,157 @@
...
+					log.Errorf("error processing batch cursor for script for core: %v", err)
+					return err
+				}
+				cur := &types.BatchCursor{
+					Batch:    postProcessedBatch,
+					Progress: batchCursor.Progress,
+				}
+				select {
+				case retCh <- cur:
+					// TODO(janexing): debug only, remove it later.
```
ryanluu12345, line 78: Looks like we don't need this debug code either

internal/sequencer/seqtest/check.go
```diff
@@ -234,6 +253,20 @@ func (c *Check) Check(ctx *stopper.Context, t testing.TB, cfg *all.WorkloadConfi
 		generator.GenerateInto(testData, hlc.New(int64(i+1), 0))
 	}
 
+	if c.Partitioned {
```
ryanluu12345, line 256: How do we know if something is partitioned? Oh this is a test.

internal/sequencer/seqtest/check.go
```diff
@@ -331,15 +352,36 @@ func (c *Check) Check(ctx *stopper.Context, t testing.TB, cfg *all.WorkloadConfi
...
-	// Set desired range.
-	c.Bounds.Set(generator.Range())
-	r.NoError(err)
-
-	// Wait to catch up.
-	now = time.Now()
-	r.NoError(generator.WaitForCatchUp(ctx, stats))
-	log.Infof("caught up in an additional %s", time.Since(now))
-	generator.CheckConsistent(ctx, t)
+	//// Set desired range.
```
ryanluu12345, line 355: Remove dead code.

internal/source/pglogical/conn.go
```diff
@@ -65,51 +62,71 @@ type Conn struct {
...
-					log.Debugf("progressed to LSN: %s", lsn)
-					c.walOffset.Set(lsn)
+				if hlc.Compare(progress, oldProgress) <= 0 {
+					return nil
+				}
+				lsn := progress.External().(pglogrepl.LSN)
+				log.Debugf("progressed to LSN: %s", lsn)
+				c.walOffset.Set(lsn)
+
+				if err := c.memo.Put(ctx, c.stagingDB, c.walMemoKey, []byte(lsn.String())); err == nil {
```
ryanluu12345, line 124: So just wanted to confirm, but this gets updated _**only after we actually apply to target right**_?

internal/types/batches.go
```diff
@@ -263,6 +263,32 @@ type TemporalBatch struct {
 
 var _ Batch[*TemporalBatch] = (*TemporalBatch)(nil)
 
+// AccumulateTableBatch is to accumulate mutations from the given table batch
+// to the temporal batch.
+func (b *TemporalBatch) AccumulateTableBatch(t *TableBatch) error {
```
ryanluu12345, line 268: Why do we need this? I thought we had `AccumulateTableBatch` already?

go.sum
```diff
@@ -113,10 +113,16 @@ github.com/cockroachdb/crlfmt v0.3.0 h1:IaPIlidTHn7s493ozBcpnkF/HNzCNe7aBJmUMqgI
 github.com/cockroachdb/crlfmt v0.3.0/go.mod h1:iRZvVcb8vDQK4dh64mPRZBucM85Xd/VLTE0cmO95/Kk=
 github.com/cockroachdb/datadriven v1.0.2 h1:H9MtNqVoVhvd9nCBwOyDjUEdZCREqbIdCJD93PBm/jA=
 github.com/cockroachdb/datadriven v1.0.2/go.mod h1:a9RdTaap04u637JoCzcUoIcDmvwSUtcUFtT/C3kJlTU=
+github.com/cockroachdb/errors v1.11.3 h1:5bA+k2Y6r+oz/6Z/RFlNeVCesGARKuC6YymtcDrbC/I=
```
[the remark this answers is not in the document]
ZhouXing19: It's because i imported cockroachdb/errors.

PR #1120, opened 2025-01-27
batchCursor: add progress only field

ZhouXing19:
Previously, determining whether a batch was a sentinel message involved checking if the `Batch` field in the `BatchCursor` struct was nil. This approach made it difficult to trace and understand where such sentinel messages were being sent within the codebase.

This PR introduces a new explicit field, `ProgressOnly`, to the `BatchCursor` struct. This field will be set to `true` only for sentinel messages, **making the code more readable and improving navigation when debugging or reviewing the logic**.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/replicator/1120)


internal/sequencer/sequtil/copier.go
```diff
@@ -125,7 +126,13 @@ top:
 		// we're waiting on the db query buffers to refill. We'll take
 		// this opportunity to flush data.
 		batch := cursor.Batch
-		if batch == nil {
+		if batch == nil && !cursor.ProgressOnly {
+			return nil, errors.New("cursor with no batch info should only for updating progress")
+		}
+		if cursor.ProgressOnly {
+			if batch != nil {
+				return nil, errors.New("cursor is for progress only, but contain not nil batch")
```
Jeremyyang920, line 134: nit: 
```suggestion
				return nil, errors.New("cursor is for progress only, but contain non nil batch")
```

internal/sequencer/sequtil/copier.go
```diff
@@ -125,7 +126,13 @@ top:
 		// we're waiting on the db query buffers to refill. We'll take
 		// this opportunity to flush data.
 		batch := cursor.Batch
-		if batch == nil {
+		if batch == nil && !cursor.ProgressOnly {
+			return nil, errors.New("cursor with no batch info should only for updating progress")
```
Jeremyyang920, line 130:
```suggestion
			return nil, errors.New("cursor with no batch info should only be used for updating progress")
```

Jeremyyang920 (approved):
lgtm with the 2 quick edits.

internal/sequencer/sequtil/copier.go
```diff
@@ -125,7 +126,13 @@ top:
 		// we're waiting on the db query buffers to refill. We'll take
 		// this opportunity to flush data.
 		batch := cursor.Batch
-		if batch == nil {
+		if batch == nil && !cursor.ProgressOnly {
+			return nil, errors.New("cursor with no batch info should only be used for updating progress")
```
noelcrl, line 130: Question: Do these errors ever occur? Or are these checks more of a safeguard against misuse?

internal/types/batch_reader.go
```diff
@@ -28,9 +28,13 @@ import (
 // BatchCursor is emitted by [BatchReader.Read].
 type BatchCursor struct {
 	// A batch of data, corresponding to a transaction in the source
-	// database. This may be nil for a progress-only update.
+	// database.
 	Batch *TemporalBatch
 
+	// If true, this cursor is for notifying the progress only, and it
+	// should not contain batch info.
```
noelcrl, line 35: Nit: How about specifically stating something like "TemporalBatch should be nil" or "this batch should be nil"?

noelcrl (approved):
LGTM! Just a question for understanding and a suggestion.

internal/sequencer/sequtil/copier.go
ZhouXing19: Yeah it's just a gate to prevent misusage

